PR #2, opened 2011-05-23
balancing multiplication

lomereiter:
Multiplication by Karatsuba or Toom-Cook method is not effective when the two numbers are of different order of magnitude (for example, 167483-digit number times 1263-digit one). It's better to slice the larger number into pieces which size is roughly equal to the smaller number size. This algorithm was implemented in Ruby, and it works much faster.

### DIFF
--- a/bn_mp_balance_mul.c
+++ b/bn_mp_balance_mul.c
@@ -1,0 +1,102 @@
+#include <tommath.h>
+#ifdef BN_MP_BALANCE_MUL_C
+/* LibTomMath, multiple-precision integer library -- Tom St Denis
+ *
+ * LibTomMath is a library that provides multiple-precision
+ * integer arithmetic as well as number theoretic functionality.
+ *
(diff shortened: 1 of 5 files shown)


minad:
@czurnieden Any comments concerning the status of this PR? Is there hope to get this working and is it worth it?

czurnieden:
@minad 
> Is there hope to get this working and is it worth it?

There's always hope ;-)

But, no, not really. I already have one ready and tested and am just waiting until the new version gets out to start a new one with balanced multiplication, fast division, FFT/NTT and more.
Oh, and yes, it is worth it. Benchmarking shows significant results starting in the range `KARA_CUTOFF * (2*KARA_CUTOFF)`. But the price is memory overhead, so it should be able to be switched on/off.

minad:
@czurnieden I guess it is fine to close this one now. Did I understand correctly that you have a patch locally which could be resubmitted at some point?

PR #8, opened 2012-12-17
Plan9 mkfile

staalmannen:
This is a mkfile to build libtommath on Plan9/APE (ANSI-POSIX-ENVIRONMENT)


karel-m:
Now you should be able to utilize `makefile.unix` like this:
```
make -f makefile.unix CC=pcc CFLAGS="-I/sys/include/ape -B -D_POSIX_SOURCE -D_SUSV2_SOURCE -D_BSD_EXTENSION"
```

PR #18, opened 2013-09-03
Various small changes, add four new functions

abokth:
There's a modified copy of libtommath here:

https://github.com/heimdal/heimdal/tree/master/lib/hcrypto/libtommath

This is a merge of the modifications done to that copy.

Contains a few small changes and a few new functions, including mp_find_prime().

### DIFF
--- a/bn_mp_find_prime.c
+++ b/bn_mp_find_prime.c
@@ -1,0 +1,26 @@
+/* TomsFastMath, a fast ISO C bignum library.
+ *
+ * This project is public domain and free for all purposes.
+ *
+ * Love Hornquist Astrand <lha@h5l.org>
+ */
+#include <tommath.h>
+
+int mp_find_prime(mp_int *a)
+{
+  int res;
+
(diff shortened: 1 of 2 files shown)


sjaeckel:
@jaltman can you please trigger a review and test of the heimdal requirements regarding ltm and if everything's okay now?
We're trying to push a release out soon and would like to close this PR before.

tomstdenis:
I will peek at this after the v1.0 release.

sjaeckel:
I think that one is ready as well.
Should be visible when looking at c82028641f77bcea872b92b61c343bcb0da02366...a4b20be8a2ed973f5c47e028dc87793b2d760c4a

tomstdenis:
That doesn't include the "mp_zero_multi" and what not (not that I've ever had a need for that... ints are set to zero by default..."

sjaeckel:
any idea what we should do with this issue?

czurnieden:
Would be obsolete, too. Propose to close.

PR #19, opened 2013-09-20
Speed up bn_map_toradix

BenGoldberg1:
Instead of repeatedly performing big number division by radix, we repeadly perform big number division by the largest power of radix which is <= MP_DIGIT_MAX.

Each remainder (which is an mp_digit) is repeatedly divided by radix, using ordinary integer division.

Since integer division is much faster than mp_div_d, this is a huge speedup.

When radix > sqrt(MP_DIGIT_MAX), the function should be the same speed as the original version.

I have not tested or benchmarked this code, but I know that the basic idea is sound.

### DIFF
--- a/bn_mp_toradix.c
+++ b/bn_mp_toradix.c
@@ -18,10 +18,12 @@
 /* stores a bignum as a ASCII string in a given radix (2..64) */
 int mp_toradix (mp_int * a, char *str, int radix)
 {
-  int     res, digs;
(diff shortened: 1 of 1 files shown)


minad:
@sjaeckel I am closing this PR since it is very outdated, does not compile according to your comment. The mp_toradix function is probably rarely a bottleneck. Please reopen if you disagree.

PR #24, opened 2014-05-11
added Android.mk makefile

tux-mind:
added an Android.mk file that allow building libtommath for the [bionic libc](http://en.wikipedia.org/wiki/Bionic_%28software%29) using the android NDK.


minad:
@sjaeckel I suggest to close this PR since we should rather simplify the build system and test suite instead of adding more to it.

PR #37, opened 2015-06-01
Fix Mingw x64 configuration

jzmaddock:
On Mingw64 long is a 32-bit type even though pointers are 64-bit types, as a result while the library builds OK on Mingw64, the functions return garbage as all the internal logic is messed up.  This patch defines mp_digit and mp_word correctly in this situation.

### DIFF
--- a/tommath.h
+++ b/tommath.h
@@ -81,8 +81,13 @@ extern "C" {
    typedef signed long long   long64;
 #endif
 
-   typedef unsigned long long mp_digit;
+#if defined(_WIN32)
+   typedef unsigned long long     mp_digit;
+   typedef unsigned __int128      mp_word;
+#else
+   typedef unsigned long      mp_digit;
    typedef unsigned long      mp_word __attribute__ ((mode(TI)));
+#endif
 
    #define DIGIT_BIT          60
 #else


sebastianas:
There is a lot of software assuming long is 64bit on 64bit architectures. Is mingw64 really that different?

jzmaddock:
> There is a lot of software assuming long is 64bit on 64bit 
> architectures. Is mingw64 really that different?
>
Well then there's a lot of software that's wrong: as that's not the 
convention that Windows uses - int and long are both 32-bit types, long 
long along with size_t and ptrdiff_t are 64-bit types.

IMO you should stick only to what the C std guarantees - and that 
certainly is *not* that sizeof(long) == sizeof(void*).  Ideally you 
would use stdint.h and int64_t etc for this, but I realise that's not 
universally available.  A backup might be to use size_t/ptrdiff_t as 
these are guaranteed to be "pointer sized".

HTH, John.

sjaeckel:
`long` is also 32bit on 64bit linux if you compile with x32 ABI...
@jzmaddock can you please fix that one?

tomstdenis:
Just migrate to stdint.h and be done with.  :-)

TazeTSchnitzel:
Also beware that Linux has the X32 target, which is like the opposite of x64 Windows: 32-bit pointers, 64-bit longs.

sjaeckel:
ty @tomstdenis ;-)
we should probably really do that... and change all the unsigned char, unsigned long... in the API's as well... also for all the other libraries...

tomstdenis:
Any platform that doesn't support stdint.h can be dealt with a user supplied version (it's what we do at my work ...).  

I'd reject this PR and submit one that uses stdint.h only.  (except for the non-standard int128 bit).

jzmaddock:
On 02/06/2015 18:00, tomstdenis wrote:
>
> Any platform that doesn't support stdint.h can be dealt with a user 
> supplied version (it's what we do at my work ...).
>
> I'd reject this PR and submit one that uses stdint.h only. (except for 
> the non-standard int128 bit).
>
Close it.  When I have a chance I'll do a fresh one using stdint.h

PR #38, opened 2015-10-25
some lintings

fperrad:
this set of patches solves problems found with a lint tool (with some [MISRA C  2012](https://en.wikipedia.org/wiki/MISRA_C) rules).

I try to have a commit per kind of problem.

### DIFF
--- a/bn_fast_mp_invmod.c
+++ b/bn_fast_mp_invmod.c
@@ -27,7 +27,7 @@ int fast_mp_invmod (mp_int * a, mp_int * b, mp_int * c)
   int     res, neg;
 
   /* 2. [modified] b must be odd   */
-  if (mp_iseven (b) == 1) {
+  if (mp_iseven (b) == MP_YES) {
     return MP_VAL;
   }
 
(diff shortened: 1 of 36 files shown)


czurnieden:
The commit 1459be0581048effc6b71b1370d0c1002dea8638 "fix parenthese when assignment in a condition" is indeed a bug not just an inelegancy. I would recommend merging at least that.

sjaeckel:
Looks nice
Seriously 42a0e6113 ? I think we should either leave it like it was, or really fix it and check the return values, but I won't merge that commit :-)

fperrad:
I purposely put this commit https://github.com/libtom/libtommath/commit/42a0e6113527a27d437226eae562621e25f98198 at the end of the serie, so it's easy to exclude it.

TazeTSchnitzel:
Oh wow, I didn't even know a cast to `(void)` was valid C.

sjaeckel:
I already forgot that under some misrable circumstances you have to write such ugly code :-)

But TBH I like that you brought this up, because it's basically an error that we don't check these return values, so we must go on and fix them instead of casting the problems away.

PR #41, opened 2015-11-25
more private stuff

fperrad:
This series follows the split between `tommath.h` and `tommath_private.h`.

These patches could potentially break some user code, but they clean the public interface of the library.

### DIFF
--- a/demo/demo.c
+++ b/demo/demo.c
@@ -1,3 +1,4 @@
+#include <string.h>
 #include <time.h>
 
 #ifdef IOWNANATHLON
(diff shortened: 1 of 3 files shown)


sjaeckel:
Nice!
Do you have more stuff like that planned?
Just asking so we could try to set another date for a release...

tomstdenis:
Ya I kinda slipped a bit.  I wanted to get something out to jump start the
process again but free time is hard to find these days.  It's not like the
release process is that involved.  I'll review these commits/etc when I get
a chance this week and maybe just make the damn tarballs already?

PR #42, opened 2015-11-26
remove register

fperrad:
let the compiler find the best register optimization


TazeTSchnitzel:
This has been merged now, but I'd be curious to see how this affects performance.

PR #40, opened 2015-11-13
more lintings

fperrad:
this set of patches contains:
- 3 patches against the new file `bn_mp_sqrtmod_prime.c`
- 4 patches against the whole code base

PR #43, opened 2015-11-27
lintings and more

fperrad:
This series contains:
- the PR #40 (6 patches)
- the end of linting (3 patches)
- preparation for using a code formatter

Now, I am ready to use the code formatter [astyle](http://astyle.sourceforge.net/astyle.html) (with https://gist.github.com/fperrad/c9a7f6c457962ee6dba9). But I don't know which is the correct indentation: 2, 3 or 4 spaces.

### DIFF
--- a/bn_error.c
+++ b/bn_error.c
@@ -27,10 +27,10 @@ static const struct {
 /* return a char * string for a given code */
 const char *mp_error_to_string(int code)
 {
-   int x;
+   size_t x;
 
    /* scan the lookup table for the given message */
-   for (x = 0; x < (int)(sizeof(msgs) / sizeof(msgs[0])); x++) {
+   for (x = 0; x < (sizeof(msgs) / sizeof(msgs[0])); x++) {
       if (msgs[x].code == code) {
          return msgs[x].msg;
       }
--- a/bn_fast_mp_invmod.c
+++ b/bn_fast_mp_invmod.c
@@ -59,7 +59,7 @@ int fast_mp_invmod(const mp_int *a, const mp_int *b, mp_int *c)
    if ((res = mp_copy(&y, &v)) != MP_OKAY) {
       goto LBL_ERR;
    }
-   mp_set(&D, 1);
+   mp_set(&D, 1uL);
 
 top:
    /* 4.  while u is even do */
@@ -128,7 +128,7 @@ top:
    /* now a = C, b = D, gcd == g*v */
 
    /* if v != 1 then there is no inverse */
-   if (mp_cmp_d(&v, 1) != MP_EQ) {
+   if (mp_cmp_d(&v, 1uL) != MP_EQ) {
       res = MP_VAL;
       goto LBL_ERR;
    }
--- a/bn_fast_s_mp_sqr.c
+++ b/bn_fast_s_mp_sqr.c
@@ -70,7 +70,7 @@ int fast_s_mp_sqr(const mp_int *a, mp_int *b)
 
       /* execute loop */
       for (iz = 0; iz < iy; iz++) {
-         _W += ((mp_word)*tmpx++)*((mp_word)*tmpy--);
+         _W += (mp_word)*tmpx++ * (mp_word)*tmpy--;
       }
 
       /* double the inner product and add carry */
@@ -78,14 +78,14 @@ int fast_s_mp_sqr(const mp_int *a, mp_int *b)
 
       /* even columns have the square term in them */
       if (((unsigned)ix & 1u) == 0u) {
-         _W += ((mp_word)a->dp[ix>>1])*((mp_word)a->dp[ix>>1]);
+         _W += (mp_word)a->dp[ix>>1] * (mp_word)a->dp[ix>>1];
       }
 
       /* store it */
-      W[ix] = (mp_digit)(_W & MP_MASK);
+      W[ix] = _W & MP_MASK;
 
       /* make next carry */
-      W1 = _W >> ((mp_word)DIGIT_BIT);
+      W1 = _W >> (mp_word)DIGIT_BIT;
    }
 
    /* setup dest */
--- a/bn_fast_s_mp_sqr.c
+++ b/bn_fast_s_mp_sqr.c
@@ -77,7 +77,7 @@ int fast_s_mp_sqr(const mp_int *a, mp_int *b)
       _W = _W + _W + W1;
 
       /* even columns have the square term in them */
-      if ((ix&1) == 0) {
+      if (((unsigned)ix & 1u) == 0u) {
          _W += ((mp_word)a->dp[ix>>1])*((mp_word)a->dp[ix>>1]);
       }
 
--- a/bn_fast_mp_montgomery_reduce.c
+++ b/bn_fast_mp_montgomery_reduce.c
@@ -28,7 +28,7 @@ int fast_mp_montgomery_reduce(mp_int *x, const mp_int *n, mp_digit rho)
    int     ix, res, olduse;
    mp_word W[MP_WARRAY];
 
-   if (x->used > MP_WARRAY) {
+   if (x->used > (int)MP_WARRAY) {
       return MP_VAL;
    }
 
@@ -77,7 +77,7 @@ int fast_mp_montgomery_reduce(mp_int *x, const mp_int *n, mp_digit rho)
        * that W[ix-1] have  the carry cleared (see after the inner loop)
        */
       mp_digit mu;
-      mu = (mp_digit)(((W[ix] & MP_MASK) * rho) & MP_MASK);
+      mu = ((W[ix] & MP_MASK) * rho) & MP_MASK;
 
       /* a = a + mu * m * b**i
        *
@@ -106,12 +106,12 @@ int fast_mp_montgomery_reduce(mp_int *x, const mp_int *n, mp_digit rho)
 
          /* inner loop */
          for (iy = 0; iy < n->used; iy++) {
-            *_W++ += ((mp_word)mu) * ((mp_word)*tmpn++);
+            *_W++ += (mp_word)mu * (mp_word)*tmpn++;
          }
       }
 
       /* now fix carry for next digit, W[ix+1] */
-      W[ix + 1] += W[ix] >> ((mp_word) DIGIT_BIT);
+      W[ix + 1] += W[ix] >> (mp_word)DIGIT_BIT;
    }
 
    /* now we have to propagate the carries and
@@ -131,7 +131,7 @@ int fast_mp_montgomery_reduce(mp_int *x, const mp_int *n, mp_digit rho)
       _W = W + ++ix;
 
       for (; ix <= ((n->used * 2) + 1); ix++) {
-         *_W++ += *_W1++ >> ((mp_word) DIGIT_BIT);
+         *_W++ += *_W1++ >> (mp_word)DIGIT_BIT;
       }
 
       /* copy out, A = A/b**n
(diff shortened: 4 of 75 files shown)


sjaeckel:
I've already added a `.clang-format` style file to libtomcrypt but it's far from perfect, so probably we can give astyle a try... I really don't like how all these auto-formatters treat initialized arrays of structs...
I usually go for 3 spaces (for the libtom projects), but align to the style of the file if it differs.
In private I usually go for the linux knf.
For all I go for a max line length of 120.

fperrad:
I've seen the config file `.clang-format` in `libtomcrypt`, but I can't find how you run it.

sjaeckel:
can you please rebase on develop :-)

fperrad:
rebased:
- the end of linting (3 patches)
- preparation for using a code formatter (8 patches)

sjaeckel:
Regarding clang-format:
In ubuntu there are multiple versions of it provided as clang-format-3.3 up to clang-format-3.6
If you're running windows the clang build provided from http://llvm.org/builds/ says it includes it as well.
After installation it's just a matter of `clang-format -i file.c`

Regarding the PR:
I'll merge when you auto-formatted the sources, okay?
May I also ask you to make sure to include that step somehow in `make new_file` where `tommath_class.h` is re-generated?

Thanks a lot for your effort!

sjaeckel:
When I run `astyle --options=astylerc *.[ch]` most of the files are re-formatted, is that intentional or did I do something wrong?

fperrad:
Yes, because the commit https://github.com/fperrad/libtommath/commit/b1929e08d133e40226851d3013e1e750d1c7dd2c preserves the fact that several indentations (2, 3 and 4 spaces) are used in the project.

So, the current way is:
- edit `indent=spaces=` in [astylerc](https://github.com/fperrad/libtommath/blob/20151127_lint/astylerc#L1) with the value used on  _onefile.c_.
- run `astyle --options=astylerc <onefile>.c`

sjaeckel:
ah damn, so you took me by my word :-)
the 'I usually go for 3 spaces (for the libtom projects), but align to the style of the file if it differs' was meant for manual edits. I think if we automate the entire process now, we should go for one style - so 3 spaces.
could you please fix that up? then it's ready to be merged

czurnieden:
This astylerc does some curious things to bn_mp_mul.c. Is that intended?

sjaeckel:
@czurnieden What do you mean?
```
$ git diff --ignore-all-space  087d2b4cf12b5a1c87235372b8ae5882629655f5..f8f61e7b46e242c7c80df2ef59d8d5a4c871941f -- bn_mp_mul.c
diff --git a/bn_mp_mul.c b/bn_mp_mul.c
index ea53d5e..443eaa4 100644
--- a/bn_mp_mul.c
+++ b/bn_mp_mul.c
@@ -43,9 +43,9 @@ int mp_mul (mp_int * a, mp_int * b, mp_int * c)
       int     digs = a->used + b->used + 1;
 
 #ifdef BN_FAST_S_MP_MUL_DIGS_C
-    if ((digs < MP_WARRAY) &&
+      if ((digs < (int)MP_WARRAY) &&
           (MIN(a->used, b->used) <=
-         (1 << ((CHAR_BIT * sizeof(mp_word)) - (2 * DIGIT_BIT))))) {
+           (int)(1u << (((size_t)CHAR_BIT * sizeof(mp_word)) - (2u * (size_t)DIGIT_BIT))))) {
         res = fast_s_mp_mul_digs(a, b, c, digs);
       } else
 #endif
```

sjaeckel:
But what I don't like is the indentation of `bn_mp_exteuclid.c` and `bn_mp_prime_random_ex.c`
resp. that it inserts a line-break for the goto line, but doesn't put the opening `{` to where it belongs but leaves the spaces...

czurnieden:
@sjaeckel Please look at the full file, the indentation of the "if/else if/else if" that gets interrupted by the preprocessor make a stair. The diff should show nothing in that area. It is more visible if you run it over my version which includes three more (tc4, tc5, and fft). It makes a well visible stair down to the right. I don't know astyle enough to say it's a bug but I nevertheless say it because indent(1) can do it. ;-) 

But on the other side: indent makes some weird things with the indentation of preprocessor directives for legacy (pre ANSI!) reasons which I don't like, so&hellip;

sjaeckel:
@czurnieden mh I think that it's desired like that as the `if-else`'s are nested that deep if all preprocessor defines are enabled ... not sure TBH it's still early :-D

czurnieden:
@sjaeckel Ah, got it, astyle sees it as 

```
if(...){
   ...
} else {
   if(...){
      ...
   } else {
      if(...){
         ...
      } else ...
   }
}
```

Yes, that's technically correct, so it's just a matter of taste which isn't open for discussion: "De gustibus, coloribus et indentibus non est disputandum" ;-)

But as long as I have nothing more to do than running a script before uploading (I can automate that, so I can use my own style for working and upstream style for uploading) I'm ok with everything as long as it doesn't choke indent(1). I just mentioned it because I thought it was not intended, that it might be a bug in astyle.

fperrad:
@sjaeckel with the last commit (step 3) https://github.com/fperrad/libtommath/commit/0229b0ec943da5882acec8e1f423466a6131ace5, all files are now indented with 3 spaces.

fperrad:
Rebased and fixed.
The problem occurs with `MP_8BIT` in [bn_mp_2expt.c](https://github.com/libtom/libtommath/blob/develop/bn_mp_2expt.c#L40).
The bad way is:
```
  a->dp[b / DIGIT_BIT] = (mp_digit)1 << ((mp_digit)b % (mp_digit)DIGIT_BIT);
```
The good way is:
```
   a->dp[b / DIGIT_BIT] = (mp_digit)1 << (mp_digit)(b % DIGIT_BIT);
```

fperrad:
I split the commit "cast refactor" in 2 parts.

sjaeckel:
I just came to the conclusion, after the issue that we recently solved, that I'd prefer to have more complete code coverage before merging this.

TazeTSchnitzel:
re: code style, isn't clang-format the generally preferred formatting tool? Should we not stick with clang-format and just complete the ruleset?

sjaeckel:
I've no opinion.. I proposed clang-format initially, but I also follow "whatever works"...

tomstdenis:
Lots of stuff here that will be pushed to v1.1 ..

fperrad:
rebased on v1.0

sjaeckel:
thx! :+1:

sjaeckel:
okay to merge @tomstdenis ?

fperrad:
rebased

fperrad:
rebased after the merge of #85

fperrad:
rebased after the merge of #88

bn_fast_mp_invmod.c
```diff
@@ -53,7 +53,7 @@ int fast_mp_invmod(const mp_int *a, const mp_int *b, mp_int *c)
    if ((res = mp_copy(&y, &v)) != MP_OKAY) {
       goto LBL_ERR;
    }
-   mp_set(&D, 1);
+   mp_set(&D, (mp_digit)1);
```
sjaeckel: can you please explain what's the purpose of this cast?

this was introduced in "cast mp_digit in call function" (currently 077d5891ad1b0c5c7314fd964b50061824952d66)

shouldn't this cast be implicit!?

bn_error.c
```diff
@@ -27,10 +27,10 @@ static const struct {
 /* return a char * string for a given code */
 const char *mp_error_to_string(int code)
 {
-   int x;
+   size_t x;
 
    /* scan the lookup table for the given message */
-   for (x = 0; x < (int)(sizeof(msgs) / sizeof(msgs[0])); x++) {
+   for (x = 0; x < (sizeof(msgs) / sizeof(msgs[0])); x++) {
```
sjaeckel: I like the changes by "refactor cast" (currently 00209f06ec69c0ed18f70919a87f6c15b3353f2c) a lot

bn_fast_s_mp_sqr.c
```diff
@@ -66,26 +66,26 @@ int fast_s_mp_sqr(const mp_int *a, mp_int *b)
        * we halve the distance since they approach at a rate of 2x
        * and we have to round because odd cases need to be executed
        */
-      iy = MIN(iy, ((ty-tx)+1)>>1);
+      iy = MIN(iy, ((ty-tx)+1)/2);
```
sjaeckel: I'm very unsure if I should like or dislike the "refactor without shift" (currently 60fb7e4c56d81dbce0b2193440c4a170bc8f62da) commit

tomstdenis:
There is value in not using division or remainder where possible... plus if
you can't handle bit twiddling you shouldn't be hacking LTM.

Tom

bn_fast_s_mp_sqr.c
sjaeckel: Just to be clear: I like the idea behind the change!

and this is just a wild guess: can't this fire back for some older/special compilers?

tomstdenis:
Some BSPs don't always have good support for division/etc.  The change
isn't "breaking" anything I just think it's redundant churn.

Tom

bn_fast_mp_invmod.c
fperrad: Yes, currently, the cast is implicit.
The C language has many implicit behaviours, and it is a big issue of C.
With [MISRA](https://en.wikipedia.org/wiki/MISRA_C) rules, all behaviour must be explicit.
The source becomes optimized for human code review (and the compiler generates the same thing).

sjaeckel: ew, I already thought there will be something like MISRA in your reply... if that's the only reason for the uglification I'd vote against this patch
also with the background that I don't know a single person who writes code like that (if he doesn't have to), so all changes that come in won't be in that format and you'd always have to fix it up ;)
@tomstdenis your opinion? :)

tomstdenis:
I find MISRA has a lot of good rules punctuated with some really bad ones
that makes 100% compliance a negative IMHO.

Casting a value that is well within range (and a function parameter to
boot) is pointless.  If it were part of an expression I could see it but
the cast is very explicit since it's being passed to an external reference
(e.g. it has to be cast before being designated (pushed) onto the "stack"
for the function).

Casts like this make more sense for macros or inlined functions.

bn_fast_s_mp_sqr.c
fperrad: The [MISRA](https://en.wikipedia.org/wiki/MISRA_C) rules allow bitwise operations (including shift) only with `unsigned` integers, but not with `signed`.

When the code looks like:
```
  /* now divide in two */
   B = B >> 1;
```
the use of shift is a way of obfuscate the source.

Some files (`bn_mp_exptmod_fast.c`, `bn_s_mp_exptmod.c`) made an intensive use of bitwise operations on `signed`, but I have not dared to refactor them.

tomstdenis:
I haven't looked at LTM code in a long time but unless we compare for <0 or
negatives it might be simpler to change them to unsigned types.

fperrad:
`refactor without shift` was removed.
`cast mp_digit in call function` becomes `literal suffix in call function`.

fperrad:
rebased

sjaeckel:
FYI it looks like one of the patches introduced some warnings in the x32 build

```
$ CC="gcc -mx32" make -j9
...
bn_mp_expt_d_ex.c: In function ‘mp_expt_d_ex’:
bn_mp_expt_d_ex.c:63:24: warning: left shift count >= width of type [-Wshift-count-overflow]
          if ((b & (1uL << (DIGIT_BIT - 1))) != 0u) {
...
```

fperrad:
fixed by PR #94

PR #31, opened 2015-11-29
Jacobi symbol erronous and incomplete, repair and extend?

sjaeckel:
The evaluation of the Jacobi symbol is incorrect at 0/1. It returns 0 instead of 1. ( definition of the null-product)

It is incomplete: the domain of  the numerator of the Jacobi symbol is the whole Z. Negative numerators can be computed with the help of quadratic reciprocity, that is: (n/m) = (-1/n)(n/m). The value of (-1/n) comes from the Legendre symbol (1) and is  (-1)^((n-1)/2). The lowest two bits of n are enough to compute this factor but working with the whole lowest limb might be more legible..

Problem: some users might have built around the old behaviour, hence the question mark.

I stumbled over it because I want to implement the [Baillie-PSW test](http://www.trnicely.net/misc/bpsw.html). One of the functions&mdash;the Lucas-Selfridge test&mdash;needs a full implementation of the Jacobi symbol.

(1) from the congruence  a^((p-1)/2) = (a/p) mod p
see for example: Henri Cohen "A Course in Computational Algebraic Number Theory" of which you might find a copy on the internet.


sjaeckel:
grml damn github creating pull requests based on master
@czurnieden can you please have a look at the last three commits...

sjaeckel:
well that really doesn't require a review...
@czurnieden I merge the fix for (0|1) and close this issue, can you please open a new issue regarding the missing functionality?

czurnieden:
@sjaeckel   took a look at the last three commits of yours but found no obvious error (although a bit unsure about the exact nature of the problem you thought to have). Opened a new issue regarding missing functionality as ordered.

Issue #35, opened 2015-04-24
Please create a new release to help packaging in distros


dod38fr:
Hello

I'm trying to create a Debian package for Perl6. This packaging is currently blocked by a [test failure](https://github.com/perl6/nqp/issues/234). The root cause of this failure is fixed in libtommath by commit 6907f6ca4582064f28e3a58374083ab2a4d85418 (merged).

To provide a proper Perl6 package, this library must be provided by [libtommath Debian package](https://tracker.debian.org/pkg/libtommath) which is at version 0.42.0 (4 years old). 

I'd like to create a new version of libtommath package so that the above fix is available for Debian users (and Perl6 package).

To make this work much easier, could you create a new release of libtommath ?

All the best

baby-gnu:
Hello,

Any news on the v0.43 release ?

Regards.

tomstdenis:
Ugh, the "maintainers" are not keen on a release.  I am the Tom of LibTom ... fuck it.  Barring anything huge let's aim for a November 13th release of v0.43.  The code is fairly stable.

dod38fr:
Good news. Thanks !

sjaeckel:
@tomstdenis now that you started it and brought up a date: what is still missing that we missed the target?
I've marked all relevant issues & PR's to https://github.com/libtom/libtommath/milestones/v0.43.0
Can you please add issues for stuff that should be done I didn't think of yet?

tomstdenis:
Honestly, the only thing that sticks out to me is that some of the CLANG
-m32 tests fail on my F23 box.  I'd like to drop clang from the test suite
(or make it optional).

Beyond that I'd be ok with making tarballs anyday now.

sjaeckel:
For me also gcc -m32 tests fail IIRC

tomstdenis:
running right now -m32 passes with gcc only clang fails.   Though my setup
is a bit weird since I have a nightly build of clang/etc not the F23
install.

Frankly right now I'm ok with leaving it at gcc when testing and we can
circle back to clang/etc later.

tomstdenis:
-mx32 fails for me but I probably don't have X32 libs installed (nor do I
care to right now).

Focus if anything should be on getting LTM 0.43 out and then updates to
TFM/LTC.

sjaeckel:
```
$ COMPILERS=gcc-5 ./testme.sh                                                                                                  
[...]

PR #51, opened 2016-01-31
Remove generated PDF per #35

TazeTSchnitzel:
See: https://github.com/libtom/libtommath/issues/35#issuecomment-177585327


tomstdenis:
Thanks I already pushed that change :-)

TazeTSchnitzel:
:D

PR #56, opened 2016-04-09
When MP_64BIT and sizeof(int) == 4, mp_rand leaves bits untouched in each digit

sjaeckel:
On machines where MP_64BIT is set, sizeof(int) is often still 32-bit, and rand()
returns only 32 random bits.  This leaves 28 bits zeroed in each digit (as DIGIT_BIT ==  60).

This C program will demonstrate the issue:

#include <stdio.h>
#include <stdlib.h>
#include <tommath.h>

$ cat > /tmp/foo.c
void main(void) {

   srandom(time(NULL));
   fprintf(stdout, "%llx %i %i\n",rand(), sizeof(int), DIGIT_BIT);

   mp_int m;
   mp_init(&m);

   mp_rand(&m,10);

   char s[65535];

   mp_tohex(&m, s);

   fprintf(stdout, "%s\n", s);

}
^D
$ gcc -I /home/bri/.local/include/libtommath/ /tmp/foo.c -o /tmp/foo -ltommath
$ /tmp/foo
77426760 4 60
7B62CF7800000006E3339CD000000012E4E3960000000121B7F97000000009274DAA0000000417A629700000001C9CFA06000000032F2D0A600000003F850DC70000000074FAA03

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ script:
 
 branches:
   only:
+    - master
     - develop
 notifications:
   irc: "chat.freenode.net#libtom"
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,7 @@ compiler:
   - gcc
 
 script:
-  - make
-  - make test
-  - make mtest
-  - ./mtest/mtest 666666 | ./test > test.log
+  - make travis_mtest
   - head -n 5 test.log
   - tail -n 2 test.log
   - ./testme.sh
--- a/README.md
+++ b/README.md
@@ -1,4 +1,6 @@
-[![Build Status](https://travis-ci.org/libtom/libtommath.png?branch=develop)](https://travis-ci.org/libtom/libtommath)
+[![Build Status - master](https://travis-ci.org/libtom/libtommath.png?branch=master)](https://travis-ci.org/libtom/libtommath)
+
+[![Build Status - develop](https://travis-ci.org/libtom/libtommath.png?branch=develop)](https://travis-ci.org/libtom/libtommath)
 
 This is the git repository for [LibTomMath](http://www.libtom.org/), a free open source portable number theoretic multiple-precision integer (MPI) library written entirely in C.
 
--- a/bn_mp_add_d.c
+++ b/bn_mp_add_d.c
@@ -49,9 +49,6 @@ mp_add_d (mp_int * a, mp_digit b, mp_int * c)
   /* old number of used digits in c */
   oldused = c->used;
 
-  /* sign always positive */
-  c->sign = MP_ZPOS;
-
   /* source alias */
   tmpa    = a->dp;
 
@@ -95,6 +92,9 @@ mp_add_d (mp_int * a, mp_digit b, mp_int * c)
       */
      ix       = 1;
   }
+
+  /* sign always positive */
+  c->sign = MP_ZPOS;
 
   /* now zero to oldused */
   while (ix++ < oldused) {
--- a/bn_mp_exptmod_fast.c
+++ b/bn_mp_exptmod_fast.c
@@ -150,15 +150,15 @@ int mp_exptmod_fast (mp_int * G, mp_int * X, mp_int * P, mp_int * Y, int redmode
      if ((err = mp_montgomery_calc_normalization (&res, P)) != MP_OKAY) {
        goto LBL_RES;
      }
-#else 
-     err = MP_VAL;
-     goto LBL_RES;
-#endif
 
      /* now set M[1] to G * R mod m */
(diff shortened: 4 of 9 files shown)


czurnieden:
Sorry, took me some time to get gcc-4.9 (should have compiled it myself, would have been faster *grr*)

Nope, same problem.

I'll leave gcc's warning regarding the improper data type in `fprintf`. 

```
$ gcc-4.9 --version
gcc-4.9.real (Ubuntu 4.9.3-8ubuntu2~14.04) 4.9.3
$ cc-4.9  -W -Wall  libtomrandom.c -o libtomrandom -ltommath
libtomrandom.c: In function ‘main’:
libtomrandom.c:57:3: warning: format ‘%llx’ expects argument of type ‘long long unsigned int’, but argument 3 has type ‘int’ [-Wformat=]
   fprintf(stdout, "%llx sizeof(int) = %lx, DIGIT_BIT =  %i\n", rand(), sizeof(int), DIGIT_BIT);
$ ./libtomrandom
6b8b4567 sizeof(int) = 4, DIGIT_BIT =  60
RAND_MAX =  7FFFFFFF
External 600174F18694D66E28113E42A2AF9EE7C8A7DCAF7F4187F26CEF2C8A205ACDDAC0B37B2285AB046181C2334F58CA8A1EDB3C291F3B3F33FB9CA526E4A33BCB465C306FF266694BFEEC6283
600174F18694D66, E28113E42A2AF9E, E7C8A7DCAF7F418, 7F26CEF2C8A205A, CDDAC0B37B2285A, B046181C2334F58, CA8A1EDB3C291F3, B3F33FB9CA526E4, A33BCB465C306FF, 266694BFEEC6283, 
Build-in 327B23C60000000643C9869000000066334873000000074B0DC51000000019495CFF00000002AE8944A0000000625558EC0000000238E1F29000000046E87CCD00000003D1B58BA
327B23C6, 643C9869, 66334873, 74B0DC51, 19495CFF, 2AE8944A, 625558EC, 238E1F29, 46E87CCD, 3D1B58BA,
```
Owwww, great! *sigh*

It is according to the standard. In 7.22.2.1 "The rand function"
> 2 The rand function computes a sequence of pseudo-random integers in the range 0 to
  RAND_MAX
And 
> 5 The value of the RAND_MAX macro shall be at least 32767
The return is `int` and `int` is, as we all know too well,  defined in 5.2.4.2.1 to be at least between -32767 and +32767.

The macro `RAND_MAX` is already defined at libtom's compile time, check for it and utter a warning/error?

```
/* use arc4random on platforms that support it */
#ifdef MP_USE_ALT_RAND
   // I don't have it installed, but AFAIR: it's "int arc4random()", too
    #define MP_GEN_RANDOM()    arc4random()
#else
    #ifdef MP_64BIT
       #if (RAND_MAX < 0xFFFFFFFFFFFFFFF )
          #warning "size of return of rand() function too small!"
       #endif
    #endif
    #define MP_GEN_RANDOM()    rand()
#endif
```

Or include a PRNG into `mp_rand` (Bob's PRNG mentioned above is Public Domain, so at least no license problems) but that's a lot of code (memory) for a small gain.

Or rewrite `mp_rand` to use `rand` more often.

sjaeckel:
> Or rewrite mp_rand to use rand more often.

like in 51c266fe?

czurnieden:
> like in 51c266f?

I would depend the size of the shift-left in line 24 `d <<= 32;` on the size of the return of `MP_GEN_RANDOM()` and put it in a loop (on shift might not suffice).

But I'm picking nits here, I think ;-)

skids:
FWIW, if rand indeed returns something between 0 and 32767 on some system, 4 values with a 15 bit shift would be needed.  For some weird reason it seems real-world (on my system at least) rand was expanded to a 32-bit but was kept positive (returning 0 to 2147483647) so the 32-bit shift would be one too many and leave one bit untouched -- there are only 31 bits of entropy available.  Luckily only 60 are needed as tommath stands now, so just the shift needs to change in the patch.

sjaeckel:
better?

I also removed the call to `abs()` as I don't really get what it was for as `rand()` returns `0..RAND_MAX`

skids:
Definitely better.  Depending how much hassle managing the build-time configuration of those ifdefs becomes, one trick I'd point out is if you want to throw away extra bits, using an xor instead of or is an option.

sjaeckel:
> if you want to throw away extra bits, using an xor instead of or is an option.

what do you exactly mean?

skids:
I mean, if you know, for example, that a rand() function returns at least 30 lsbs of randomness, and you want to fill 60 bits, you can just ((rand() << 30) ^ rand()) | 60_bit_mask and any overlapping bits are still effectively random whether rand() actually returns more bits or not, is all.  What you have will work fine, though, I think.

PR #57, opened 2016-04-21
PR for #56

sjaeckel:
This is a buildable PR for issue #56

### DIFF
--- a/tommath.h
+++ b/tommath.h
@@ -102,8 +102,10 @@ extern "C" {
 /* use arc4random on platforms that support it */
 #ifdef MP_USE_ALT_RAND
     #define MP_GEN_RANDOM()    arc4random()
+    #define MP_GEN_RANDOM_MAX  0xffffffff
 #else
     #define MP_GEN_RANDOM()    rand()
+    #define MP_GEN_RANDOM_MAX  RAND_MAX
 #endif
 
 #define MP_DIGIT_BIT     DIGIT_BIT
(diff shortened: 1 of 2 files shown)


timo:
Is there a reason this couldn't be merged yet? Perl6 is relying on libtommath for its big integer needs and we just stumbled upon this issue again (@skids clearly noticed this quite some time ago already) ...

dod38fr:
Then, a new release of libtommath with these fixes will be welcome by Debian project.

All the best

tommath.h
```diff
@@ -102,8 +102,10 @@ extern "C" {
 /* use arc4random on platforms that support it */
 #ifdef MP_USE_ALT_RAND
     #define MP_GEN_RANDOM()    arc4random()
+    #define MP_GEN_RANDOM_MAX  0xffffffff
```
sjaeckel: @dod38fr I guess this change would require us to increase `current` of the resulting so, right?

If yes, I'd say it'd be better to move that change to `bn_mp_rand.c` so we only have to increase `revision`

dod38fr: I'm going on a limb here... You should increase `current` if a change breaks the ABI and change the behavior. Here, a random number is generated in a different way, but I don't see this change done in an incompatible way. 

As rakudo packager for Debian, I'll make sure that rakudo uses the new version of libtommath at runtime to avoid the bug mentioned by @timo .

All the best

sjaeckel: Mh okay, true. It's ABI compatible even if the header files have additions.

PR #69, opened 2017-01-17
Fix bad detection of 64bit arch on x32

dod38fr:
This commit comes from a [Debian bug 850723](  https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=850723).

The important part is:

> As for libtommath… its build logs do not show this warning, but
> that may be due to a difference in CFLAGS, and it does not have
> such an extensive testsuite as heimdal, so it might also need to
> have this patch applied (and forwarded upstream).
> 
> A better fix *might* be to fix MP_MASK to read something along
> the lines of shifting 1ULL left, but the cast should already do
> that, so I’m unsure why this doesn’t help. Do note that x32 is
> an ILP32 architecture with 64-bit wide CPU registers (but 32-bit
> pointers and longs), so 64-bit mode “should” work (and fast, at
> that) but can be a bit tricky, and that the patch I attached is
> positively known to fix several issues the heimdal testsuite shows.

This PR is a port of the patch proposed by Thorsten. On my side, I will patch Debian's version of libtommath.

All the best

### DIFF
--- a/tommath.h
+++ b/tommath.h
@@ -27,7 +27,7 @@ extern "C" {
 #endif
 
 /* detect 64-bit mode if possible */
-#if defined(__x86_64__)
+#if defined(__x86_64__) && !defined(__ILP32__)
    #if !(defined(MP_32BIT) || defined(MP_16BIT) || defined(MP_8BIT))
       #define MP_64BIT
    #endif


sjaeckel:
IIUC this patch simply compiles ltm in 32bit mode on x32!?!

I'd say this is not a proper patch :-)

How is your testsuite setup & run? (Alternatively, can you test this in case I come up with an alternative patch?)

dod38fr:
Ahem.. I just realized that the tests are not run during build. I'll have to fix that.

Otherwise, you can test this patch by compiling heimdal for x32 arch (which may be done in a chroot)

HTH

brianmay:
> IIUC this patch simply compiles ltm in 32bit mode on x32!?!

What is wrong with that?

dod38fr:
FTR @brianmay is Debian packager for Heimdal

sjaeckel:
1. Can you please help me with the buildlog that is attached to the original bug, I can't find which test exactly fails!? I'd like to try to port it to the ltm tests...

2. Have you tried a more up-to-date version of this library? I build for x32 without any warnings on latest state (develop branch) and the test_standalone build has no errors.

```
$ git describe
v1.0-rc2-28-ge8e8fc4
$ ./testme.sh --with-mx32
uname=Linux j 3.13.0-66-generic #108-Ubuntu SMP Wed Oct 7 15:20:27 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux
gcc=4.8
Run test gcc -mx32 
Tests successful
$ file test
test: ELF 32-bit LSB  executable, x86-64, version 1 (SYSV), dynamically linked (uses shared libs), for GNU/Linux 3.4.0, BuildID[sha1]=3e433ecd32dc6c3502deac80471eaccc8e67f945, not stripped
$ ./test | head -n 5
Digit size 64 Bit 
Size of mp_digit: 8
Size of mp_word: 16
DIGIT_BIT: 60
MP_PREC: 32
...
```

> > IIUC this patch simply compiles ltm in 32bit mode on x32!?!
> 
> What is wrong with that?

3. Performance is wrong... but you're right, as ltm isn't made for performance heimdal could also live with it

brianmay:
I have a suspicion that Heimdal might be using a really old version of this library.

Full history of this directory available here:
https://github.com/heimdal/heimdal/commits/master/lib/hcrypto/libtommath

Hmmm. Might be version 0.41 (2010) + some patches. Even Debian Wheezy has 0.42.

sjaeckel:
This just reminds me of #18 
would it be possible to test heimdal with ltm from the latest release or even better develop?
@jaltman who could do that?

brianmay:
This bug report looks relevant: https://bugzilla.redhat.com/show_bug.cgi?id=1118462

Probably Heimdal needs to be modified to use the system version of libtommath - assuming this is possible - and then we retest.

brianmay:
Also see https://pagure.io/packaging-committee/issue/387

sjaeckel:
Thanks! I searched that one but couldn't find it... now I remember, this has been solved in the branch that was merged here 61d8c8aa42261d96943e211f1c8b231f8203a755

heimdal can now use [`int mp_expt_d_ex(mp_int *a, mp_digit b, mp_int *c, int fast)`](https://github.com/libtom/libtommath/blob/develop/bn_mp_expt_d_ex.c) resp. [`int mp_n_root_ex(mp_int *a, mp_digit b, mp_int *c, int fast)`](https://github.com/libtom/libtommath/blob/develop/bn_mp_n_root_ex.c) with `fast`set to `0`, which will then execute the original (timing invariant) implementation.

brianmay:
Ok, thanks for this. I guess I should open up a ticket with Heimdal's upstream then.

brianmay:
Ok, see heimdal/heimdal#273.

sjaeckel:
Please re-open if you have further questions/comments.

PR #70, opened 2017-02-19
add cmake build system

wheybags:
I want to add this as a package to hunter (https://github.com/ruslo/hunter), and as part of that I need to add a cmake build system.
Since I've done it anyway, I figured you guys might be interested?
Also, I feel I am very qualified to work on this library, as my name is also tom :3

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+build/
+
 # suppress compiler/linker output
 *.[oa]
 *.l[oa]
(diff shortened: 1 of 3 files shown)


minad:
@sjaeckel Same here. tommath is simple enough for vendoring and integrating into other build systems. However I think it is out of scope to maintain many different build systems.

Qix-:
Please reconsider. Makefiles are not great for adding vendored code into larger systems. CMake is an excellent and fairly standard way of supporting a wide variety of compilers/build systems/IDEs/etc.

Not having CMake means I will spend the next hour or so writing an updated CMake configuration just to use this library since asking users to install this system-wide before they can even build the next part of the application.

PR #75, opened 2017-04-19
Tcl fixes

sjaeckel:
Backport of the fixes on https://github.com/tcltk/tcl

This is only part of the fixes, tcl/tk has as well some changes of `mp_int*` API-parameters to `const mp_int*`. Because it's only a minimal subset of the API I didn't include that as I prefer to go all or nothing :)

@kennykb is it fine for you if I include these patches here? (esp. regarding licensing)

Regarding the `mp_iseven()` patch... that one was in early versions like in this PR and changed by @tomstdenis in v0.22 with the comment

> Fixed a bug in both mp_invmod() and fast_mp_invmod() which tested for odd via "mp_iseven() == 0" which is not valid [since zero is not even either].

I also wasn't sure so I asked the interwebs [[1]](http://www.wolframalpha.com/input/?i=is+0+even+or+odd%3F) [[2]](https://en.wikipedia.org/wiki/Parity_of_zero) and the conclusion is that the original version was correct.

### DIFF
--- a/bn_mp_read_radix.c
+++ b/bn_mp_read_radix.c
@@ -71,7 +71,13 @@ int mp_read_radix (mp_int * a, const char *str, int radix)
     }
     ++str;
   }
-  
+
+  /* if an illegal character was found, fail. */
+  if (*str != '\0') {
+      mp_zero(a);
+      return MP_VAL;
+  }
+
   /* set the sign only if a != 0 */
   if (mp_iszero(a) != MP_YES) {
      a->sign = neg;
--- a/bn_mp_read_radix.c
+++ b/bn_mp_read_radix.c
@@ -73,7 +73,7 @@ int mp_read_radix (mp_int * a, const char *str, int radix)
   }
 
   /* if an illegal character was found, fail. */
-  if (*str != '\0') {
+  if (!(*str == '\0' || *str == '\r' || *str == '\n')) {
       mp_zero(a);
       return MP_VAL;
   }
--- a/tommath.h
+++ b/tommath.h
@@ -203,7 +203,7 @@ int mp_init_size(mp_int *a, int size);
 
 /* ---> Basic Manipulations <--- */
 #define mp_iszero(a) (((a)->used == 0) ? MP_YES : MP_NO)
-#define mp_iseven(a) ((((a)->used > 0) && (((a)->dp[0] & 1u) == 0u)) ? MP_YES : MP_NO)
+#define mp_iseven(a) ((((a)->used == 0) || (((a)->dp[0] & 1u) == 0u)) ? MP_YES : MP_NO)
 #define mp_isodd(a)  ((((a)->used > 0) && (((a)->dp[0] & 1u) == 1u)) ? MP_YES : MP_NO)
 #define mp_isneg(a)  (((a)->sign != MP_ZPOS) ? MP_YES : MP_NO)
 


kennykb:
On 04/19/2017 10:24 AM, Steffen Jaeckel wrote:
>
> Backport of the fixes on https://github.com/tcltk/tcl
>
> This is only part of the fixes, tcl/tk has as well some changes of 
> |mp_int*| API-parameters to |const mp_int*|. Because it's only a 
> minimal subset of the API I didn't include that as I prefer to go all 
> or nothing :)
>
> @kennykb <https://github.com/kennykb> is it fine for you if I include 
> these patches here? (esp. regarding licensing)
>
> Regarding the |mp_iseven()| patch... that one was in early versions 
> like in this PR and changed by @tomstdenis 
> <https://github.com/tomstdenis> in v0.22 with the comment
>
>     Fixed a bug in both mp_invmod() and fast_mp_invmod() which tested
>     for odd via "mp_iseven() == 0" which is not valid [since zero is
>     not even either].
>
> I also wasn't sure so I asked the interwebs [1] 
> <http://www.wolframalpha.com/input/?i=is+0+even+or+odd%3F> [2] 
> <https://en.wikipedia.org/wiki/Parity_of_zero> and the conclusion is 
> that the original version was correct.
>
> ------------------------------------------------------------------------
>
Do please backport, by all means!

This is of the particular variety of 'fine' called 'high time.' Tom 
rejected the patches back in the day, so we in Tcl just forked. It'll be 
a relief if we can land the fork after so many years.

Thanks for the attention!

-- 
73 de ke9tv/2, Kevin

sjaeckel:
> Tom rejected the patches back in the day, so we in Tcl just forked.

regarding the `mp_iseven()` I don't know why, but for the `mp_read_radix()` I understand (now)... we always try to be as tolerant as possible for input, but I see the case for this patch.

would this also be acceptable?

```
diff --git a/bn_mp_read_radix.c b/bn_mp_read_radix.c
index 6cae99b..608048f 100644
--- a/bn_mp_read_radix.c
+++ b/bn_mp_read_radix.c
@@ -73,7 +73,7 @@ int mp_read_radix (mp_int * a, const char *str, int radix)
   }
 
   /* if an illegal character was found, fail. */
-  if (*str != '\0') {
+  if (!(*str == '\0' || *str == '\r' || *str == '\n')) {
       mp_zero(a);
       return MP_VAL;
   }
```

sjaeckel:
@kennykb \*ping\*

sjaeckel:
@kennykb \*ping\*

sjaeckel:
@nijtmans are you probably the right person to ask?

nijtmans:
@sjaeckel I fully agree with Kenny's remarks: Thank you very much for your attention! I also would like those changes to be adopted upstream. All Tcl code is BSD-licensed, so there is no restriction at all.

However, since Tom is ignoring our arguments for so many years, I'm not so hopeful that we (= Tcl) can get rid of our fork: the value 0 should be considered even, anything else is not acceptable for Tcl. Sorry Tom ;-)

sjaeckel:
As you can see in this PR I fully agree that 0 should be considered even ;-)
my only open question is if you could live with the version of `mp_read_radix()` proposed in https://github.com/libtom/libtommath/pull/75#issuecomment-295683414
Then there's only your API changes regarding `const` pointers left (which we can probably at one time also include)

nijtmans:
It appears that the read_radix change doesn't influence Tcl: see:  [http://core.tcl.tk/tcl/info/f51addad6e8912b2](url)

sjaeckel:
There will be a new release soon where this will be included.

Issue #60, opened 2016-05-15
error in fast_s_mp_mul_digs


dmitry-lipetsk:
```
  c->used = pa;

  {
    mp_digit *tmpc;
    tmpc = c->dp;
    for (ix = 0; ix < (pa + 1); ix++) {
      /* now extract the previous digit [below the carry] */
      *tmpc++ = W[ix];
    }

```

Should be:
`for (ix = 0; ix < pa ; ix++) {
`

czurnieden:
If I understand it correctly: the pointer `tmpc` needs to be put above `pa` otherwise the following loop would overwrite the last digit in `c` with a zero which is not what the author intended.

dmitry-lipetsk:
>> for (ix = 0; ix < (pa + 1); ix++) 

> following loop would overwrite the last digit in c with a zero

Yes.

Or (as I undestand) can corrupt the memory heap.

ralphgorin:
I agree with the original  comment by dmitry-lipetsk.

The array W contains valid elements from W[0] through W[pa-1].  These pa items need to be copied to c.

As written, the loop accesses c[0] through c[pa].  But c may have only pa elements allocated: the reference to c[pa] is an error. 

With the proposed correction, the loop that zeros unused portions of c would NOT overwrite c[pa-1].  The use of tmpc++ in the corrected loop advances tmpc to point to c[pa].  This is the right place at which to start zeroing any of c that remains.

czurnieden:
It is a bit more complicated than that.

The number `digs` gets the value  `a->used + b->used + 1` in the file `bn_mp_mul.c`.
If there is not enough memory allocated in `c` it gets the amount of `digs` elements (`mp_digit`) allocated.
`pa` on the other side gets the value of `MIN(digs, a->used + b->used)`. That means that `pa` is always one element smaller than there is memory in `c` **iff** called from `mp_mul()`.

The last increment of the pointer `tmpc` in the loop in question will point outside of `c->dp` if `c->alloc` is smaller than  `a->used + b->used + 2`.  That is a problem, indeed.

The first test in `fast_s_mp_mul_digs ()`  is `if (c->alloc < digs)` and `mp_grow()` allocates at least `MP_PREC` elements more than asked for. That means that the problem only exists if `c->alloc`  equals `digs` exactly. Simplest solution: just change that test to `if (c->alloc <= digs)` (or `if (c->alloc < (digs +1))` to keep the style) to ensure enough memory. That won't change the programms logic.

[...]

PR #80, opened 2017-07-05
W[MP_WARRAY] should be cleared before usage.

przemysx:
the W[MP_WARRAY] should be cleared before use.

### DIFF
--- a/bn_fast_s_mp_mul_digs.c
+++ b/bn_fast_s_mp_mul_digs.c
@@ -1,4 +1,5 @@
 #include <tommath_private.h>
+#include <string.h>
 #ifdef BN_FAST_S_MP_MUL_DIGS_C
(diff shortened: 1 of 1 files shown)


czurnieden:
There are, under some circumstances, an shortage in the allocation of `c` by one element and an indeterminate value at `W[pa]`. The first problem can be helped with a change of line 45 `if (c->alloc < digs) {` to `if (c->alloc <= digs) {` and the second one by skipping `W[pa]` in the loop at line 94.  Or by setting `W[pa]` to zero before that loop but a clean-up of the last two loops at lines 94ff. seems to be more elegant, at least to me.

This edit seems to relate to issue 60 (https://github.com/libtom/libtommath/issues/60) although no user with the name "przemysx" participated there.

sjaeckel:
Why should it be cleared before usage?

sjaeckel:
Fixed via #90

PR #94, opened 2017-10-18
restore previous cast

fperrad:
avoid warning with gcc -mx32
    left shift count >= width of type [-Wshift-count-overflow]


sjaeckel:
```
$ git co pr/94                                                      
Switched to branch 'pr/94'
Your branch is up-to-date with 'origin/pr/94'.
$ CC="gcc -mx32" make -j9 test_standalone >/dev/null 
$ ./test                                                            
...
Testing: mp_sqrt
^C   0
```
...and there it hangs...

fperrad:
fixed with more restore

PR #105, opened 2018-04-05
Fix #103

### LINKED ISSUE #103: MP_GEN_RANDOM() calling rand() might be insecure
https://github.com/libtom/libtommath/blob/develop/tommath.h#L113
On non-BSD platforms MP_GEN_RANDOM() calls rand(). It's used by mp_rand() which is in turn used by RSA blinding. Is that secure? 

It would probably be best for a crypto library should totally avoid calling rand() anywhere - it would be easy for someone to assume mp_rand() has good quality entropy.
It also raises a warning in Coverity.

sjaeckel:
I've re-factored the way `mp_rand()` gets its random values.

There are now 4 big different cases

1. `arc4random()` for *BSD
2. the Windows CSP
3. `getrandom()` for linux with new glibc 
4. reading from `/dev/urandom` for linux with old glibc and all other platforms

Some of the code is based on `rng_get_bytes()` of libtomcrypt.

This fixes #103

### DIFF
--- a/bn_mp_rand.c
+++ b/bn_mp_rand.c
@@ -13,34 +13,171 @@
  * guarantee it works.
  */
 
-#if defined(MP_8BIT) || defined(MP_16BIT)
-#define MP_GEN_RANDOM_SHIFT  DIGIT_BIT
-#else
-#if MP_GEN_RANDOM_MAX == 0xffffffffu
-#define MP_GEN_RANDOM_SHIFT  32
-#elif MP_GEN_RANDOM_MAX == 32767
-/* SHRT_MAX */
-#define MP_GEN_RANDOM_SHIFT  15
-#elif MP_GEN_RANDOM_MAX == 2147483647
-/* INT_MAX */
-#define MP_GEN_RANDOM_SHIFT  31
-#elif !defined(MP_GEN_RANDOM_SHIFT)
-#error Thou shalt define their own valid MP_GEN_RANDOM_SHIFT
-#endif
-#endif
+/* First the OS-specific special cases
+ * - *BSD
+ * - Windows
+ */
+#if defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__NetBSD__) || defined(__DragonFly__)
+#define MP_ARC4RANDOM
+#define MP_GEN_RANDOM_MAX     0xffffffffu
+#define MP_GEN_RANDOM_SHIFT   32
 
-/* makes a pseudo-random int of a given size */
-static mp_digit s_gen_random(void)
+static int s_read_arc4random(mp_digit* p)
 {
    mp_digit d = 0, msk = 0;
    do {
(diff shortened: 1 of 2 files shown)


sjaeckel:
@jnthn @timo can you please test this on @MoarVM where the `arc4random()` support originates from?

sjaeckel:
I've just tested this on Windows and it's horribly slow... it's a bit better if I change the `hProv` to be static and only do `CryptAcquireContext()` once...

bn_mp_rand.c
```diff
@@ -13,34 +13,156 @@
...
+
+   if (CryptGenRandom(hProv, sizeof(*p), p) == TRUE) {
+      ret = MP_OKAY;
+   }
+   CryptReleaseContext(hProv, 0);
+   return ret;
+}
+#endif /* WIN32 */
+
+#if !defined(MP_WIN_CSP) && defined(__linux__) && defined(__GLIBC_PREREQ) && __GLIBC_PREREQ(2, 25)
```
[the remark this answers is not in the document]
sjaeckel: Yeah, I've seen that as well on windows but I'm busy right now, will update later.

bn_mp_rand.c
```diff
@@ -13,34 +13,168 @@
...
+#define MP_GETRANDOM
+#include <sys/random.h>
+#include <errno.h>
+
+static int s_read_getrandom(mp_digit* p)
+{
+   int ret;
+   do {
+      ret = getrandom(p, sizeof(*p), 0);
+   } while((ret == -1) && (errno == EINTR));
```
sebastianas: -   return ret;
+   if (ret == sizeof(*p))
+          return MP_OKAY;
+   return -1;

otherwise you do this and fallback to /dev/urandom because getrandom returns the number of random bytes.

bn_mp_rand.c
```diff
@@ -13,34 +13,168 @@
...
+
+static int s_read_dev_urandom(mp_digit* p)
+{
+   ssize_t r;
+   int fd;
+   do {
+      fd = open(MP_DEV_URANDOM, O_RDONLY);
+   } while((fd == -1) && (errno == EINTR));
+   if (fd == -1) return -1;
+   r = read(fd, p, sizeof(*p));
```
sebastianas: this read() needs a do while loop which will retry for errno==EINTR just like open.

sebastianas: yes, I think this is good now.

sjaeckel: done

bn_mp_rand.c
sjaeckel: done

sjaeckel:
@mkj can you please re-review?

mkj:
Sorry won't be able to look at this until Saturday

sjaeckel:
Perfectly fine!

PR #106, opened 2018-04-10
Fix build on Solaris by using ERR_ instead of ERR

samcv:
libtommath won't build on Solaris 11.3 with a goto statement named
`ERR`. Use ERR_ instead so it doesn't conflict.


samcv:
@sjaeckel sounds good. I wasn't sure what to name it, but figured I would make a PR and you'd tell me if you had a more suitable name :)

PR #108, opened 2018-04-12
Make the build logic more robust for BSD systems

ararslan:
This properly sets MAKE on BSDs to `gmake` rather than `make`, which refers to the incompatible BSD Make. Further, it betters detection of Clang as the default compiler, which is the case on FreeBSD 11.0+ and OpenBSD 6.0+, and it fixes the unrecognized call to `arch` on FreeBSD.

### DIFF
--- a/makefile_include.mk
+++ b/makefile_include.mk
@@ -17,15 +17,34 @@ ifndef CROSS_COMPILE
   CROSS_COMPILE=
 endif
 
-ifeq ($(CC),cc)
-  CC = $(CROSS_COMPILE)gcc
+# We only need to go through this dance of determining the right compiler if we're using
+# cross compilation, otherwise $(CC) is fine as-is.
+ifneq (,$(CROSS_COMPILE))
+ifeq ($(origin CC),default)
+CSTR := "\#ifdef __clang__\nCLANG\n\#endif\n"
+ifeq ($(PLATFORM),FreeBSD)
+  # XXX: FreeBSD needs extra escaping for some reason
+  CSTR := $$$(CSTR)
 endif
+ifneq (,$(shell echo $(CSTR) | $(CC) -E - | grep CLANG))
+  CC := $(CROSS_COMPILE)clang
+else
+  CC := $(CROSS_COMPILE)gcc
+endif # Clang
+endif # cc is Make's default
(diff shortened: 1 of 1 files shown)


sjaeckel:
Thanks a lot!

We made sure that `makefile.unix` is Posix/BSD `make` compatible, but that's a nice addition that the original `makefile` and `makefile.shared` can now also be used on these platforms.

Btw. did you try both, `makefile` and `makefile.shared`?

Should we also port these changes to libtomcrypt?

ararslan:
(Sorry, had to fix some syntax. I was working on this between two computers yesterday and messed something up.)

> We made sure that `makefile.unix` is Posix/BSD `make` compatible

I don't see a `makefile.unix` anywhere. Am I missing something? Even the FreeBSD port for LibTomMath uses GNU Make (and has to work around some of the issues I've addressed here).

> Btw. did you try both, `makefile` and `makefile.shared`?

Yes, both `gmake` and `gmake -f makefile.shared` work as expected on FreeBSD with these changes.

> Should we also port these changes to libtomcrypt?

Hm, I've not looked at libtomcrypt before, but it's quite likely that the same set of changes would be beneficial. I can submit a PR there.

sjaeckel:
> (Sorry, had to fix some syntax. I was working on this between two computers yesterday and messed something up.)

No worries

> I don't see a makefile.unix anywhere. Am I missing something? Even the FreeBSD port for LibTomMath uses GNU Make (and has to work around some of the issues I've addressed here).

oops, right, that port of `makefile.icc` to `makefile.unix` was only done for ltc then... sorry :)

> Yes, both gmake and gmake -f makefile.shared work as expected on FreeBSD with these changes.

:+1: 

> Hm, I've not looked at libtomcrypt before, but it's quite likely that the same set of changes would be beneficial. I can submit a PR there.

Cool! That'd be great - thx!

ararslan:
PR to libtomcrypt: https://github.com/libtom/libtomcrypt/pull/382

sjaeckel:
Can you please bring this in sync with the PR in ltc?

ararslan:
Done.

sjaeckel:
Sorry, but can you please rebase :) I'll merge afterwards!

ararslan:
Done!

PR #110, opened 2018-05-02
Do not use tommath.h from the system for building

czurnieden:
The source files include `tommath_private.h` globally (with angle brackets ) which itself does include `tommath.h` globally. The first is not much more than a matter of style but the latter can cause problems if the system `tommath.h` is too different from the header in the packet. I had such a problem with my system header which is so old that it does not use `const` which gave `clang` a little hiccup but it missed the definition of the new file I tried to add and which I had put in the local `tommath.h` but not in the system's `tommath.h`.

It is unlikely that it can cause any grave problems but it is easy to change and should not cause any surprising side effects.

Changing the form of inclusion of the `tommath_*class.h` headers caused too much trouble and had to be dropped, hence the two commits.

### DIFF
--- a/dep.pl
+++ b/dep.pl
@@ -120,8 +120,8 @@ print {$class} << 'EOS';
 #   define LTM_LAST
 #endif
 
-#include "tommath_superclass.h"
-#include "tommath_class.h"
+#include <tommath_superclass.h>
+#include <tommath_class.h>
 #else
 #   define LTM_LAST
 #endif
--- a/dep.pl
+++ b/dep.pl
@@ -45,7 +45,7 @@ EOS
       print {$out} $line;
    } else {
       print {$out} << "EOS";
-#include <tommath.h>
+#include "tommath_private.h"
 #ifdef $define
 $line
 EOS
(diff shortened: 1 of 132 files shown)


sjaeckel:
:+1:

czurnieden:
One of the integration-checks timed-out after 10 minutes. Still OK?

dep.pl
```diff
@@ -45,7 +45,7 @@
       print {$out} $line;
    } else {
       print {$out} << "EOS";
-#include <tommath.h>
+#include "tommath_private.h"
```
sjaeckel: shouldn't this be `#include "tommath.h"`?

sjaeckel: oh, the patch you added is apparently correct :)

thx!

czurnieden: That is the include-line for new files and the new files should get the same treatment as the old files and include `tommath_private.h` instead of `tommath.h`. 

My perl knowledge is a bit rusty and it took me some time to parse through that file but it should be the correct place.

czurnieden: (Why do I have to refresh to see your second comment but the mail came?)
Second try of build 149.7 was successful, merge?

sjaeckel: > (Why do I have to refresh to see your second comment but the mail came?)

good question, probably @github can help? :D

PR #114, opened 2018-05-19
makefile.shared: Respect LIBTOOL.

orbea:
Hi,

I am using the libtool implementation called slibtool which seems to work just fine with libtommath.

https://git.midipix.org/cgit.cgi/slibtool
https://git.midipix.org/cgit.cgi/slibtool/plain/README

The problem is that the standard way to use slibtool is to export `LIBTOOL` in `MAKEFLAGS`.
```
export MAKEFLAGS='LIBTOOL=slibtool'
```
Which will work with almost all autotools builds. However libtommath doesn't respect this variable and wants `LT` instead.

If the `makefile.shared` can use `LIBTOOL` instead of `LT` this would make it a lot easier for anyone that doesn't want the default libtool.

### DIFF
--- a/makefile.shared
+++ b/makefile.shared
@@ -10,14 +10,14 @@ endif
 include makefile_include.mk
 
 
-ifndef LT
+ifndef LIBTOOL
   ifeq ($(PLATFORM), Darwin)
-    LT:=glibtool
+    LIBTOOL:=glibtool
   else
-    LT:=libtool
+    LIBTOOL:=libtool
   endif
 endif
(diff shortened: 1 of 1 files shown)


sjaeckel:
Looks like that's really the "standard way", I didn't find one on the GNU pages but also [FreeBSD](https://docs.freebsd.org/doc/5.2-RELEASE/usr/share/doc/en_US.ISO8859-1/books/porters-handbook/using-automake.html) uses `LIBTOOL`

Would you mind opening a companion PR on https://github.com/libtom/libtomcrypt ?

orbea:
Done, please see the referenced PR. :)

PR #115, opened 2018-05-28
do not use mp_min_u32

karel-m:
The only place where we use `mp_min_u32` is `mp_get_int` but it is IMO more appropriate to use `unsigned long res` instead so that it matches the function's return type.

### DIFF
--- a/bn_mp_get_int.c
+++ b/bn_mp_get_int.c
@@ -17,7 +17,7 @@
 unsigned long mp_get_int(const mp_int *a)
 {
    int i;
-   mp_min_u32 res;
+   unsigned long res;
 
    if (a->used == 0) {
       return 0;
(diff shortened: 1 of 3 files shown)


czurnieden:
Simpler? I'm all for it, but`gcc -mx32` doesn't like it. May need a check if the size of `unsigned long`, which is 32-bit with `-mx`,  corresponds with `DIGIT_BIT` which might be 60 under some circumstances.

karel-m:
I see. I have pushed another improvement.

sjaeckel:
Can you please elaborate on why this is more appropriate than the approach before?

It was cleaner, before the fix you had to do, but now?

karel-m:
OK, you are right except that `mp_min_u32` should be better named as `mp_digit_or_uint32_whichever_is_bigger`

PR #116, opened 2018-05-28
improved detection of MP_64BIT

karel-m:
This patch makes it more explicit that we support `MP_64BIT` only on gcc via `__attribute__((mode(TI)))`.

It also removes a bit misleding `uint128_t` which is AFAIK not a standard - https://stackoverflow.com/questions/34588650/uint128-t-does-not-name-a-type

### DIFF
--- a/tommath.h
+++ b/tommath.h
@@ -37,7 +37,13 @@ extern "C" {
     defined(__ia64) || defined(__ia64__) || defined(__itanium__) || defined(_M_IA64) || \
     defined(__LP64__) || defined(_LP64) || defined(__64BIT__)
 #   if !(defined(MP_32BIT) || defined(MP_16BIT) || defined(MP_8BIT))
-#      define MP_64BIT
+#      if defined(__GNUC__)
+          /* we support 128bit integers only via: __attribute__((mode(TI))) */
+#         define MP_64BIT
+#      else
+          /* otherwise we fall back to MP_32BIT even on 64bit platforms */
+#         define MP_32BIT
+#      endif
 #   endif
 #endif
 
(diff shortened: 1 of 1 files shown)


karel-m:
> Do all compilers with 128bit integer emulation implement the GCC way via mode(TI)?

I do not know but there is IMO no reason to send non-gcc 64bit compilers to blow up at line with `uint128_t` if we can safely turn them into `MP_32BIT` mode

sjaeckel:
> there is IMO no reason to send non-gcc 64bit compilers to blow up

good point, but with that patch there's no more way to enable `MP_64BIT` for those compilers w/o touching the sources. 
I thought we could borrow something like the concept of `__uint32_t_defined` from `stdint.h` and check for something like `__ltm_uint128_t_defined`... but I don't know yet how the integration could work out ...

karel-m:
> there's no more way to enable MP_64BIT for those compilers w/o touching the sources.

I think it was not possible w/o touching the source even before my patch - now it at least does not croak but falls back to `MP_32BIT`.

Or maybe something like:
```
// ...

/* detect 64-bit mode if possible */
#if defined(__x86_64__) || defined(_M_X64) || defined(_M_AMD64) || \
    defined(__powerpc64__) || defined(__ppc64__) || defined(__PPC64__) || \
    defined(__s390x__) || defined(__arch64__) || defined(__aarch64__) || \
    defined(__sparcv9) || defined(__sparc_v9__) || defined(__sparc64__) || \
    defined(__ia64) || defined(__ia64__) || defined(__itanium__) || defined(_M_IA64) || \
    defined(__LP64__) || defined(_LP64) || defined(__64BIT__)
#   if !(defined(MP_32BIT) || defined(MP_16BIT) || defined(MP_8BIT))
#      if defined(__GNUC__)
#         define MP_64BIT
          typedef unsigned long mp_u128 __attribute__((mode(TI)));
#      elif defined(__ANOTHER_COMPILER__)
#         define MP_64BIT
          typedef __int128 mp_u128;
#      else
          /* otherwise we fall back to MP_32BIT even on 64bit platforms */
#         define MP_32BIT
#      endif
#   endif
#endif

// ...

#elif defined(MP_64BIT)
 typedef uint64_t mp_digit;
 typedef mp_u128 mp_word;
 #define DIGIT_BIT 60
#else

// ...
```

karel-m:
> You're right, it's most likely the best way

You mean merging this PR as it is now, right?

sjaeckel:
Yes, please merge as it is now

PR #130, opened 2018-12-02
helper.pl + makefiles cleanup

karel-m:
This PR **does not change any code**.

It introduces `helper.pl` idea as it is used in `libtomcrypt` project.


sjaeckel:
thanks!

PR #135, opened 2018-12-08
Add doc checking to helper.pl

karel-m:
As suggested in another PR I have added a check whether all macros and functions from `tommath.h` are documented in `doc/bn.tex`.

Current status:
```
$ ./helper.pl -d
[missing_doc_for_function] mp_2expt
[missing_doc_for_function] mp_abs
[missing_doc_for_function] mp_addmod
[missing_doc_for_function] mp_clamp
[missing_doc_for_function] mp_cnt_lsb
[missing_doc_for_function] mp_complement
[missing_doc_for_function] mp_copy
[missing_doc_for_function] mp_count_bits
[missing_doc_for_function] mp_div_3
[missing_doc_for_function] mp_dr_is_modulus
[missing_doc_for_function] mp_exch
[missing_doc_for_function] mp_export
[missing_doc_for_function] mp_fread
[missing_doc_for_function] mp_fwrite
[missing_doc_for_function] mp_get_double
[missing_doc_for_function] mp_import
[missing_doc_for_function] mp_is_square
[missing_doc_for_function] mp_mod_2d
[missing_doc_for_function] mp_mulmod
[missing_doc_for_function] mp_n_root_ex
[missing_doc_for_function] mp_rand
[missing_doc_for_function] mp_read_signed_bin
[missing_doc_for_function] mp_reduce_2k_l
[missing_doc_for_function] mp_reduce_2k_setup_l
[missing_doc_for_function] mp_reduce_is_2k
[missing_doc_for_function] mp_reduce_is_2k_l
[missing_doc_for_function] mp_set_double
[missing_doc_for_function] mp_signed_bin_size
[missing_doc_for_function] mp_sqrmod
[missing_doc_for_function] mp_sqrt
[missing_doc_for_function] mp_submod
[missing_doc_for_function] mp_tc_and
[missing_doc_for_function] mp_tc_div_2d
[missing_doc_for_function] mp_tc_or
[missing_doc_for_function] mp_tc_xor
[missing_doc_for_function] mp_to_signed_bin
[missing_doc_for_function] mp_to_signed_bin_n
[missing_doc_for_function] mp_to_unsigned_bin_n
[missing_doc_for_function] mp_toradix_n
[missing_doc_for_function] mp_zero
[missing_doc_for_macro] mp_iseven
[missing_doc_for_macro] mp_isneg
[missing_doc_for_macro] mp_isodd
[missing_doc_for_macro] mp_iszero
[missing_doc_for_macro] mp_mag_size
[...]

PR #136, opened 2018-12-09
Docsupdate

czurnieden:
Added some of the documentation @karel-m 's little perl script found in #135 . Please check for typeos etc. first.

### DIFF
--- a/doc/bn.tex
+++ b/doc/bn.tex
@@ -1,3 +1,13 @@
+\def\fixedpdfdate{D:20181209230255+01'00'}
+\pdfinfo{
+  /CreationDate (\fixedpdfdate)
+  /ModDate (\fixedpdfdate)
+}
+\def\fixedpdfdate{D:20181209230255+01'00'}
+\pdfinfo{
+  /CreationDate (\fixedpdfdate)
+  /ModDate (\fixedpdfdate)
+}
 \documentclass[synpaper]{book}
 \usepackage{hyperref}
 \usepackage{makeidx}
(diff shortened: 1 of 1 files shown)


tommath_class.h
```diff
@@ -1133,12 +1201,8 @@
 #   define LTM_LAST
 #endif
 
-#include "tommath_superclass.h"
-#include "tommath_class.h"
```
karel-m: here you are reverting changes that were agreed in #127

tommath_class.h
```diff
@@ -1133,12 +1201,8 @@
...
+#include <tommath_superclass.h>
+#include <tommath_class.h>
 #else
 #   define LTM_LAST
 #endif
-
-/* ref:         $Format:%D$ */
-/* git commit:  $Format:%H$ */
-/* commit time: $Format:%ai$ */
```
karel-m: missing final set of comments causes `./helper.pl -a` to fail

tommath_class.h
```diff
@@ -1,15 +1,3 @@
...
- * LibTomMath is a library that provides multiple-precision
- * integer arithmetic as well as number theoretic functionality.
- *
- * The library was designed directly after the MPI library by
- * Michael Fromberger but has been written from scratch with
- * additional optimizations in place.
- *
- * The library is free for all purposes without any express
- * guarantee it works.
- */
```
karel-m: missing "standard" initial comment causes `./helper.pl -a` to fail

czurnieden: Sorry, @karel-m my bad, had the wrong base initially. Should be the correct one now.

Issue #76, opened 2017-05-02
Licensing, Public Domain issues


samcv:
I see that libtommath is "dual licensed" under "Public Domain" and "WTFPL".

"WTFPL" may not be legally viable and "Public Domain" is not a license because it is dependent on the country of the software's use. I *think* the creator may be located in Canada? So here are the Canadian rules on Public Domain:

If it was created in Canada, treat it under Canadian law
If it wasn't created in Canada (Berne Convention Country), treat it as if it were created in Canada

It is not to my knowledge allowable to give up your copyright and put it into Public Domain yourself, you must wait 50 years after death, and that is different on different countries. I could be wrong about not being able to give it up under Canadian law, but even so users of the software are still at the mercy of their countries laws regarding Public Domain and Copyright.

May I suggest you choose CC-0 which puts your work into the public domain but also has provisions which take effect if the persons country does not recognize a work as being in public domain. Thereby achieving the desired goal of public domain, but also making users confident that they are able to use the software worldwide without having to do extensive research on Public Domain for their country.

Given that software and the Internet is **Global** it would be awesome if this work could be licensed under CC-0. Thank you all for developing this great software. I ask this only in the hope that a change such as this will allow more usage of this software in the spirit of your intention, of putting it into the Public Domain regardless of a persons country.

[From the CC site](https://creativecommons.org/share-your-work/public-domain/cc0/):

    Dedicating works to the public domain is difficult if not impossible for those wanting to contribute 
    their works for public use before applicable copyright or database protection terms expire. Few if any 
    jurisdictions have a process for doing so easily and reliably. Laws vary from jurisdiction to jurisdiction 
    as to what rights are automatically granted and how and when they expire or may be voluntarily 
    relinquished. More challenging yet, many legal systems effectively prohibit any attempt by these 
    owners to surrender rights automatically conferred by law, particularly moral rights, even when the 
    author wishing to do so is well informed and resolute about doing so and contributing their work to 
    the public domain.

hikari-no-yume:
Was about to suggest CC0.

However, how well does CC0 function for software? The CC licenses, AFAIK, really aren't drafted with source code in mind. It perhaps doesn't matter for this particular license, but perhaps something ultra-liberal (say, MIT) would be better.

samcv:
[...]

PR #138, opened 2018-12-25
Re-license

### LINKED ISSUE #126: please add a BSD2 or MIT style  option for the license
I've been talking with GHC dev collaborators, and we'd like to consider libtommath as a non GPL'd option for ghc's big nat/integer facilities.

to do so, we'd love for you to augment your dual licensing with perhaps a MIT or BSD2 license. 

I realize that wttfyw (wtfpl )/ public domain  dedication *should* be enough, but some people are sticklers, and apparently in some countries the notion of public domain doesn't exist. :(  

i'm happy to help / try  email all the listed contributors if thats something that could be deemed helpful.

Either way this is a fine library and we want to help put it in the hands of more people :)

sjaeckel:
In order to finally be able to close #126 and #76 we had some internal discussions and came to the conclusion that the proposed license is the only one that is acceptable to all maintainers of the libtom projects.

I will (most likely) write a longer news post about reasoning etc. but this should hopefully bring all those endless discussions regarding the PD/WTFPL to an end.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,29 +1,26 @@
-LibTomMath is licensed under DUAL licensing terms.
+                          The LibTom license
 
-Choose and use the license of your needs.
+This is free and unencumbered software released into the public domain.
 
-[LICENSE #1]
+Anyone is free to copy, modify, publish, use, compile, sell, or
+distribute this software, either in source code form or as a compiled
+binary, for any purpose, commercial or non-commercial, and by any
+means.
 
-LibTomMath is public domain.  As should all quality software be.
+In jurisdictions that recognize copyright laws, the author or authors
+of this software dedicate any and all copyright interest in the
+software to the public domain. We make this dedication for the benefit
+of the public at large and to the detriment of our heirs and
+successors. We intend this dedication to be an overt act of
+relinquishment in perpetuity of all present and future rights to this
+software under copyright law.
 
-Tom St Denis
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+IN NO EVENT SHALL THE AUTHORS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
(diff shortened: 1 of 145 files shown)


karel-m:
We should perhaps also replace the initial comment in all source files like this:
```c
/* LibTomMath is free and unencumbered software released into
 * the public domain.
 *
 * For full licence see LICENCE file in the root of LibTomMath 
 * distribution or <http://unlicense.org/>.
 */
```
I can commit these changes to `relicense` branch once the wording is fixed.

karel-m:
It would be also nice to create CONTRIBUTING file in the root saying something like this (original taken from https://unlicense.org/WAIVER):

```
By contributing to LibTomMath project you agree with the following
copyright waiver:

I dedicate any and all copyright interest in this software to the
public domain. I make this dedication for the benefit of the public at
large and to the detriment of my heirs and successors. I intend this
dedication to be an overt act of relinquishment in perpetuity of all
present and future rights to this software under copyright law.

To the best of my knowledge and belief, my contributions are either
originally authored by me or are derived from prior works which I have
verified are also in the public domain and are not subject to claims
of copyright by other parties.

To the best of my knowledge and belief, no individual, business,
organization, government, or other entity has any copyright interest
in my contributions, and I affirm that I will not make contributions
that are otherwise encumbered.
```
And in the pull request template we can add a checkbox like this:

* [ ] I agree with LibtomMath contribution rules described in [CONTRIBUTING](https://github.com/libtom/libtommath/blob/develop/CONTRIBUTING)

sjaeckel:
> We should perhaps also replace the initial comment in all source files like this:

not sure though if it wouldn't be sufficient to replace the last sentence 'The library is free for all purposes without any express guarantee it works.' by something like 'SPDX-License-Identifier: Unlicense'? ...I somehow like the initial comment as it is... but we could also drop it entirely...

> create CONTRIBUTING file

I also thought about it already...

> And in the pull request template we can add a checkbox like this:

oh yeah, I see already how well this will work if I look at the current checkboxes we provide \[/sarcasm\] ;-)

buggywhip:
> On Dec 26, 2018, at 6:34 AM, Steffen Jaeckel <notifications@github.com> wrote:
> 
> not sure though if it wouldn't be sufficient to replace the last sentence 'The library is free for all purposes without any express guarantee it works.' by something like 'SPDX-License-Identifier: Unlicense'? ...I somehow like the initial comment as it is... but we could also drop it entirely...

A brief reminder is not a bad idea so I’d keep as is but drop the word “express”.  ...less is more.  

If you don’t you may need to add “or implied“ and then you are rewriting the license.  Nah. 

You might consider adding “For details see the LICENSE file.”


Sent fm iTouch

PR #139, opened 2018-12-26
some linting

fperrad:
after merge of #113 

the following warning at https://github.com/libtom/libtommath/blob/develop/bn_mp_prime_is_prime.c#L338 is not fixed
```
bn_mp_prime_is_prime.c  338  Warning 534: Ignoring return value of function 'mp_div_2d(const mp_int *, int, mp_int *, mp_int *)'
```

### DIFF
--- a/bn_mp_get_bit.c
+++ b/bn_mp_get_bit.c
@@ -45,7 +45,7 @@ int mp_get_bit(const mp_int *a, int b)
    bit = (mp_digit)(1) << (b % DIGIT_BIT);
 
    isset = a->dp[limb] & bit;
-   return (isset != 0) ? MP_YES : MP_NO;
+   return (isset != 0u) ? MP_YES : MP_NO;
 }
 
 #endif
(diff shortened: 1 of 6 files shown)


czurnieden:
Ouch! Forgetting a check **must** not happen, sorry!

Running `astyle` before saving is automated now in my editor. I hope that slightly embarrassing comment `improve indentation` won't be seen anymore in regards to my code ;-)

So we want *everything* explicit? Yepp, with all of the compilers libtommath needs to work with--it's a good idea.

Thank you very much, @fperrad your work is highly appreciated!

On the other side:
`!= MP_NO` instead of `== MP_YES`?
Really?
Well, it's not wrong, that's true.

But sit back down, it is **not** a formal complaint, just an observation. I'm pretty sure you have good reasons to do it that way (and I can even think of one myself), so please leave it as it is. But it should be noted and put somewhere at the end of our todo-lists to move all bool-type flags to use the `_Bool` type, if available. Sometime. Next year. Or so.

PR #141, opened 2018-12-27
more linting

fperrad:


demo/demo.c
```diff
@@ -745,12 +745,12 @@ int main(void)
 
             if (mp_cmp(&c, &d) != MP_EQ) {
 /* *INDENT-OFF* */
```
sjaeckel: I guess those comments can then go

PR #142, opened 2018-12-28
refactor around macros

fperrad:


minad:
@fperrad I think the MP_YES and MP_NO macros should be removed to modernize the code base. Instead C99 bool should be used. Furthermore C99 supports inline functions, which should be used instead of the macros.

C99 is 20 years ago. It is time.

minad:
@sjaeckel Why?

minad:
Ok, compiler support is always a problem. I just looked at https://en.wikipedia.org/wiki/C99#Implementations. My opinion is still to drop support for old platforms to avoid accumulating cruft No compiler update since > 20 years or so, which I think is very conservative. But the goals of the project might vary...

In particular, what I find problematic is the missing support for continuous integration on legacy platforms. If CI would be in place for every supported platform, e.g., via emulation, I would be much happier.

If you want to stay with C89, concerning the macros in this PR: I don't get the purpose of the the `if (f(...) != MP_NO)` conditionals, which are equivalent to `if (f(...))`. Then there are expressions like `IS_ODD(a) ? MP_YES : MP_NO`, as if `IS_ODD` returns some magic boolean type which must then be converted to the `MP_YES` and `MP_NO` int values. `IS_ODD` already returns 0 or 1.

karel-m:
@minard Do you have any "real life issues" (TM) with the current source code?

I do not want to repeat the way we have introduced stdint.h dependency. It was IMO a change by which we gain nothing but cause "real life issues" (TM) to those outside of the GNU C wonderland (and few others).

Ad "No compiler update since > 20 years or so, which I think is very conservative." Please check the linked wiki page carefully e.g. "Visual C++ 2012 and earlier did not support C99" - as you see it is far less than 20 years.

There simply must be a sounder reason than just "Hey guys it is 2019, let's require C99 feature(s)".

minad:
@karel-m Visual Studio 2012 is not relevant to my argument since there exists VS2015 for those platforms and which is more recent. I do not see a reason to support legacy compilers. It is unfortunate that VS2012 has to count as one, despite being only 7 years old, since they did not manage to support C99 after 13 years. It seems I did not leave without reason for the wonderland after using VS6 for quite a while :)

My argument is readability and avoiding cruft. In particular I am not a fan of those `!= MP_YES` checks which are redundant. Those would look even more redundant with `!= false` in C99. That got me started at first to consider proposing C99.

fperrad:
rebased on top of 1.1.0

there are 2 parts in this PR :
- 3 first commits : remove the internal use of some public macros
- 3 last commits : refactor with new internal macros

PR #158, opened 2019-02-11
disable mk68 double tests for now, see #156

minad:


minad:
@sjaeckel @czurnieden Let's go with the simplest "solution" for now.

Issue #159, opened 2019-02-12
bn_mp_set_double.c does not compile under MacOS plus


buggywhip:
Under MacOS...
```
$ gcc --version
Apple LLVM version 9.0.0 (clang-900.0.39.2)
Target: x86_64-apple-darwin16.7.0
Thread model: posix
```

a simple `make` returns...
```
   * cc bn_mp_set_double.o
bn_mp_set_double.c:55:6: warning: "mp_set_double implementation is only available on platforms with IEEE754 floating point format" [-W#warnings]
#    warning "mp_set_double implementation is only available on platforms with IEEE754 floating point format"
     ^
1 warning generated.
```
...resulting in no binary for mp_set_double().  (mp_get_double() compiles fine.)

---
#### Notes:

 * LTM and LTC will not crash if mp_get_double() and mp_set_double() are not compiled.  They are not called fm anywhere else in LTM or LTC.  ...and only mp_set_double(), demo.c, and mpi.c are conditioned with `#if defined(__STDC_IEC_559__) || defined(__GCC_IEC_559)`

  * Looking at the list of enabled defines in the attached file ( [bn_mp_set_double.i.txt](https://github.com/libtom/libtommath/files/2855908/bn_mp_set_double.i.txt) ), make note of those with FLT, DBL, and LDBL.

  * It appears when Apple's defines `__DBL_HAS_DENORM__`, `__DBL_HAS_INFINITY__`, and `__DBL_HAS_QUIET_NAN__` are true the processor supports IEEE 754/IEC 60559.  Perhaps some combination of these 3 defines could be used to set `__STDC_IEC_559__`? 

  * LTM only implements get/set for DBL.  Apple supports FLT and LDBL as well.  So, for the sake of completeness, perhaps we should also implement mp_get|set_float() and mp_get|set_longdouble()?

  * Finally, out of a concern for some misusing mp_get|set_double(), the doc should provide a strong caution about the possible loss of precision.

  * FWIW, all Macs since 2006 have Intel processors with integrated IEEE-754 compliant FPUs.  ...as do all PPC Macs.  M680x0 Mac FPUs varies.

czurnieden:
Apple! \*sigh\*

[...]

PR #160, opened 2019-02-14
made set_double arch/mem-layout independent

czurnieden:
It is a bit slower (although completely unoptimized yet) but independent of anything but `float.h` and the macros `DBL_EXP_MAX` and `FLT_RADIX` therein. The same code can be used for `float`with minimal changes (just change `DBL_EXP_MAX`  to `FLT_EXP_MAX` ). The same could be said for `long double` but that needs some more tests because of the slightly different format of `long double` and not everyone has it.

It is now independent from the type of the architecture and its memory-layout. The values of DBL_EXP_MAX` and `FLT_RADIX`  could be hardcoded to avoid the inclusion of `float.h` but the existence of `float.h` is a good sign for the existence for the availability of at least some floating point functions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,11 +24,21 @@ Tests are located in `demo/` and can be built in two flavors.
 * `make test` creates a test binary that is intended to be run against `mtest`. `mtest` can be built with `make mtest` and test execution is done like `./mtest/mtest | ./test`. `mtest` is creating test vectors using an alternative MPI library and `test` is consuming these vectors to verify correct behavior of ltm
 * `make test_standalone` creates a stand-alone test binary that executes several test routines.
 
-[Valgrind](http://valgrind.org/) does not support `long double` but does not reset the limits in `float.h` accordingly. Please build `demo.c` with the additional compiler flag `-DLTM_MEMCHECK_VALGRIND`, e.g.:
+[Valgrind](http://valgrind.org/) does not support `long double` but does not reset the limits in `float.h` accordingly.
+Please build `test.c` with the additional compiler flag `-DLTM_MEMCHECK_VALGRIND`, e.g.:
 
-`CFLAGS=" -DLTM_MEMCHECK_VALGRIND "  make test_standalone`
+```
+CFLAGS=" -DLTM_MEMCHECK_VALGRIND " make test_standalone
+````
 
 It will skip the tests for `long double` and prints a reminder to compile and run the tests without that flag to assure numerical correctness.
+
+In case your platform doesn't provide Valgrind at all and the auto-detection fails, please build `test.c` with the additional compiler flag `-DLTM_NO_VALGRIND`, e.g.:
+
+```
+CFLAGS=" -DLTM_NO_VALGRIND " make test_standalone
+````
+
 
 ## Building and Installing
 
--- a/README.md
+++ b/README.md
@@ -24,6 +24,12 @@ Tests are located in `demo/` and can be built in two flavors.
 * `make test` creates a test binary that is intended to be run against `mtest`. `mtest` can be built with `make mtest` and test execution is done like `./mtest/mtest | ./test`. `mtest` is creating test vectors using an alternative MPI library and `test` is consuming these vectors to verify correct behavior of ltm
 * `make test_standalone` creates a stand-alone test binary that executes several test routines.
 
+[Valgrind](http://valgrind.org/) does not support `long double` but does not reset the limits in `float.h` accordingly. Please build `demo.c` with the additional compiler flag `-DLTM_MEMCHECK_VALGRIND`, e.g.:
+
+`CFLAGS=" -DLTM_MEMCHECK_VALGRIND "  make test_standalone`
+
+It will skip the tests for `long double` and prints a reminder to compile and run the tests without that flag to assure numerical correctness.
+
 ## Building and Installing
 
 Building is straightforward for GNU Linux only, the section "Building LibTomMath" in the documentation in `doc/bn.pdf` has the details.
--- a/bn_mp_set_double.c
+++ b/bn_mp_set_double.c
@@ -11,20 +11,82 @@
  *
  * SPDX-License-Identifier: Unlicense
  */
-
 #include <float.h>
-#if ( (defined DBL_MAX_EXP) && (FLT_RADIX == 2) )
+/* Minimum information needed */
+#if ( (defined DBL_MAX) && (FLT_RADIX == 2) )
 
+/* We can use a faster method if we have an IEEE compliant machine and a working stdint.h */
+#if (((defined __STDC_IEC_559__) || (defined __GCC_IEC_559)) && (defined UINT64_MAX))
+
+/* To avoid "magic numbers". Some people don't like them, even if their meaning is obvious */
+/* Bits in mantissa without the implied bit  */
+#define SIGNIFICANT (DBL_MANT_DIG - 1)
+#define BIAS (DBL_MAX_EXP - 1)
+/* Maximal unbiased exponent */
+#define MAX_UNBIASED_EXPONENT ((2*DBL_MAX_EXP) - 1)
+/* Length of a binary64 without sign bit */
+#define PAYLOAD 63
+int mp_set_double(mp_int *a, double b)
+{
+   uint64_t frac;
+   int exp, res;
+   union {
+      double   dbl;
+      uint64_t bits;
+   } cast;
+   cast.dbl = b;
+
+   exp = (int)((unsigned)(cast.bits >> SIGNIFICANT) & (unsigned)MAX_UNBIASED_EXPONENT);
+   frac = (cast.bits & ((1ULL << SIGNIFICANT) - 1ULL)) | (1ULL << SIGNIFICANT);
+
+   if (exp == MAX_UNBIASED_EXPONENT) { /* +-inf, NaN */
+      return MP_VAL;
+   }
+   exp -= BIAS + SIGNIFICANT;
+
+   res = mp_set_long_long(a, frac);
+   if (res != MP_OKAY) {
+      return res;
+   }
+
+   res = (exp < 0) ? mp_div_2d(a, -exp, a, NULL) : mp_mul_2d(a, exp, a);
+   if (res != MP_OKAY) {
+      return res;
+   }
+
+   if (((cast.bits >> PAYLOAD) != 0ULL) && (mp_iszero(a) == MP_NO)) {
+      SIGN(a) = MP_NEG;
+   }
+
+   return MP_OKAY;
+}
+
+#else
 static double s_math_h_less_frexp(double x, int *exp)
 {
-   int exponent = 0;
+   int exponent = 0, i;
+   double high = 2.0;
+   double low = 0.5;
 
-   while (x >= 1.0) {
-      exponent++;
-      if (exponent > DBL_MAX_EXP) {
-         break;
+   if (x >= 1.0) {
+      for (i = 0; x >= high; i++) {
+         exponent += (1 << i);
+         x *= low;
+         low *= low;
+         high *= high;
       }
-      x /= 2.0;
+      if (x < 0.5) {
+        while (x < 0.5) {
+          x *= 2.0;
+          exponent--;
+        }
+      }
+      else {
+         while (x >= 1.0) {
+            x /= 2.0;
+            exponent++;
+         }
+      }
    }
    *exp = exponent;
    return x;
@@ -32,8 +94,10 @@ static double s_math_h_less_frexp(double x, int *exp)
 
 int mp_set_double(mp_int *a, double b)
 {
-   int exp = 0, res, sign = MP_ZPOS;
+   int exp = 0, res = MP_OKAY, sign = MP_ZPOS;
+
    /* Check for NaN */
+   /* TODO: there might be some ompilers who do not return true in case of NaN */
    if (b != b) {
       return MP_VAL;
    }
@@ -42,18 +106,19 @@ int mp_set_double(mp_int *a, double b)
       b = b * (-1.0);
       sign = MP_NEG;
    }
+
+   /* Check for infinity */
+   if ( b > DBL_MAX) {
+      return MP_VAL;
+   }
+
    mp_zero(a);
    /* Numbers smaller than 1 truncate to zero */
-   if (b < 1.0) {
+   if (b < 1.0) {puts("b < 1.0");
       a->sign = sign;
       return MP_OKAY;
    }
-
    b = s_math_h_less_frexp(b, &exp);
-   /* +/-inf if exp > DBL_MAX_EXP */
-   if (exp > DBL_MAX_EXP) {
-      return MP_VAL;
-   }
 
    while (exp-- >= 0) {
       b *= 2.0;
@@ -63,7 +128,6 @@ int mp_set_double(mp_int *a, double b)
          }
          b -= 1.0;
       }
-
       if (exp >= 0) {
          if ((res = mp_mul_2d(a, 1, a)) != MP_OKAY) {
             return res;
@@ -74,20 +138,17 @@ int mp_set_double(mp_int *a, double b)
          break;
       }
    }
-
    if (res != MP_OKAY) {
       return res;
    }
-
    res = (exp < 0) ? mp_div_2d(a, -exp, a, NULL) : mp_mul_2d(a, exp, a);
    if (res != MP_OKAY) {
       return res;
    }
-
    SIGN(a) = sign;
-
-   return MP_OKAY;
+   return res;
 }
+#endif
 #else
 /* pragma message() not supported by several compilers (in mostly older but still used versions) */
 #  ifdef _MSC_VER
@@ -97,6 +158,15 @@ int mp_set_double(mp_int *a, double b)
 #    warning "mp_set_double implementation is only available on platforms with IEEE754 floating point format"
 #    warning "At least DBL_MAX_EXP must be defined and set and, for now, FLT_RADIX must be 2."
 #  endif
+#if (FLT_RADIX == 16)
+#  ifdef _MSC_VER
+#    pragma message("No radices other than two are supported. IBM's z/OS uses IEEE-754 compliant floats")
+#    pragma message("with the compiler option FLOAT(IEEE)")
+#  else
+#    warning "No radices other than two are supported. IBM's z/OS uses IEEE-754 compliant floats"
+#    warning "with the compiler option FLOAT(IEEE)"
+#  endif
+#endif
 #endif
 #endif
 
--- a/bn_mp_set_double.c
+++ b/bn_mp_set_double.c
@@ -16,7 +16,7 @@
 #if ( (defined DBL_MAX) && (FLT_RADIX == 2) )
 
 /* We can use a faster method if we have an IEEE compliant machine and a working stdint.h */
-#if ( (defined __STDC_IEC_559__) && (defined UINT64_MAX) )
+#if ( (defined LTM_NEARLY_IEC_559 ) && (defined UINT64_MAX) )
 int mp_set_double(mp_int *a, double b)
 {
    uint64_t frac;
@@ -135,7 +135,7 @@ int mp_set_double(mp_int *a, double b)
    if (res != MP_OKAY) {
       return res;
    }
-   a-> = sign;
+   a->sign = sign;
    return res;
 }
 #endif
--- a/bn_mp_set_double.c
+++ b/bn_mp_set_double.c
@@ -12,47 +12,83 @@
  * SPDX-License-Identifier: Unlicense
  */
 
-#if defined(__STDC_IEC_559__) || defined(__GCC_IEC_559)
-int mp_set_double(mp_int *a, double b)
+#include <float.h>
+#if ((defined DBL_MAX_EXP) && (FLT_RADIX == 2))
+static double s_math_h_less_frexp(double x, int *exp)
 {
-   uint64_t frac;
-   int exp, res;
-   union {
-      double   dbl;
-      uint64_t bits;
-   } cast;
-   cast.dbl = b;
+  int exponent = 0;
+  while (x >= 1.0) {
+    exponent++;
+    /* exp > DBL_MAX_EXP + 1, it is inf */
+    if (exponent == (DBL_MAX_EXP + 1)) {
+      *exp = exponent;
+      return x;
+    }
+    x /= 2.0;
+  }
 
-   exp = (int)((unsigned)(cast.bits >> 52) & 0x7FFU);
-   frac = (cast.bits & ((1ULL << 52) - 1ULL)) | (1ULL << 52);
+  *exp = exponent;
+  return x;
+}
 
-   if (exp == 0x7FF) { /* +-inf, NaN */
+
+int mp_set_double(mp_int *c, double d)
+{
+   int expnt, res, sign = MP_ZPOS;
+   double frac;
+
+
+   if (d != d) {
       return MP_VAL;
    }
-   exp -= 1023 + 52;
-
-   res = mp_set_long_long(a, frac);
-   if (res != MP_OKAY) {
-      return res;
+   if (d < 0) {
+      d = d * (-1.0);
+      sign = MP_NEG;
+   }
+   /* We do not work on copy, hence clear output memory */
+   mp_zero(c);
+   /* Integers only */
+   if (d < 1.0) {
+      c->sign = sign;
+      return MP_OKAY;
    }
 
-   res = (exp < 0) ? mp_div_2d(a, -exp, a, NULL) : mp_mul_2d(a, exp, a);
-   if (res != MP_OKAY) {
-      return res;
+   frac = s_math_h_less_frexp(d, &expnt);
+   /* +/-inf if exp > DBL_MAX_EXP */
+   if (expnt == (DBL_MAX_EXP + 1)) {
+      return MP_VAL;
    }
 
-   if (((cast.bits >> 63) != 0ULL) && !IS_ZERO(a)) {
-      a->sign = MP_NEG;
+   if (frac == 0.0) {
+      c->sign = sign;
+      return MP_OKAY;
    }
 
+   while (expnt-- >= 0) {
+      frac *= 2.0;
+      if (frac >= 1.0) {
+         if ((res = mp_add_d(c, 1, c)) != MP_OKAY) {
+            return res;
+         }
+         frac -= 1.0;
+      }
+      if (expnt > 0) {
+         if ((res = mp_mul_2d(c, 1, c)) != MP_OKAY) {
+            return res;
+         }
+      }
+   }
+   c->sign = sign;
    return MP_OKAY;
 }
 #else
 /* pragma message() not supported by several compilers (in mostly older but still used versions) */
 #  ifdef _MSC_VER
-#    pragma message("mp_set_double implementation is only available on platforms with IEEE754 floating point format")
+#    pragma message("mp_set_double implementation is only available on platforms with IEEE754 floating point format.")
+#    pragma message("At least DBL_MAX_EXP must be defined and set and, for now, FLT_RADIX must be 2.")
 #  else
 #    warning "mp_set_double implementation is only available on platforms with IEEE754 floating point format"
+#    warning "At least DBL_MAX_EXP must be defined and set and, for now, FLT_RADIX must be 2."
 #  endif
 #endif
 #endif
--- a/bn_mp_set_double.c
+++ b/bn_mp_set_double.c
@@ -16,16 +16,7 @@
 #if ( (defined DBL_MAX) && (FLT_RADIX == 2) )
 
 /* We can use a faster method if we have an IEEE compliant machine and a working stdint.h */
-#if (((defined __STDC_IEC_559__) || (defined __GCC_IEC_559)) && (defined UINT64_MAX))
-
-/* To avoid "magic numbers". Some people don't like them, even if their meaning is obvious */
-/* Bits in mantissa without the implied bit  */
-#define SIGNIFICANT (DBL_MANT_DIG - 1)
-#define BIAS (DBL_MAX_EXP - 1)
-/* Maximal unbiased exponent */
-#define MAX_UNBIASED_EXPONENT ((2*DBL_MAX_EXP) - 1)
-/* Length of a binary64 without sign bit */
-#define PAYLOAD 63
+#if ( (defined __STDC_IEC_559__) && (defined UINT64_MAX) )
 int mp_set_double(mp_int *a, double b)
 {
    uint64_t frac;
@@ -36,13 +27,13 @@ int mp_set_double(mp_int *a, double b)
    } cast;
    cast.dbl = b;
 
-   exp = (int)((unsigned)(cast.bits >> SIGNIFICANT) & (unsigned)MAX_UNBIASED_EXPONENT);
-   frac = (cast.bits & ((1ULL << SIGNIFICANT) - 1ULL)) | (1ULL << SIGNIFICANT);
+   exp = (int)((unsigned)(cast.bits >> (DBL_MANT_DIG - 1)) & (unsigned)((2*DBL_MAX_EXP) - 1));
+   frac = (cast.bits & ((1ULL << (DBL_MANT_DIG - 1)) - 1ULL)) | (1ULL << (DBL_MANT_DIG - 1));
 
-   if (exp == MAX_UNBIASED_EXPONENT) { /* +-inf, NaN */
+   if (exp == ((2*DBL_MAX_EXP) - 1)) { /* +-inf, NaN */
       return MP_VAL;
    }
-   exp -= BIAS + SIGNIFICANT;
+   exp -= (DBL_MAX_EXP - 1) + (DBL_MANT_DIG - 1);
 
    res = mp_set_long_long(a, frac);
    if (res != MP_OKAY) {
@@ -53,9 +44,9 @@ int mp_set_double(mp_int *a, double b)
    if (res != MP_OKAY) {
       return res;
    }
-
-   if (((cast.bits >> PAYLOAD) != 0ULL) && (mp_iszero(a) == MP_NO)) {
-      SIGN(a) = MP_NEG;
+   /* 63 = number of bits without sign-bit in binary64 */
+   if (((cast.bits >> 63) != 0ULL) && (!IS_ZERO(a))) {
+      a->sign = MP_NEG;
    }
 
    return MP_OKAY;
@@ -76,12 +67,11 @@ static double s_math_h_less_frexp(double x, int *exp)
          high *= high;
       }
       if (x < 0.5) {
-        while (x < 0.5) {
-          x *= 2.0;
-          exponent--;
-        }
-      }
-      else {
+         while (x < 0.5) {
+            x *= 2.0;
+            exponent--;
+         }
+      } else {
          while (x >= 1.0) {
             x /= 2.0;
             exponent++;
@@ -108,13 +98,13 @@ int mp_set_double(mp_int *a, double b)
    }
 
    /* Check for infinity */
-   if ( b > DBL_MAX) {
+   if (b > DBL_MAX) {
       return MP_VAL;
    }
 
    mp_zero(a);
    /* Numbers smaller than 1 truncate to zero */
-   if (b < 1.0) {puts("b < 1.0");
+   if (b < 1.0) {
       a->sign = sign;
       return MP_OKAY;
    }
@@ -145,7 +135,7 @@ int mp_set_double(mp_int *a, double b)
    if (res != MP_OKAY) {
       return res;
    }
-   SIGN(a) = sign;
+   a-> = sign;
    return res;
 }
 #endif
--- a/bn_mp_set_double.c
+++ b/bn_mp_set_double.c
@@ -20,7 +20,7 @@
 int mp_set_double(mp_int *a, double b)
 {
    uint64_t frac;
-   int exp, res;
+   int exp, res = MP_OKAY;
    union {
       double   dbl;
       uint64_t bits;
--- a/bn_mp_set_float.c
+++ b/bn_mp_set_float.c
@@ -1,5 +1,5 @@
 #include "tommath_private.h"
-#ifdef BN_MP_SET_DOUBLE_C
+#ifdef BN_MP_SET_FLOAT_C
 /* LibTomMath, multiple-precision integer library -- Tom St Denis
  *
  * LibTomMath is a library that provides multiple-precision
@@ -12,16 +12,18 @@
  * SPDX-License-Identifier: Unlicense
  */
 
-#include <float.h>
-#if ( (defined DBL_MAX_EXP) && (FLT_RADIX == 2) )
 
-static double s_math_h_less_frexp(double x, int *exp)
+
+#include <float.h>
+#if ( (defined FLT_MAX_EXP) && (FLT_RADIX == 2) )
+
+static float s_math_h_less_frexp(float x, int *exp)
 {
    int exponent = 0;
 
    while (x >= 1.0) {
       exponent++;
-      if (exponent > DBL_MAX_EXP) {
+      if (exponent > FLT_MAX_EXP) {
          break;
       }
       x /= 2.0;
@@ -30,7 +32,7 @@ static double s_math_h_less_frexp(double x, int *exp)
    return x;
 }
 
-int mp_set_double(mp_int *a, double b)
+int mp_set_float(mp_int *a, float b)
 {
    int exp = 0, res, sign = MP_ZPOS;
    /* Check for NaN */
@@ -38,30 +40,30 @@ int mp_set_double(mp_int *a, double b)
       return MP_VAL;
    }
 
-   if (b < 0.0) {
-      b = b * (-1.0);
+   if (b < 0.0f) {
+      b = b * (-1.0f);
       sign = MP_NEG;
    }
    mp_zero(a);
    /* Numbers smaller than 1 truncate to zero */
-   if (b < 1.0) {
+   if (b < 1.0f) {
       a->sign = sign;
       return MP_OKAY;
    }
 
    b = s_math_h_less_frexp(b, &exp);
-   /* +/-inf if exp > DBL_MAX_EXP */
-   if (exp > DBL_MAX_EXP) {
+   /* +/-inf if exp > FLT_MAX_EXP */
+   if (exp > FLT_MAX_EXP) {
       return MP_VAL;
    }
 
    while (exp-- >= 0) {
-      b *= 2.0;
-      if (b >= 1.0) {
+      b *= 2.0f;
+      if (b >= 1.0f) {
          if ((res = mp_add_d(a, 1, a)) != MP_OKAY) {
             return res;
          }
-         b -= 1.0;
+         b -= 1.0f;
       }
 
       if (exp >= 0) {
@@ -69,7 +71,7 @@ int mp_set_double(mp_int *a, double b)
             return res;
          }
       }
-      if (b == 0.0) {
+      if (b == 0.0f) {
          exp--;
          break;
       }
@@ -91,15 +93,14 @@ int mp_set_double(mp_int *a, double b)
 #else
 /* pragma message() not supported by several compilers (in mostly older but still used versions) */
 #  ifdef _MSC_VER
-#    pragma message("mp_set_double implementation is only available on platforms with IEEE754 floating point format.")
-#    pragma message("At least DBL_MAX_EXP must be defined and set and, for now, FLT_RADIX must be 2.")
+#    pragma message("mp_set_float implementation is only available on platforms with IEEE754 floating point format.")
+#    pragma message("At least FLT_MAX_EXP must be defined and set and, for now, FLT_RADIX must be 2.")
 #  else
-#    warning "mp_set_double implementation is only available on platforms with IEEE754 floating point format"
-#    warning "At least DBL_MAX_EXP must be defined and set and, for now, FLT_RADIX must be 2."
+#    warning "mp_set_float implementation is only available on platforms with IEEE754 floating point format"
+#    warning "At least FLT_MAX_EXP must be defined and set and, for now, FLT_RADIX must be 2."
 #  endif
 #endif
 #endif
-
-/* ref:         $Format:%D$ */
-/* git commit:  $Format:%H$ */
-/* commit time: $Format:%ai$ */
+/* ref:         \$Format:\%D$ */
+/* git commit:  \$Format:\%H$ */
+/* commit time: \$Format:\%ai$ */
--- a/bn_mp_set_float.c
+++ b/bn_mp_set_float.c
@@ -15,7 +15,7 @@
 #if ( (defined FLT_MAX_EXP) && (FLT_RADIX == 2) )
 
 /* We can use a faster method if we have an IEEE compliant machine and a working stdint.h */
-#if ( (defined __STDC_IEC_559__) && (defined UINT32_MAX) )
+#if ( (defined LTM_NEARLY_IEC_559 ) && (defined UINT32_MAX) )
 int mp_set_float(mp_int *a, float b)
 {
    uint64_t frac;
--- a/bn_mp_set_float.c
+++ b/bn_mp_set_float.c
@@ -13,17 +13,9 @@
  */
 #include <float.h>
 #if ( (defined FLT_MAX_EXP) && (FLT_RADIX == 2) )
-/* We can use a faster method if we have an IEEE compliant machine and a working stdint.h */
-#if (((defined __STDC_IEC_559__) || (defined __GCC_IEC_559)) && (defined UINT32_MAX))
 
-/* To avoid "magic numbers". Some people don't like them, even if their meaning is obvious */
-/* Bits in mantissa without the implied bit  */
-#define SIGNIFICANT (FLT_MANT_DIG - 1)
-#define BIAS (FLT_MAX_EXP - 1)
-/* Maximal unbiased exponent */
-#define MAX_UNBIASED_EXPONENT ((2*FLT_MAX_EXP) - 1)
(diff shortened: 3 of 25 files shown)


minad:
I don't like this implementation since it is probably much slower. This was the reason why I started accessing the floating point bits directly. However I would be interested in some benchmarks. Do you have some measurements? If too slow, I would like to keep the current version available and use it on the platforms where the current version works (e.g. I tested it on x86_64, x86, arm and arm64).

czurnieden:
@minad It is too slow for what exactly?

minad:
@czurnieden For my use case. But at first I would like to know if it is slower and by how much.

minad:
By **too slow** I mean the slowdown to the current version. It is just a matter of what you willing to accept. Factor 2 or 10? But if the slowdown is only marginal and the other function is easier, better to maintain, more portable, why not?

czurnieden:
The while-loop does the work, so it is faster for very small input (< 10.0 or so) but the loop might run up to 1024 times. It is, as I mentioned, completely unoptimized.
You can get rid of it by multiplying `frac` with `2^53` (9007199254740992.0 hence needs an `uint64_t` or something of similar min.size) put it into an `mp_int` and multiply with `expnt` (`mp_mul2d`). I think I did something like that in one of my other commits for `mp_set_double`.

Skips the loop completely but needs memory that can hold `frac * 9007199254740992.0` and a working cast mechanism to get the result in an `uint64_t` (or similar size) variable.

It would than be as fast as your version but still independent from architecture and memory-layout (with the exception re. `uint65_t` casting above).

Want me to do it?

minad:
@czurnieden If you have time, it is fun for you to work on it, sure go on! Maybe you have other more interesting things to hack on ;)

For me the current version works well. But I am also using tommath only on the aforementioned very mainstream architectures (Therefore for me it would also be acceptable to just restrict compilation to a whitelist of supported platforms). My requirements are just that a new version of the function should produce exactly the same outputs as the current one. Furthermore I would prefer not to have a slowdown of more than about 1.5 if we get other advantages in return.

czurnieden:
Knowing what it does in theory does seldom reflect the behaviour in the real world. So I did a littke bit of benchmarking,
```
tests of limits (inf, NaN, a.s.f., neglible)
loop 1000 times
   x_{n+1} = 2x_n
   x_{n+1} = 2x_n - 1
   x_{n+1} = 2x_n - 1 (yes, two times, positive and negative)
end loop
srand(0xdeadbeef)
loop 1.000.000 times
   x = rand_1 * rand_2 + 1 (rand_1 != rand_2)
   x = -x
end loop
```
Mine (as it is, still not optimized at all) is about 5 times slower than yours, 4 times without the recursions (which involve very large numbers, hence more loops).

Even a bit lower than I would have expected but in the range.
An optimized version would probably run close to 1:1.

czurnieden:
Allowing `stdint.h` get's it down to about two times. 

Optimizing my homegrown `frexp` to use a giant-steps/baby-steps like algorithm with tables needs a fixed `DBL_EXP_MAX` which would ruin the flexibility of the code (needs extra code for `float` and `long double` instead of a couple of preprocessor branches) and would get it down to a mere 1.7 times.

Please bear in mind that the actual runtime for all of the 2,906,007 calls is ~2.067 sec., ~1.888 sec,.  and ~1.075 sec. for the `uint64_t` version, the fixed `DBL_EXP_MAX` version and your version respectively. 

Nearly one and a half million calls per second with a single core of an elderly AMD A8-6600K is too slow? Even with the slowest but most portable method you still get over half a million calls per second!

minad:
If you pay too much for every operation you perform, it accumulates quickly. But I agree that this function would hardly become a bottleneck. However I don't see the advantages of replacing the current version with a more portable version. The current function makes assumptions about the float layout, however the function is also simpler than what you are proposing.

minad:
I would not add a long double version since I've not seen much long double usage anywhere. The float versions might be good for platforms which don't support double.

minad:
But concerning the performance issue. I prefer to keep the current version since I have yet to see where it causes real problems.
It works on every platform Debian supports.
The issues we had on Mac are due to the preprocessor and the issue on the mk68 is still there even with this version, right?

The current version is the reasonable implementation for every modern platform.

minad:
current version:

* assumes IEEE754
* there are issues with detection of supported platform in the preprocessor
* simple: accesses the bits, sets the mantissa, shifts by exp, no loops
* fast
* I confirmed that it works on the most common platforms (x86, x86_64, arm, arm64)
* works on all platforms debian supports

version in this PR:

* assumes FLT_RADIX=2 and that DBL_MAX_EXP is defined
* complex: performs all operations in loops, repeated bigint multiplication and addition
* slow (factor 1.7 in your benchmark, worse in the worst case if the loops run fully?)
* should work on all the platforms the current version works too (I have not tested this)

czurnieden:
> It works on every platform Debian supports.

LTM is not restricted to what Debian supports; it needs to support a whole plethora of platforms.

> The issues we had on Mac is due to the preprocessor 

Yepp. Did you patch it?

>  issue on the mk68 is still there even with this version, right?

The reason for this issue was an unexpected, well, forgotten behaviour of the M68K. The patch I offered was denied by you because you decided that my patch, using an algorithm that is not only commonly used in the industry (e.g.: in Boost) but also provable correct for our domain, was not good enough for you.

> assumes IEEE754

LTM needs to support more than that.

> there are issues with detection of supported platform in the preprocessor

Have you patched that?

> simple: accesses the bits, sets the mantissa, shifts by exp, no loops

I wouldn't call a nearly bare-metall operation simple. Loops on the other side get taught to beginners.

> fast

Fast in what relation? A turtle is also fast in relation to a common garden snail as a MIG is fast in relation to a common jetliner.

> I confirmed that it works on the most common platforms (x86, x86_64, arm, arm64)

ARM can switch between big and little endian. Does it work with both?

LTM can run on an S/390 and its relatives e.g.: IBM System z, pSeries, iSeries. Does your version run on S/390 or any of its relatives, too?

Did you check both, AMD-64 and Intel-64?

LTM is small enough to run on most of the bigger MCUs. Most of them don't have a FPU but some have partial support (at least that's what Digikey told me within the allotted 5 minutes of research).

What about softfloat support? You were lucky that Qemu's M86K simulation took Hauser's program which allows for gutting. I do not know if it would work with an actual M68881/M68882 or any other softfloat implementation/compiler.

LTM need to work with a lot of compilers, not all are standard-compliant (As far as I know: the only one with 100% c11 cover is Intel's compiler), some don't even come close, so LTM needs to rely on as little standard compliance as reasonable possible and check for the rest.

> assumes FLT_RADIX=2 and that DBL_MAX_EXP is defined

No, it doesn't *assume* it, it tests for it. You assume it without testing it. `FLT_RADIX` isn't necessarily 2 (IBM 360, for values see e.g.: www.netlib.org/fp/dtoa.c ) and the restriction to `2` is only temporary, I'm already working at it. It's `FLT_RADIX=16` there, a power of 2, so it is not that much work but testing on the S/360 is quite a chore.

> complex: performs all operations in loops, repeated bigint multiplication and addition

I wouldn't call loops complex. The bigint is used to avoid `stdint.h` (some users do not want it and I'm not here to judge). It is easily replaced if you want but it needs a bit more work for `long double`. And it is not that much faster to be honest.

> I would not add a long double version since I've not seen much long double usage anywhere

A user asked for it in #159 and as it was no work to implement it&hellip;
It is quite slow (relative to the others) and needs some touching but it works as it is.

> should work on all the platforms the current version works too

And more.

>  (I have not tested this)

Why have you not tested it?

> The current version is the reasonable implementation for every modern platform.

I would be cautious with "every", you'll never know what is to come, but otherwise: yes, you can say that for modern platforms but LTM is used on old platforms, too. Probably even more so, because modern platforms with a lot of HP under the hood most likely use GMP if they are OK with the licence (I think even Mathematica used GMP), so it is more or less only the license that makes LTM attractive for modern platforms.

LTM is lightweight, and can be made even more so if you don't need all functions, it runs on everything and your kitchen sink, and it is simple, everyone can hack it (even I can ;-) ) but there is one thing that it is not: it is not highly optimized for speed. I have a lot in my backlog but it will need more memory so some may find their way into the main version, some into a separate branch - maybe `needforspeed` or would we get in trouble with EA for that? ;-)

It is also the main reason that I used a branch named `withfloats` and I'm a bit with Tom St Denis here, who did not want floats in LTM in the first place: 'It'd be nice to not have any float code "standard" in LTM.' (in #3 at the end) 

No, I'm sorry, but LTM has to serve a lot of different users with different demands. Changing something is a complicated process and not always welcome by everyone. At least not in the beginning.

minad:
@sjaeckel What do you think?

sjaeckel:
That I'm currently on mobile and only back at home on Sunday :-)

minad:
@czurnieden Sorry, it seems I edited your comment while answering. I tried to revert to the old version manually, since the horrible interface does not support reverting :(

minad:
@czurnieden Thanks for the changes! I am more happy now :)

buggywhip:
> ...Mac ipreprocessor
> 
> ...assumes IEEE754
> 
> LTM needs to support more than that.
> 
> there are issues with detection of supported platform in the preprocessor
> 
[etc.]

...but there is more.   

You all are *assuming* the platform that is running LTM is the one that created the double.  That is NOT always true.  One example: I have had to manipulate floats received in binary files.  The sender being IEEE754 compliant or not was known to me but not my compiler.

To condition the compile based solely of the platform being IEEE754 is incorrect.  There are no FP operations performed so it matters not if the FPU is IEEE754 compliant of not.  Nothing will blow up!

Any conditioning should be based of the state of the data being in the correct format.  Unfortunately you cannot know that until runtime and even then you cannot really know.  Nor can the get/set functions really know the intended format except by arguments like FLT_RADIX and DBL_MAX_EXP to the functions.  

Please, mp_get_double() and mp_set_double() should ALWAYS be compiled (not conditioned), and the functions' signatures should be extended to include the necessary arguments.  (The existing signatures could be transformed to functions that set supported default values before calling the extended signature versions.)

minad:
@buggywhip Yes, we assume that and it is the right thing to do. You are doing something wrong when you store invalid doubles received from an untrusted source inside a C variable of type `double`. I think I said that before. The function receives a `double` as argument and not a byte array.

However what you might want is a function `mp_set_double_bits` which receives as arguments the unpacking as specified by the user. However I think this is out of scope. You could and should write such a function yourself based on the format you are processing. If you wonder why the `mp_set/get_double` functions are in scope then - they are since they are necessary to interact with the C builtin `double` type.

czurnieden:
@buggywhip 
> ...but there is more.

There's always more and always will be, no matter what. ;-)

> There are no FP operations performed so it matters not if the FPU is IEEE754 compliant or not. 

There are none if the FPU is compliant but there are if it is not. There is also the problem of rounding lingering we didn't even touch yet but which can only be handled with a reasonable amount of work if the machine that runs LTM is IEEE-compliant in that regard, or at least offers some method to read the rounding mode.

But that's not what you meant, I think?

> Any conditioning should be based of the state of the data being in the correct format. Unfortunately you cannot know that until runtime and even then you cannot really know.

The conditioning that exist now is just a filter to restrict the input to something the function knows how to handle. That makes the function short and simple. For larger values of "short" and "simple" ;-)

To be able to handle input with the least amount of restriction we need some more information.

Let the input be a floating point number of the form `s * m * r^x` with `s` in `{1,-1}`, and `m, r, x` in **N**  but in an unknown format, maybe just a byte-array because you were talking about a binary file. Then we need the information where `s, m, r, x` are situated, their limits, and the endianness of the byte-array (and a bias for `x` if there is any and also everything else I forgot).

The other function that does it in reverse needs the same information, of course.

So a function that can do it would look like: `mp_set_float(const char *input, struct float_format_t format)` with reasonable defaults and maybe some presets.

That would be general enough to cover most of the cases (Just because I couldn't come up with something that it doesn't cover doesn't mean there isn't anything lurking in the dark waiting for the moment to spread utter despair and hopelessness all over the quivering mass of innocent little programmers&mdash;but I digress) and the user can easily roll their own little wrappers to do their deeds.

At least that's what I think you want from reading:

> Please, mp_get_double() and mp_set_double() should ALWAYS be compiled (not conditioned), and the functions' signatures should be extended to include the necessary arguments. (The existing signatures could be transformed to functions that set supported default values before calling the extended signature versions.)

Not that I wouldn't like the idea, I do, but: who wants to write it? Any volunteers?
\*crickets\*
Yeah, feared that.
\*sigh\*

Yes, I can already hear you saying that you don't need such a broad function but a large general function with a handful of small wrappers is almost always easier to maintain than a handful of medium, closely related functions; and the number "handful" *will* grow in size, no matter what.

I think, in contrast to my colleague, that it still would be a "nice to have", but just that: a "nice to have" with the "nice to have" priority that comes along with it. So: no reason to give up all hope but it might take a while and might not even find its way into LTM at the end.

@minad 
> Ahh. you have that.

Well, it was just C&P and a very small bit of adjustment.
I'm lazy, yes, but not *that* lazy ;-)

> Hmm. I don't see why we need an epsilon test. On compliant platforms the result is exact.

LTM, including the floating point operations, runs well on the M68K where the FPUs have that little quirk but are otherwise compliant enough to be included. And it is *only* for the test and *only* for the M68K which is still in use (it's a tough little cpu that takes a knock or two and is [still available](https://www.mouser.de/Semiconductors/Integrated-Circuits-ICs/Embedded-Processors-Controllers/Microprocessors-MPU/M68000-Core/_/N-a86qz?P=1yzrsew) ) and it checks if `a==b` first. All of the other architectures get a simple `a==b` from the preprocessor instead. It might be an idea to log the cases where `a!=b`, though.
A Q&A at [Stackexchange](https://stackoverflow.com/questions/17663780/is-there-a-document-describing-how-clang-handles-excess-floating-point-precision) gives a bit more details.

The `-ffloat-store` option is not a bad solution for Debian but it is not available at all compilers that are able to compile LTM correctly. 

Oh, and I put the recursions `x_{n+q} = 2x_n - 1.0 ` (with x_0 = 2.0) back in. They fail only in some cases of non-standard floating point implementations but those failures will, hopefully, give us some feedback by way of user complaints and although I'm pretty sure we covered all bases, you'll never know what's out there.

Ah&hellip;what else was there&hellip; the current S/390 OS is z/OS 1.1.0 and the compiler there has a switch to get IEEE-754 compliant floats (but only in Unix-mode?). The code will utter a `#warning` (should it be an `#error` instead?) if it sees `FLT_RADIX == 16` and informs the user about that compiler option. Not yet tested, because testing it would be an \*cough\* herculean task.

I'm currently setting up a 16-bit test environment (only Freedos with qemu) with DJGPP, [BCC](http://edn.embarcadero.com/article/20841), [Mars](https://sourceforge.net/p/freedos/mailman/message/36400642/), and [OpenWatcom](https://open-watcom.github.io/open-watcom/) for some full 16-bit tests. This may take while ;-)
I don't know how the FPU is emulated by qemu but it should be a good test for the floats, too.

Yes, I am aware of the limitations of that project but its better than nothing and, best of all: fun to do!

That was a long one, sorry for that.

minad:
@czurnieden 

> So a function that can do it would look like: `mp_set_float(const char *input, struct float_format_t format)` with reasonable defaults and maybe some presets.

This is similar to the function `mp_set_double_bits` I suggested in my comment to @buggywhip. However I think it is out of scope. This is the job of a serialization library.

> Well, it was just C&P and a very small bit of adjustment.
> I'm lazy, yes, but not _that_ lazy ;-)

You are not the one being lazy here. You are doing the stuff and I am coming and shooting it down ;)
Sorry for that! But I think we have found an acceptable solution now.

> LTM, including the floating point operations, runs well on the M68K where the FPUs have that little quirk but are otherwise compliant enough to be included. And it is _only_ for the test and _only_ for the M68K which is still in use (it's a tough little cpu that takes a knock or two and is [still available](https://www.mouser.de/Semiconductors/Integrated-Circuits-ICs/Embedded-Processors-Controllers/Microprocessors-MPU/M68000-Core/_/N-a86qz?P=1yzrsew) ) and it checks if `a==b` first. All of the other architectures get a simple `a==b` from the preprocessor instead. It might be an idea to log the cases where `a!=b`, though.

Concerning the epsilon test I am ok to use it for M68K but not x86_64 and processors where the result is precise.

Btw, you didn't write about `long double`. Do you want to keep it?

czurnieden:
> Btw, you didn't write about long double. Do you want to keep it?

Wasn't written in less than 5 minutes, so: yes, I want to keep it ;-)

The CPUs supporting it still have quite a large distribution and it makes the thing more complete. I think I add a little measure to check if `sizeof(long double) > 8`and get rid of most of the Arch-checks but I have to check how the quad-precision folks have it implemented (can't check for `sizeof(long double) == 16` because amd64 uses 16 bytes for 80 bit `long double` would have to check mantissa size) but the only one with actual hardware support is Sparc64. Arm64 is the other one with quad support but only as a software implementation. Another weird construct is IBM's extended double which is actually a `double-double` (more precision, not bigger numbers; the exponent is still that of a normal `double`).

So, no quad-precision but I will add a line or two to include more architectures in `mp_set_long_double()`. It makes no sense to deny entry if I can just send them to `mp_set_double()` instead.

> Concerning the epsilon test I am ok to use it for M68K but not x86_64 and processors where the result is precise.

It probably is not only the M68K. See for example the GCC manual for [-mfpmath](https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html). But that needs some more testing. I don't think I have actual i368 hardware *with* an FPU, qemu must do it for now.

bn_mp_set_double.c
```diff
@@ -11,8 +11,21 @@
...
+/* Minimum information needed */
+#if ( (defined DBL_MAX) && (FLT_RADIX == 2) )
 
-#if defined(__STDC_IEC_559__) || defined(__GCC_IEC_559)
+/* We can use a faster method if we have an IEEE compliant machine and a working stdint.h */
+#if (((defined __STDC_IEC_559__) || (defined __GCC_IEC_559)) && (defined UINT64_MAX))
+
+/* To avoid "magic numbers". Some people don't like them, even if their meaning is obvious */
+/* Bits in mantissa without the implied bit  */
+#define SIGNIFICANT (DBL_MANT_DIG - 1)
```
minad: @czurnieden It would be nice to avoid unqualified macros (even if private). I think this is better for clarity but there is also a technical reason. I am using tommath in a amalgamated form embedded in a project. This works perfectly and allows cross module optimization (without fancy LTO). This ease of embedding is one of the reasons why I like tommath so much :)  I haven't looked around the codebase but it would be nice to get rid of all unqualified identifiers. What do you think?

Ping @sjaeckel

bn_mp_set_double.c
```diff
@@ -41,19 +54,119 @@ int mp_set_double(mp_int *a, double b)
...
-      a->sign = MP_NEG;
+   if (((cast.bits >> PAYLOAD) != 0ULL) && (mp_iszero(a) == MP_NO)) {
+      SIGN(a) = MP_NEG;
    }
 
    return MP_OKAY;
 }
+
+#else
+static double s_math_h_less_frexp(double x, int *exp)
```
minad: Is s_ the prefix used for static private functions? I guess I am fine with that.

doc/bn.tex
```diff
@@ -826,6 +826,55 @@ \subsection{Long Long Constants}
 
 This will return the 64 least significant bits of the mp\_int $a$.
 
+\subsection{Floating Point Constants}
```
minad: :+1: thanks for the docs! great!

tommath.h
```diff
@@ -23,6 +23,12 @@
 extern "C" {
 #endif
 
+/* Check for a modern-ish MacOS that supports IEEE-754 floating point numbers */
+#if    ( (defined __APPLE__) && (defined __MACH__) \
+    && ( (defined __DBL_HAS_DENORM__) && ( defined __DBL_HAS_INFINITY__) ) )
+#   define __STDC_IEC_559__ 1
+#endif
```
minad: I would avoid overwriting standard macros. I suggest you define a MP_HAS_NICE_FLOATING_POINT and use it everywhere.

bn_mp_set_double.c
```diff
@@ -11,8 +11,21 @@
  *
  * SPDX-License-Identifier: Unlicense
  */
+#include <float.h>
+/* Minimum information needed */
+#if ( (defined DBL_MAX) && (FLT_RADIX == 2) )
 
-#if defined(__STDC_IEC_559__) || defined(__GCC_IEC_559)
+/* We can use a faster method if we have an IEEE compliant machine and a working stdint.h */
+#if (((defined __STDC_IEC_559__) || (defined __GCC_IEC_559)) && (defined UINT64_MAX))
```
minad: Check for MP_HAS_NICE_FLOATING_POINT instead and move this #if to the header where you do the check for apple.

bn_mp_set_float.c
```diff
@@ -0,0 +1,180 @@
...
+ * The library was designed directly after the MPI library by
+ * Michael Fromberger but has been written from scratch with
+ * additional optimizations in place.
+ *
+ * SPDX-License-Identifier: Unlicense
+ */
+#include <float.h>
+#if ( (defined FLT_MAX_EXP) && (FLT_RADIX == 2) )
+/* We can use a faster method if we have an IEEE compliant machine and a working stdint.h */
+#if (((defined __STDC_IEC_559__) || (defined __GCC_IEC_559)) && (defined UINT32_MAX))
```
minad: MP_HAS_NICE_FLOATING_POINT

bn_mp_set_double.c
```diff
@@ -41,19 +54,119 @@ int mp_set_double(mp_int *a, double b)
       return res;
    }
 
-   if (((cast.bits >> 63) != 0ULL) && !IS_ZERO(a)) {
-      a->sign = MP_NEG;
+   if (((cast.bits >> PAYLOAD) != 0ULL) && (mp_iszero(a) == MP_NO)) {
+      SIGN(a) = MP_NEG;
```
minad: I think @fperrad removed all the usage of the SIGN macro etc. Basically you reverted @fperrad's changes.

bn_mp_set_double.c
```diff
@@ -41,19 +54,119 @@ int mp_set_double(mp_int *a, double b)
...
+      }
+   }
+   if (res != MP_OKAY) {
+      return res;
+   }
+   res = (exp < 0) ? mp_div_2d(a, -exp, a, NULL) : mp_mul_2d(a, exp, a);
+   if (res != MP_OKAY) {
+      return res;
+   }
+   SIGN(a) = sign;
```
minad: SIGN

bn_mp_set_double.c
```diff
@@ -41,19 +54,119 @@ int mp_set_double(mp_int *a, double b)
...
+   }
+
+   /* Check for infinity */
+   if ( b > DBL_MAX) {
+      return MP_VAL;
+   }
+
+   mp_zero(a);
+   /* Numbers smaller than 1 truncate to zero */
+   if (b < 1.0) {puts("b < 1.0");
```
minad: PUTS, please grep for such things

bn_mp_set_float.c
```diff
@@ -0,0 +1,180 @@
...
+
+   res = (exp < 0) ? mp_div_2d(a, -exp, a, NULL) : mp_mul_2d(a, exp, a);
+   if (res != MP_OKAY) {
+      return res;
+   }
+
+   if (((cast.bits >> PAYLOAD) != 0ULL) && (mp_iszero(a) == MP_NO)) {
+      SIGN(a) = MP_NEG;
+   }
+
```
minad: SIGN and mp_iszero

bn_mp_set_float.c
```diff
@@ -0,0 +1,180 @@
...
+   if (res != MP_OKAY) {
+      return res;
+   }
+
+   res = (exp < 0) ? mp_div_2d(a, -exp, a, NULL) : mp_mul_2d(a, exp, a);
+   if (res != MP_OKAY) {
+      return res;
+   }
+
+   SIGN(a) = sign;
```
minad: SIGN

bn_mp_set_long_double.c
```diff
@@ -0,0 +1,263 @@
...
+   if (res != MP_OKAY) {
+      return res;
+   }
+
+   res = (exp < 0) ? mp_div_2d(a, -exp, a, NULL) : mp_mul_2d(a, exp, a);
+   if (res != MP_OKAY) {
+      return res;
+   }
+
+   SIGN(a) = sign;
```
minad: SIGN, please grep not to miss some

bn_mp_set_double.c
czurnieden: I kept the official macros from `float.h` (So instead of using `SIGNIFICANT` I used `(DBL_MANT_DIG - 1)`); `PAYLOAD`  and `EXPONENT_BIT` (in `mp_set_long_double`) have been reset to literals already. If you want *all* of your literals back (for easier c&p) drop me a note.

But I would not really like it, to be honest, the all-literals approach, that is.

bn_mp_set_double.c
czurnieden: Yes, it is. Didn't know it either but got hit over my head for not using it.

doc/bn.tex
czurnieden: Would you do me a favour and check it? I don't want another one of this quite embarrassing discussions, you know which one ;-)

There are still some functions undocumented

tommath.h
czurnieden: Oh, I forgot to check if `__STDC_IEC_559__` is already defined before.

But regarding your actual question: Using a local macro instead of one with a well defined meaning? A set `__STDC_IEC_559__` means that we have IEEE-754 compliance and that's what I want (even though I need only a small subset) and not "a nice floating point".

Moved a bit more into tommath.h but kept `__STDC_IEC_559__`. Can be easily replaced now, so if you still insist on `MP_HAS_NICE_FLOATING_POINT`&hellip;

bn_mp_set_double.c
czurnieden: I thought that, too, was unsure, grepped for it, found a large bunch of&hellip;oh, they are all in mpi.c?

bn_mp_set_double.c
czurnieden: I do, normally, so it can be that I commited the wrong version here, will check. (had a bit of trouble with merging with the HEAD of develop, might be a residue of that fight)

bn_mp_set_float.c
czurnieden: What, `mp_iszero`, too? Can remember that I (or somebody else, can't remember) got hit over my/their head for using `a->used == 0` and now I'm supposed to use it again? Always this rapidly paced changes \*sigh\*

bn_mp_set_double.c
czurnieden: Can do the first, but the check for `uint64_t` needs to stay because of the "without stdint.h" folks. I think I have mentioned them?

bn_mp_set_float.c
minad: I think the current way is to use IS_ZERO :)

tommath.h
minad: My point is: You should not overwrite or defined a macro which is usually defined by the compiler or by libc. And here you are defining `__STDC_IEC_559__`. Instead a custom macro should be used.

czurnieden:
Successfully tested this branch with an architecture that has `double == long double`:

Guest:
```
$ cat dbl_print.c
#include <stdio.h>
#include <stdlib.h>
#include <float.h>

int main (void) {
   printf("%d == %d  %d == %d\n",
           LDBL_MANT_DIG, DBL_MANT_DIG,
           LDBL_MAX_EXP,    DBL_MAX_EXP);
   exit(EXIT_SUCCESS);
}
$ ./dbl_print 
53 == 53  1024 == 1024
$ uname -a
Linux debian 3.16.0-6-versatile #1 Debian 3.16.57-2 (2018-07-14) armv5tejl GNU/Linux
$ lsb_release -a
No LSB modules are available.
Distributor ID: Debian
Description:    Debian GNU/Linux 8.11 (jessie)
Release:        8.11
Codename:       jessie
$ cat /proc/cpuinfo
processor       : 0
model name      : ARM926EJ-S rev 5 (v5l)
BogoMIPS        : 500.22
Features        : swp half thumb fastmult vfp edsp java 
CPU implementer : 0x41
CPU architecture: 5TEJ
CPU variant     : 0x0
CPU part        : 0x926
CPU revision    : 5

Hardware        : ARM-Versatile PB
Revision        : 0000
Serial          : 0000000000000000
```

Host:
```
$ qemu-system-arm --version
QEMU emulator version 3.1.0
Copyright (c) 2003-2018 Fabrice Bellard and the QEMU Project developers
```

(oldstable only, couldn't get Debian 9.8.0 to boot in qemu-system-arm, was too impatient)

README.md
```diff
@@ -23,3 +23,9 @@ The project can be build by using `make`. Along with the usual `make`, `make cle
 Tests are located in `demo/` and can be built in two flavors.
 * `make test` creates a test binary that is intended to be run against `mtest`. `mtest` can be built with `make mtest` and test execution is done like `./mtest/mtest | ./test`. `mtest` is creating test vectors using an alternative MPI library and `test` is consuming these vectors to verify correct behavior of ltm
 * `make test_standalone` creates a stand-alone test binary that executes several test routines.
+
+[Valgrind](http://valgrind.org/) does not support `long double` but does not reset the limits in `float.h` accordingly. Please build `demo.c` with the additional compiler flag `-DLTM_MEMCHECK_VALGRIND`, e.g.:
```
minad: What does that mean? To which instructions translate the long double operations? Are they not present in the valgrind cpu? On x86 or x86_64?

README.md
```diff
@@ -23,3 +23,9 @@ The project can be build by using `make`. Along with the usual `make`, `make cle
...
+
+`CFLAGS=" -DLTM_MEMCHECK_VALGRIND "  make test_standalone`
```
minad: I suggest the has_include check, see other comment.

bn_mp_set_double.c
minad: I've also seen s_mp_*, fast_s_mp_*, bn_reverse, mp_s_*, KARATSUBA_*, TOOM_* in a list of symbols. Maybe we should clean that up at some point.

README.md
czurnieden: Valgrind makes quite a mess with `long double`. Uses `double` instead but doesn't tell anybody, well, can't tell anybody.

README.md
czurnieden: It is not in the C-standard (c-11, too stingy for c18) but in >=C++17. It is supported in clang, GCC (since v5), and  MSVC (see e.g.: https://en.cppreference.com/w/cpp/compiler_support). My native GCC is too old (4.9) and doesn't know it, the version of GCC-5 here is "5.5.0 20171010" and does know it. 

Mmh&hellip;no, I wouldn't use it. At least not for now and not for a single special case that is even for testing only.

README.md
minad: One more argument to just get rid of long double support here.

czurnieden:
Works with SPARC64

``$ cat /proc/cpuinfo 
cpu		: TI UltraSparc IIi (Sabre)
fpu		: UltraSparc IIi integrated FPU
pmu		: ultra12
prom		: OBP 3.10.24 1999/01/01 01:01
type		: sun4u
ncpus probed	: 1
ncpus active	: 1
D$ parity tl1	: 0
I$ parity tl1	: 0
Cpu0ClkTck	: 0000000005f5e100
cpucaps		: flush,stbar,swap,muldiv,v9,mul32,div32,v8plus,vis
MMU Type	: Spitfire
MMU PGSZs	: 8K,64K,512K,4MB
$ uname -a
Linux debiansparc64 4.15.0-2-sparc64 #1 Debian 4.15.11-1 (2018-03-20) sparc64 GNU/Linux
$ gcc -dM -E -x c /dev/null | grep BYTE_ORDER
#define __BYTE_ORDER__ __ORDER_BIG_ENDIAN__
$ gcc -dM -E -x c /dev/null | grep LDBL
#define __LDBL_MAX__ 1.18973149535723176508575932662800702e+4932L
#define __LDBL_MAX_EXP__ 16384
#define __LDBL_HAS_INFINITY__ 1
#define __LDBL_MIN__ 3.36210314311209350626267781732175260e-4932L
#define __LDBL_HAS_QUIET_NAN__ 1
#define __LDBL_HAS_DENORM__ 1
#define __LDBL_DECIMAL_DIG__ 36
#define __LDBL_EPSILON__ 1.92592994438723585305597794258492732e-34L
#define __LDBL_MANT_DIG__ 113
#define __LDBL_MIN_EXP__ (-16381)
#define __LDBL_MAX_10_EXP__ 4932
#define __LDBL_DENORM_MIN__ 6.47517511943802511092443895822764655e-4966L
#define __LDBL_MIN_10_EXP__ (-4931)
#define __LDBL_DIG__ 
```

**Comment by minad** [comment]
Cool! Do you have a sparc or is this a qemu instance? You mentioned qemu a few times before, so I assume that. May I ask you a favor - could you somehow share these qemu images you produced for the different architectures? I know I could produce them myself and I also did that a few times but it still requires a lot of effort imho. This would allow me to also run the test on those - in particular sparc and sparc64 are interesting because of big endian etc. I think there is a website where some guy offers many images for different architectures but they were very outdated.

**Comment by minad** [comment]
Right now I have access to x86, x86_64, arm, arm64 architecture but not more. And none of those is genuinely big endian :( It seems also that debian dropped architectures and there are not many big endians left?

**Comment by czurnieden** [comment]
> Do you have a sparc or is this a qemu instance?

I have an old sparcstation in the cellar but no Sparc64 so this was a qemu session. You can get Ultrasparcs from Ebay quite cheap (250-300€) but since machines with enough HP to run all the machines virtually started to get affordable and qemu grew up enough to actually run these virtual machines, I was finally able to skip the heavy-iron part.

>  could you somehow share these qemu images you produced for the different architectures?

20 Gigs each, not really portable, in both meanings of the word "portable". And you don't want that mess, I can tell you! ;-) I think the only installation that is "clean" enough to show around, is the Sparc64 one. Let's see how it is compressible.

Nuh, still way to much. Let me check if I can build a Sparc64 system as small as reasonable. You don't need no X, do you? (Couldn't it get to work for the Sparc64 anyways ;-) ) Just a boot into `bash` and, if the space allows, all necessary building tools installed?

It was probably not the best idea to use Debian but I know it well and it is available for nearly as many architectures as e.g.: NetBSD and very well documented.

> but it still requires a lot of effort imho

Not much effort anymore, just a bit of Google time and a loooot of patience. I'm installing an arm64 machine in the background and qemu is running with two cores for over 75 minutes now and has just installed the basic system. 105 minutes at the end.

> there are not many big endians left?

It's a bit fluent now, with the all of the bi-endian architectures, but the modern OpenRISC is big-endian (and one of the main reasons I care, tbh); the NXPs Coldfire, a M68K relative; Atmel has one, I think, and the old dinosaurs like IBM's z/arch, of course. Well, still quite a number.

**Comment by minad** [comment]
> > Do you have a sparc or is this a qemu instance?
> 
> I have an old sparcstation in the cellar but no Sparc64 so this was a qemu session. You can get Ultrasparcs from Ebay quite cheap (250-300€) but since machines with enough HP to run all the machines virtually started to get affordable and qemu grew up enough to actually run these virtual machines, I was finally able to skip the heavy-iron part.
> 
> > could you somehow share these qemu images you produced for the different architectures?
> 
> 20 Gigs each, not really portable, in both meanings of the word "portable". And you don't want that mess, I can tell you! ;-) I think the only installation that is "clean" enough to show around, is the Sparc64 one. Let's see how it is compressible.
> 
> Nuh, still way to much. Let me check if I can build a Sparc64 system as small as reasonable. You don't need no X, do you? (Couldn't it get to work for the Sparc64 anyways ;-) ) Just a boot into `bash` and, if the space allows, all necessary building tools installed?
> 
> It was probably not the best idea to use Debian but I know it well and it is available for nearly as many architectures as e.g.: NetBSD and very well documented.
> 
> > but it still requires a lot of effort imho
> 
> Not much effort anymore, just a bit of Google time and a loooot of patience. I'm installing an arm64 machine in the background and qemu is running with two cores for over 75 minutes now and has just installed the basic system. 105 minutes at the end.
> 
> > there are not many big endians left?
> 
> It's a bit fluent now, with the all of the bi-endian architectures, but the modern OpenRISC is big-endian (and one of the main reasons I care, tbh); the NXPs Coldfire, a M68K relative; Atmel has one, I think, and the old dinosaurs like IBM's z/arch, of course. Well, still quite a number.

**Comment by minad** [comment]
> I have an old sparcstation in the cellar but no Sparc64 so this was a qemu session. You can get Ultrasparcs from Ebay quite cheap (250-300€) but since machines with enough HP to run all the machines virtually started to get affordable and qemu grew up enough to actually run these virtual machines, I was finally able to skip the heavy-iron part.

Hehe, while I really like old machines and I considered this, but they also need quite some space. And people look weird if you show them your computer museum ;)

> Nuh, still way to much. Let me check if I can build a Sparc64 system as small as reasonable. You don't need no X, do you? (Couldn't it get to work for the Sparc64 anyways ;-) ) Just a boot into `bash` and, if the space allows, all necessary building tools installed?

Yes, I only need the shell. But please don't spend time on tweaking it. I just thought you may have some images lying around which could be shared without efforts. I always wondered why there are no such prebuilt images. Maybe there are not enough people needing that. Or people who care about cross platform stuff usually have their own images or physical machines. But then you find things like docker images for every ***... why

> It was probably not the best idea to use Debian but I know it well and it is available for nearly as many architectures as e.g.: NetBSD and very well documented.

Hehe, NetBSD. I tried to install it recently somewhere on real hardware, but that didn't work out. To test BSD compatibility I recommend OpenBSD. FreeBSD is too Linux like and doesn't complain enough if you abuse libc functions. But maybe I should look if there are prebuild netbsd qemu images for weird platforms.

> It's a bit fluent now, with the all of the bi-endian architectures, but the modern OpenRISC is big-endian (and one of the main reasons I care, tbh); the NXPs Coldfire, a M68K relative; Atmel has one, I think, and the old dinosaurs like IBM's z/arch, of course. Well, still quite a number.

Hmm, only old stuff mostly? I don't think many news are developed since little endian also has the advantage of easier casts to larger values. And little endian is dominant, so why bother I guess. At least the pretty new RISCV is little.

**Comment by czurnieden** [comment]
> could you somehow share these qemu images you produced for the different architectures?

Made one with Debian 9 on Sparc64 (5G raw, mostly empty hence only 0.5G compressed). Dropped in a box aptly called Dropbox, so drop me note such that I can drop you a note how to get into the Box where I dropped it. Please don't forget to inlude you pubkey.

It didn't want to start with Grub, so no bootloader, only direct (but what do you need a bootloader for in a VM?).

> Hmm, only old stuff mostly?

Well it was bleeding edge when I was a young man ;-)

But back to business: are you ok with it now? Although I checked the box to put it into `develop` I still think it might be a better idea to keep the floating point separate from the main branch. At least for now. On the other side: quite a pain in the behind to keep them in sync.

There is also commit [b8bae34](https://github.com/libtom/libtommath/pull/160/commits/b8bae346a2448c34abe3c06bab8d1b5fcba2db18) which is a correction of the licence in the preamble, just housekeeping, and belongs into `develop`. So even if you don't want the rest: please cherry-pick that commmit to keep the preambles in unison.

Sorry for messing it up.

**Comment by minad** [comment]
Cool thank you! Looking forward to try the image! I will send you a mail. I am a bit sleepy right now, but scrolling through did not reveal some obvious issues :+1: I guess we can finally merge this!

**Comment by czurnieden** [comment]
Build with OpenBSD 6.4 on Sparc64 (qemu)
Static library: build & test succeeded (with gmake)
Warnings:
```
   * cc bn_mp_set_float.o
bn_mp_set_float.c: In function 'mp_set_float':
bn_mp_set_float.c:85: warning: 'res' may be used uninitialized in this function
   * cc bn_mp_set_int.o
   * cc bn_mp_set_long.o
   * cc bn_mp_set_long_double.o
bn_mp_set_long_double.c: In function 'mp_set_long_double':
bn_mp_set_long_double.c:180: warning: 'res' may be used uninitialized in this function
   * cc bn_mp_set_long_long.o
```
False positives.
Initialize `res` with `MP_OKAY` to get rid of warning?

Dynamic library: Build failed at linking time with
```
bn_error.lo: file not recognized: File format not recognized
collect2: ld returned 1 exit status
```
Hu?

**Comment by minad** [comment]
Why do you see the false positive only on openbsd? Do you use a particularly old compiler there? If that's the case I would not fix it.

**Comment by czurnieden** [comment]
>  In that case I think you should initialize the variable beforehand

Done.

**Comment by minad** [comment]
@sjaeckel I suggest to merge this PR, but I am unable to do it. Do you want to review it too?

**Comment by czurnieden** [comment]
[eef51ff ](https://github.com/libtom/libtommath/pull/160/commits/eef51ffca2c534d52d6b6d668b8117a103453c21) successfully tested on a <del>Falcon Heavy</del>Atari Falcon (virt. with Aranym which emulates a M68040 with MMU and FPU).

**Comment by sjaeckel** [comment]
> Do you want to review it too?

yes

... and when I look at these changes I understand why Tom didn't want to have float support ...

**Comment by buggywhip** [comment]
> ... and when I look at these changes I understand why Tom didn't want to have float support ...
> 
I've been swamped and then some problems, sorry, nevertheless, from a draft I didn't send...  

Rather than supporting floats/doubles in LTM per se, why not create a new "utilities" library for these and any other handy functions that come to mind?

**Comment by czurnieden** [comment]
Build&test successfull on MIPS (32-bit big-endian, emulated with Qemu)

```
user@debianmips32:~/GITHUB/libtommath$ cat /proc/cpuinfo 
system type		: MIPS Malta
machine			: mti,malta
processor		: 0
cpu model		: MIPS 24Kc V0.0  FPU V0.0
BogoMIPS		: 688.12
wait instruction	: yes
microsecond timers	: yes
tlb_entries		: 16
extra interrupt vector	: yes
hardware watchpoint	: yes, count: 1, address/irw mask: [0x0ff8]
isa			: mips1 mips2 mips32r1 mips32r2
ASEs implemented	: mips16
shadow register sets	: 1
kscratch registers	: 0
package			: 0
core			: 0
VCED exceptions		: not available
VCEI exceptions		: not available

user@debianmips32:~/GITHUB/libtommath$  uname -a
Linux debianmips32 4.9.0-8-4kc-malta #1 Debian 4.9.144-3.1 (2019-02-19) mips GNU/Linux
user@debianmips32:~/GITHUB/libtommath$ gcc -dM -E -x c /dev/null | grep BYTE_ORDER
#define __BYTE_ORDER__ __ORDER_BIG_ENDIAN__
user@debianmips32:~/GITHUB/libtommath$ gcc -dM -E -x c /dev/null | grep LDBL
#define __LDBL_MAX__ 1.7976931348623157e+308L
#define __LDBL_MAX_EXP__ 1024
#define __LDBL_HAS_INFINITY__ 1
#define __LDBL_MIN__ 2.2250738585072014e-308L
#define __LDBL_HAS_QUIET_NAN__ 1
#define __LDBL_HAS_DENORM__ 1
#define __LDBL_EPSILON__ 2.2204460492503131e-16L
#define __LDBL_MANT_DIG__ 53
#define __LDBL_MIN_EXP__ (-1021)
#define __LDBL_MAX_10_EXP__ 308
#define __LDBL_DENORM_MIN__ 4.9406564584124654e-324L
#define __LDBL_MIN_10_EXP__ (-307)
#define __LDBL_DIG__ 15
```

Next: Debian-s390x if I get it to run.

**Comment by czurnieden** [comment]
Build&test successfull on S390x (emulated with Qemu)
```
user@debians390x:~/GITHUB/libtommath$ cat /proc/cpuinfo
vendor_id       : IBM/S390
# processors    : 1
bogomips per cpu: 13370.00
max thread id   : 0
features	: esan3 zarch stfle msa ldisp eimm etf3eh highgprs 
processor 0: version = 00,  identification = 000000,  machine = 2827
user@debians390x:~/GITHUB/libtommath$ uname -a
Linux debians390x 4.9.0-8-s390x #1 SMP Debian 4.9.144-3.1 (2019-02-19) s390x GNU/Linux
user@debians390x:~/GITHUB/libtommath$ gcc -dM -E -x c /dev/null | grep BYTE_ORDER
#define __BYTE_ORDER__ __ORDER_BIG_ENDIAN__
user@debians390x:~/GITHUB/libtommath$ gcc -dM -E -x c /dev/null | grep LDBL
#define __LDBL_MAX__ 1.18973149535723176508575932662800702e+4932L
#define __LDBL_MAX_EXP__ 16384
#define __LDBL_HAS_INFINITY__ 1
#define __LDBL_MIN__ 3.36210314311209350626267781732175260e-4932L
#define __LDBL_HAS_QUIET_NAN__ 1
#define __LDBL_HAS_DENORM__ 1
#define __LDBL_EPSILON__ 1.92592994438723585305597794258492732e-34L
#define __LDBL_MANT_DIG__ 113
#define __LDBL_MIN_EXP__ (-16381)
#define __LDBL_MAX_10_EXP__ 4932
#define __LDBL_DENORM_MIN__ 6.47517511943802511092443895822764655e-4966L
#define __LDBL_MIN_10_EXP__ (-4931)
#define __LDBL_DIG__ 33
```

**Comment by czurnieden** [comment]
Re: build failure of shared library with OpenBSD

OpenBSD's buildin `libtool` was the culprit, we need the GNU libtool. OpenBSD offers both `gmake` and GNU `libtool` as packets installable with `pkg_add`. No easy solution found that solves it with a line of code or two, so I added a line or two of text to the documentation. See commit [
346e86e ](https://github.com/libtom/libtommath/pull/165/commits/346e86edfc7a7eb0b4a7516f7b6dc15bff6103c2) in  #165

**Comment by czurnieden** [comment]
Build&test successfull on amd64 (emulated with Qemu) in NetBSD 8.0. Documentation updated to reflect the quirks in building in NetBSD.

**Comment by sjaeckel** [comment]
IIUC this supersedes #156 right?

**Comment by czurnieden** [comment]
> IIUC this supersedes #156 right?

Mmh&hellip;the m68k problem? Yes, solved with [eef51ff](https://github.com/libtom/libtommath/pull/160/commits/eef51ffca2c534d52d6b6d668b8117a103453c21).

Also:
Successfully build&tested the static and shared libraries with Intel's compiler "icc (ICC) 19.0.2.187 20190117" just with a `CC=/path/to/icc make`. The comment in `makefile.unix` is wrong now. For that version of ICC on amd64. Updated documentation is on the way. I have currently no means to test it on Windows and with Intel processors.

BTW: all this compiler/arch/OS checks are valid for `develop`, too and should be a separate thread. But it is not really an "Issue", just Information and putting it in `docs/` is too far away, you shouldn't need to download it or search the repository to get to that information. With a list in `README.md` with a table of supported architectures/OSs/compilers? But that would be quite static: it needs a pull with all of the administrative overhead every time to get it changed.

Does github.com offer anything in that line? I really don't want to clutter this thread anymore with that stuff. It already bit me back and took me a sec' to get what @sjaeckel meant with "this".

**Comment by sjaeckel** [comment]
> Does github.com offer anything in that line?

There's https://github.com/libtom/libtommath/wiki :-) 

For now I only used the wiki in ltc for [1] and I also wanted to create something similar for ltm... but then forgot about it again...

[1] https://github.com/libtom/libtomcrypt/wiki/Maintainers-of-distribution-packages

**Comment by czurnieden** [comment]
> There's https://github.com/libtom/libtommath/wiki :-)

Yeah, why do I even ask \*sigh\* ;-)

OK, started it: home page with the first paragraph from https://www.libtom.net/LibTomMath/
And now to the tedious task of transfering manually fiddled LaTeX to Github's markup.

**Comment by czurnieden** [comment]
Started https://github.com/libtom/libtommath/wiki/Successes-and-Failures-of-Building-LibTomMath
I have neither Windows (all versions, all architectures) nor a modern MacOS, so please add, thank you.

A small caveat: Github's markup does not support code-blocks inside (nested?) lists. Either find a way to use Markup to do so or use HTML as I did. You also cannot use Markup inside HTML lists as it seems, you need to do e.g.: HTML links.

**Comment by czurnieden** [comment]
Learned that Travis has quite a long timeout for a single job: [50 minutes](https://docs.travis-ci.com/user/customizing-the-build/#build-timeouts) so I added the ability to run Valgrind to `testme.sh` and one test to `.travis.yml`. Seems to work.
I think @sjaeckel wanted to have it in general and for all tests?

File: bn_mp_set_long_double.c [hunk 15]
```diff
@@ -0,0 +1,276 @@
...
+         uint16_t sign:1;
+         /* fill up (i368: to 12 bytes, x86_64: to 16 bytes */
+         uint16_t padding;
+#endif
+      } ldbl_guts;
+   } cast;
+
+   cast.ldbl = b;
+
+   exp = (int)cast.ldbl_guts.exponent;
```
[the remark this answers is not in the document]
czurnieden: > Can we avoid using bitfields here?

Yes, we can (I think). But I found bit-fields simpler.

> bitfields are not well specified

They are quite well defined in the standard (vid. e.g.: 6.7.2.1 in INCITS/ISO/IEC 9899-2011[2012] for a start), but this one uses "or some other implementation-defined type[s]" so it is not "standard" in the most narrow sense, admitted but `long double` is not "fully standard" either.

That's why I have it restricted to the two compilers GCC and clang. The MingW port of GCC already needed some arm-twisting to make it complicit. MSVC can also be persuaded, at least that's what they told me, but I don't have one to test it and it is also moot because MSVC does not support `long double` as far as I remember.

And it is for `long double`, a type in widespread use but restricted to the x87 architecture. The number of different hardware implementations is: two. I think this code handles both of them sufficiently.

Yes, the world would be a happier place if `long double` goes extinct and gets replaced with a proper binary128. Or binary96 if you insist. But it won't happen, at least not in the near future. Maybe together with 128-bit registers?

minad:
@czurnieden I am going to close your outdated PRs for now, since they would need a lot of rebasing and discussion somehow died off without a merge a while ago. I am also at least partially responsible for that due to my refactorings etc. For now it would probably be best if we concentrate on fixing a few correctness issues, clean up the remaining API issues (#465 etc). But since you still seem to have a lot in your pipeline, it would be nice to have some kind of overview of what you have and what would be nice to have upstreamed here. But if possible, the PRs should be small and easy to review such that the time frame for a merge is shorter and the PRs don't start to bit rot again.

PR #165, opened 2019-02-27
offer alternative to arch

czurnieden:
Re: https://groups.google.com/forum/#!msg/libtom/jzP63O_JeHc/WcwlnYkWAgAJ

We had that problem once elsewhere but I cannot remember where (I think with the Debian sbuild/buildd environment?). 
`/usr/bin/arch` is not a link/alias to `uname -m` on my machine, so I took both.

### DIFF
--- a/makefile_include.mk
+++ b/makefile_include.mk
@@ -92,7 +92,7 @@ endif
 ifeq ($(PLATFORM),FreeBSD)
   _ARCH := $(shell sysctl -b hw.machine_arch)
 else
-  _ARCH := $(shell sh -c "if command -v arch 1> /dev/null;then arch;else uname -m;fi")
+  _ARCH := $(shell uname -m)
 endif
 
 # adjust coverage set
--- a/makefile_include.mk
+++ b/makefile_include.mk
@@ -92,7 +92,7 @@ endif
 ifeq ($(PLATFORM),FreeBSD)
   _ARCH := $(shell sysctl -b hw.machine_arch)
 else
-  _ARCH := $(shell arch)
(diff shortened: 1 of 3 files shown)


makefile_include.mk
```diff
@@ -92,7 +92,7 @@ endif
 ifeq ($(PLATFORM),FreeBSD)
   _ARCH := $(shell sysctl -b hw.machine_arch)
 else
-  _ARCH := $(shell arch)
+  _ARCH := $(shell sh -c "if command -v arch 1> /dev/null;then arch;else uname -m;fi")
```
sjaeckel: why not simply always `uname -m`?

czurnieden: I checked if it is a link/alias which it isn't. If I would have taken the venturous enterprise of RTFM I would have seen "arch - print machine hardware name (same as uname -m)" right at the beginning. But who reads the fine manual?

Yepp, I shouldn't shoot from the hip so often, it's bad for the resume ;-)

And it gets worse. While trying to find out if there is a Posix way to get the machine architecture 
I stumbled upon this little gem: https://superuser.com/questions/835514/why-do-uname-p-and-uname-m-and-arch-output-different-architectures
Apple. \*sigh\*
(Doesn't matter now but might matter later)

So it's `uname -m` then.

sjaeckel:
@buggywhip can you please check the implications of this patch on your mac as @czurnieden mentioned those differences in behavior on a mac?

sjaeckel:
@buggywhip feel free to open a new issue if there's a problem on Mac

PR #167, opened 2019-02-28
split demo.c test suite

minad:
Extract function tests to separate functions in test.h


minad:
This is the first mini step towards a better test suite. Followups could be:

* extract mtest-vs-test from demo.c to separate mtest_vs_test.c
* add more function tests to test.h
* simplify or get rid of shell test harness

czurnieden:
You made it worse! ;-)

But I see where you are going to.

Some <del>random rants</del>suggestions from a grumpy old fart:

* no program-logic in header files, please. Use individual C-files and glue them together with one header file.
* if we spend all the money for refactoring we should not save in the wrong place and get rid of global variables, too. And give them some meaningful names, when we are already at it. Yes, I know ;-)
* the preprocessor logic for the `LTM_DEMO_TEST_VS_MTEST ` branching is indeed not, how shall I put it very elegant. So, yes, put them in separate files as you said, that's a good idea.
* The individual functions have a good design: they take no arguments and return values for PASS/FAIL only but they should also be side-effect free to make it round.
* Please use parentheses and curly brackets, don't drop them just because you can.
* some tests are gone. Happened for a reason?
* misc.

minad:
Since this is a refactoring some minor things get worse and might not be in their final state. In particular the code duplication. The question is how to proceed here. I would prefer if things could be merged early and improved gradually (instead of a fully finished and polished PR).

>     * no program-logic in header files, please. Use individual C-files and glue them together with one header file.

I agree. But for the intermediate steps during the refactoring it is ok like this.

>     * if we spend all the money for refactoring we should not save in the wrong place and get rid of global variables, too. And give them some meaningful names, when we are already at it. Yes, I know ;-)

You probably mean the arrays used by kronecker etc? I will move them inside the corresponding functions.

>     * the preprocessor logic for the `LTM_DEMO_TEST_VS_MTEST ` branching is indeed not, how shall I put it very elegant. So, yes, put them in separate files as you said, that's a good idea.

Ok, we agree. This has to go away at some point.

>     * The individual functions have a good design: they take no arguments and return values for PASS/FAIL only but they should also be side-effect free to make it round.

Yes. I think they are side effect free. But I didn't read through them.

>     * Please use parentheses and curly brackets, don't drop them just because you can.

You mean the prologue and epilogue? I am usually not using braces for single line conditionals, hard to change habits...

>     * some tests are gone. Happened for a reason?

Now, this is serious! Which?

czurnieden:
> I agree. But for the intermediate steps during the refactoring it is ok like this.

"Release early and often!", he said ;-)

> I am usually not using braces for single line conditionals,

Not only that but the preprocessor stuff needs a bit of parenthesing, too. It needs to be as unambiguous as possible. LTM code gets used by more compilers than just GCC. Some of them are more standard compliant, most are less but all respect parentheses.

And it is the coding style here, older than us. I don't discuss it I just obey ;-)

> hard to change habits...

Yeah, I know.

> Now, this is serious! Which?

I was in the wrong branch at first. Checked all now, found none missing.

minad:
@czurnieden Thanks for checking!

sjaeckel:
thanks a lot! that starts to look good

> > * no program-logic in header files, please. Use individual C-files and glue them together with one header file.
> 
> I agree. But for the intermediate steps during the refactoring it is ok like this.

TBH that's the only point that's really bothering me in this PR

IIUC you're targeting on a final solution, can you already explain where we're going to? :-) 

Because if it's really going in this direction:

>    * Use individual C-files and glue them together with one header file.

why would you want to wait?

minad:
> thanks a lot! that starts to look good

Thanks!

> TBH that's the only point that's really bothering me in this PR
> 
> IIUC you're targeting on a final solution, can you already explain where we're going to? :-)

@sjaeckel There is no final state here. The test suite will be fluent as people add stuff and the current code is being cleaned up. This can happen stepwise as I mentioned before. The first step was this restructuring and splitting of the tests to ensure isolation and improve extensibility.

This PR as of now is fully functional and just a refactoring of the existing code. In order to avoid divergence with branches it would be nice if you merge it soon (if there are no real showstoppers). If there are minor issues they can also be fixed in followup PRs. This is at least the development style that I would prefer...

minad:
@czurnieden What do you think?

czurnieden:
> There is no final state here.

But you need to architect a design. I don't care for all of the gory details&mdash;I don't want a description in ISO-Z (altern. 2nd ed.) just a rough sketch.

Do you want to implement (different levels of) verbosity besides "testing 'f': FAIL"? A full core dump is very rarely necessary, a simple "expected a but got b" is enough in almost all cases.

> The test suite will be fluent as people add stuff and the current code is being cleaned up

I would prefer clean code when I add stuff.

> This PR as of now is fully functional and just a refactoring of the existing code

But in its current state it's a bigger mess than the original. And that means something ;-)
You should have at least the already mentioned individual c-files with a (single) header.

> In order to avoid divergence with branches

Neither API changes nor new functions are planned in the near future (as far as I know!), so `demo/demo.c` will not get changed beside a bugfix here and there which you can merge manually with some C&P.  `demo/demo.c` is also independent of the rest, merging the changes from `develop` should be no problem.

> This is at least the development style that I would prefer...

Well, you are a member of a quite divergent team now, eat it! ;-)

minad:
> Do you want to implement (different levels of) verbosity besides "testing 'f': FAIL"? A full core dump is very rarely necessary, a simple "expected a but got b" is enough in almost all cases.

Yes, that would be great. Maybe using some assert macros or helper functions, which print expected values etc. It will be done hopefully at some later point. But I don't see what it has to do with *this* cleanup work to get things rolling.

> But in its current state it's a bigger mess than the original. And that means something ;-) You should have at least the already mentioned individual c-files with a (single) header.

I disagree. The individual c files will be done. But this requires changes to the build system, which I didn't do yet. And there is a reason for that - I want to keep things in a working state step by step. Having all these different makefiles doesn't simplify things. However we want to keep them since portability is a big goal.

The state of this PR is also far from the perfectly polished test suite we could have in the end. It is an intermediate stage. However this intermediate stage is already strictly better than the current demo.c. The tests are separated. There are stronger guarantees that the tests don't interfere which each other. The refactoring was also done in such a way that there are no changes to test suite itself such that we don't loose anything.

And please be friendly and fair. If you would truly apply your high quality measures to the current test suite you should complain very loudly. However it is better to just improve things and that's what I am doing here.

> Neither API changes nor new functions are planned in the near future (as far as I know!), so `demo/demo.c` will not get changed beside a bugfix here and there which you can merge manually with some C&P.  `demo/demo.c` is also independent of the rest, merging the changes from `develop` should be no problem.

Well, for example your PR #160 changes demo.c. I would like to avoid merging back and forth...

czurnieden:
Yes, [85ff43c ](https://github.com/libtom/libtommath/pull/167/commits/85ff43cee02e1ac771818326197c69c08d5f9b64) looks much better!

> But I don't see what it has to do with this cleanup work to get things rolling.

If I ask it now or later, not much of difference, but if I ask later I might have forgotten what I wanted to ask in the meantime.

> Having all these different makefiles doesn't simplify things

Then make your life simpler with a single `demo/makefile` and call that from all of the other ones with `make -C`. You can take a look at how it is done in LTM for `doc/` .

> And please be friendly and fair. 

Fair I am (That's what I aim for at least, please hit me when I'm not!) but I reached the age where it seems to be my society-imposed duty to be a bit grumpy.

> If you would truly apply your high quality measures to the current test suite you should complain very loudly. 

It is not my style to get loud and I did not complain very loudly because I *know* that somebody *will* come and say something in the line of "That's a nice idea, and very nice of you to do it!" if I complain loudly.

> Well, for example your PR #160  changes demo.c. 

It is still unknown if #160 will make it into the main branch. The founder of LibTomMath came over and left some hints, so it is even less likely.

If you do it right it should not be much of a problem to insert new tests into your test-suite and don't forget: it is not your duty to implement the tests for the new functions but the duty of the author(s) of the new function!

Now that I went through it I see that you have `ndraw()` twice but that's not what I'm in for. The buffer there is fixed and does not get checked for overflow. `16,000` bytes is a lot but it needs to be checked. Either manually or automatically (preferred). For now the biggest seems to be 4095 characters read in as a base-64 which gets squared, that is `4095*6 * 2 = 49140` bits that are 6143 bytes (assuming `CHAR_BIT == 8`) and it is not printed in base-2, so not even close.
But it still makes me a bit uneasy, we should keep an eye on it.
(Why not use `mp_fwrite()` you might ask? It's because it might not be compiled in, the I/O functions are optional in LTM)

minad:
> If I ask it now or later, not much of difference, but if I ask later I might have forgotten what I wanted to ask in the meantime.

Ok, thats fine! Always ask. It sounded a bit like you want to have things already in this PR.

> Then make your life simpler with a single `demo/makefile` and call that from all of the other ones with `make -C`. You can take a look at how it is done in LTM for `doc/` .

Yes, we should do that at some point. But for now we can live with the modified old makefiles.

> Fair I am (That's what I aim for at least, please hit me when I'm not!) but I reached the age where it seems to be my society-imposed duty to be a bit grumpy.

Your comment came around as a bit harsh, therefore my request. But its ok!

> It is not my style to get loud and I did not complain very loudly because I _know_ that somebody _will_ come and say something in the line of "That's a nice idea, and very nice of you to do it!" if I complain loudly.

Hehe, that's the trap :-/

> It is still unknown if #160 will make it into the main branch. The founder of LibTomMath came over and left some hints, so it is even less likely.

I don't interpret it like this. I think if we keep things restricted to `mp_get/set_((long)? double|float)` we are fine. While the functions are ugly, I consider these functions essential and they were requested multiple times. We should not put more floating point stuff in however. I assume @tomstdenis and @sjaeckel aggree.

> If you do it right it should not be much of a problem to insert new tests into your test-suite and don't forget: it is not your duty to implement the tests for the new functions but the duty of the author(s) of the new function!

Sure it is not. But when I added mp_get_double I also added my mess to demo.c. As a user of the lib I am also more confident the better things are in shape.

> Now that I went through it I see that you have `ndraw()` twice but that's not what I'm in for. 

Yes, this can be addressed when we refactor the functions using ndraw.

> The buffer there is fixed and does not get checked for overflow. `16,000` bytes is a lot but it needs to be checked. Either manually or automatically (preferred). For now the biggest seems to be 4095 characters read in as a base-64 which gets squared, that is `4095*6 * 2 = 49140` bits that are 6143 bytes (assuming `CHAR_BIT == 8`) and it is not printed in base-2, so not even close.

This unchecked buffer is not my fault. This is bad. Who wrote such a thing? This needs refactoring later. But again - this has not much to do with the work being done here.

@sjaeckel @czurnieden Are you ok with merging this PR?

sjaeckel:
1. can you please rebase on top of develop
2. 
```
$ CFLAGS="-DLTM_DEMO_REAL_RAND" make test_standalone 
   * cc demo/main.o
demo/main.c: In function ‘main’:
demo/main.c:9:9: error: ‘fd_urandom’ undeclared (first use in this function)
    if (!fd_urandom) {
         ^
demo/main.c:9:9: note: each undeclared identifier is reported only once for each function it appears in
makefile:22: recipe for target 'demo/main.o' failed
make: *** [demo/main.o] Error 1
```

minad:
@sjaeckel Fixed

czurnieden:
> This unchecked buffer is not my fault.

Doesn't change the fact.

> This is bad.

No, it isn't, at least for now. But it could get a problem later.

> Who wrote such a thing?

You might try `git blame demo.c` if you really want to know but I don't think it matters.

> But again - this has not much to do with the work being done here.

You also fix all of the bugs you stumble upon while refactoring or at least take notes (e.g.: by inserting a `TODO` comment). Don't ignore them.

> Yes, we should do that at some point.

Please give a more precise date, "at some point" is a bit too fuzzy for my taste.

> Your comment came around as a bit harsh

Yeah, sometimes I'm a bit too german, sorry.

sjaeckel:
just one comment for now

>     * `ndraw()` is still in both files. Would suggest a third c-file e.g.: `utils.c` gathering all functions that are not pure tests.

no `utils.c` please :)

I'd prefer to have it in then in `main.c` and the prototype in `shared.h`

czurnieden:
> no utils.c please :)

Got bitten once? ;-)

> I'd prefer to have it in then in `main.c` and the prototype in shared.h

No problem.

minad:
@czurnieden atexit(cleanup) and double T-entry. Let's keep duplicate ndraw for now. This will be removed when the test which uses it will be refactored.

minad:
@czurnieden ...and added the better ndraw you suggested

minad:
Why are the tests failing now: The job exceeded the maximum log length, and has been terminated.

sjaeckel:
> Why are the tests failing now? The job exceeded the maximum log length, and has been terminated.

looks like a travis error where the job gets killed because it was silent for too long IIRC

minad:
@sjaeckel I try to restart one of the green builds to see what happens. I didn't change anything here which could have made the tests fail. But let's see.

sjaeckel:
> I try to restart one of the green builds to see what happens.

sure, do that, but leave that other one running as well... I want to see if it's only test vs. mtest that is failing :-D

demo/opponent.c
```diff
@@ -1,15 +1,32 @@
...
+   char *buf;
+   int size, size16;
+
+   mp_radix_size(a, 10, &size);
+   mp_radix_size(a, 16, &size16);
+   if (size16 > size) {
+      size = size16;
+   }
+
+   buf = malloc((size_t) size);
```
minad: What is a bit bad about this ndraw - it always allocates and ndraw is called quite often in mtest-vs-test.
However in this case the malloc requests should be satisfied by the malloc thread local cache, which is usually there to work around stupid allocation schemes like this. Now @czurnieden will come and find some historical systems where that's not the case :D Maybe he also comes up with a better solution? Allocate the buffer only once and make it global again? Or use alloca :hankey:?

minad:
@sjaeckel The develop build worked. something is wrong here

demo/opponent.c
czurnieden: As far as I understand that old code, `ndraw` and `draw` are only called in case of an error. O would have suggested a rough approximation of `ceil(log_2(mp_int)/log_2(10))+angst_allowance` to check if   `buf[16000]` is big enough.

> czurnieden will come and find some historical systems where that's not the case

Of course I will! ;-)

Ah, I'm too late. OK, off to #169

minad: I have to look more carefully. You are right. It happens only in case of failure. No problem with the malloced ndraw in the new PR #169.

PR #168, opened 2019-03-03
Split test suite

minad:
Squashed the old PR #167


sjaeckel:
why not simply force-push to the other branch?

minad:
I did that multiple times now. The CI got too messy and I lost track in the list of CI builds in the travis interface. It is unclear which commit failed and why (I think it was the malloc/ndraw one). And before that builds were canceled. So please let this one go through..

minad:
I also don't know how force pushing works together with CI. If I force push a branch for a currently running travis build, I think the build will fail too. And then there is the question if force pushed builds can be restarted. Force pushing is just bad I think.

PR #169, opened 2019-03-04
Split test suite

minad:
@sjaeckel Ok, continuing #167 here. The build failure is fixed and was due to a missing printf during cleanup. See the corresponding commit. Besides that, the issues mentioned by you and @czurnieden should be fixed.


minad:
@sjaeckel Yes, this should be done next.

PR #170, opened 2019-03-05
Space for speed

czurnieden:
First sketch of a method to implement optional speed (big-O) optimization.

The example used is a O(1) replacement for the O(n^2) method currently used in  `mp_radix_size()`. The method presented here works well for the replacements of a function in a single file, not so well if there is more than one function/file involved, for example: the balancing of multiplication which would need a file for the function itself and a hook in `bn_mp_mul.c`. Others, like an O(M(n) log(n)) version of the `mp_toradix()` (currently O(n^2)) would depend on the fast version of  `mp_radix_size()` to function.


czurnieden:
After a lot of time-outs I decided to give the original version of `mp_n_root()` a proper start-value. It's usable now. And Travis, too, seems to be OK with it.

minad:
Could you split this PR? For example one PR per function which has been improved? Otherwise I don't really get what you are doing here.

czurnieden:
> Could you split this PR?

Of course I can, no problem, I just found it easier for the user to bundle all of the "space for speed" optimizations into one packet. I mean: the current method via macros (e.g.: `tommath_superclass.h`) is a pain in the behind, isn't it? ;-)

But, as I said above: no problem, all what I already have and what is planned is independent of each other. Most of them are not independent of LTM, but they need only small changes, like the multiplication balancing and all of them can be put into preprocessor brackets (like the one in `bn_mp_div.c` switching between "normal" and "small").

Before somebody thinks that I already started splitting: the nth-root algorithm in this PR is the optimized one, the one in #189 is only a bugfix of Tom's original.

minad:

Issue #164, opened 2019-02-22
Cleanup naming of internal functions?


minad:
~~~ c
int s_mp_add(const mp_int *a, const mp_int *b, mp_int *c);
int s_mp_sub(const mp_int *a, const mp_int *b, mp_int *c);
#define s_mp_mul(a, b, c) s_mp_mul_digs(a, b, c, (a)->used + (b)->used + 1)
int fast_s_mp_mul_digs(const mp_int *a, const mp_int *b, mp_int *c, int digs);
int s_mp_mul_digs(const mp_int *a, const mp_int *b, mp_int *c, int digs);
int fast_s_mp_mul_high_digs(const mp_int *a, const mp_int *b, mp_int *c, int digs);
int s_mp_mul_high_digs(const mp_int *a, const mp_int *b, mp_int *c, int digs);
int fast_s_mp_sqr(const mp_int *a, mp_int *b);
int s_mp_sqr(const mp_int *a, mp_int *b);
int mp_karatsuba_mul(const mp_int *a, const mp_int *b, mp_int *c);
int mp_toom_mul(const mp_int *a, const mp_int *b, mp_int *c);
int mp_karatsuba_sqr(const mp_int *a, mp_int *b);
int mp_toom_sqr(const mp_int *a, mp_int *b);
int fast_mp_invmod(const mp_int *a, const mp_int *b, mp_int *c);
int mp_invmod_slow(const mp_int *a, const mp_int *b, mp_int *c);
int fast_mp_montgomery_reduce(mp_int *x, const mp_int *n, mp_digit rho);
int mp_exptmod_fast(const mp_int *G, const mp_int *X, const mp_int *P, mp_int *Y, int redmode);
int s_mp_exptmod(const mp_int *G, const mp_int *X, const mp_int *P, mp_int *Y, int redmode);
void bn_reverse(unsigned char *s, int len);

extern const char *const mp_s_rmap;
extern const uint8_t mp_s_rmap_reverse[];
extern const size_t mp_s_rmap_reverse_sz;
~~~

`s_mp_*` for static (or private nonstatic) functions is not used consistently

czurnieden:
It doesn't look nice but it works for everybody and isn't very painful for us to work with, he says, fully aware of how that sounds like. ;-)

I wouldn't touch the API in LTM, you would inconvenience some users, that is: break their stuff. Such things need to get announced early enough and loud enough and at least two releases in advance.

Otherwise: yes, why not.

[...]

PR #172, opened 2019-03-07
Rename internals

minad:
See #164. These are only internal functions, no API or ABI changes. The renaming is done by the script. If something needs tweaking only the script should be changed and executed again.

### DIFF
--- a/tommath_private.h
+++ b/tommath_private.h
@@ -90,6 +90,8 @@ int func_name (mp_int * a, type b)                       \
    return res;                                           \
 }
 
+/* deprecated functions */
+
 #ifdef __cplusplus
 }
 #endif
--- a/tommath_private.h
+++ b/tommath_private.h
@@ -50,6 +50,7 @@ int fast_s_mp_mul_high_digs(const mp_int *a, const mp_int *b, mp_int *c, int dig
 int s_mp_mul_high_digs(const mp_int *a, const mp_int *b, mp_int *c, int digs);
 int fast_s_mp_sqr(const mp_int *a, mp_int *b);
 int s_mp_sqr(const mp_int *a, mp_int *b);
+int mp_balance_mul(const mp_int *a, const mp_int *b, mp_int *c);
 int mp_karatsuba_mul(const mp_int *a, const mp_int *b, mp_int *c);
 int mp_toom_mul(const mp_int *a, const mp_int *b, mp_int *c);
 int mp_karatsuba_sqr(const mp_int *a, mp_int *b);
--- a/tommath_private.h
+++ b/tommath_private.h
@@ -10,6 +10,14 @@
 extern "C" {
 #endif
 
+#if defined(__GNUC__) && (__GNUC__ * 100 + __GNUC_MINOR__ >= 301)
+#  define MP_DEPRECATED(x) __attribute__((deprecated("replaced by " #x)))
+#elif defined(_MSC_VER) && _MSC_VER >= 1500
+#  define MP_DEPRECATED(x) __declspec(deprecated("replaced by " #x))
+#else
+#  define MP_DEPRECATED
+#endif
+
 /* define heap macros */
 #ifndef MP_MALLOC
 /* default to libc stuff */
--- a/tommath_private.h
+++ b/tommath_private.h
@@ -6,37 +6,32 @@
 
 #include "tommath.h"
 
-#ifndef MIN
-#define MIN(x, y) (((x) < (y)) ? (x) : (y))
-#endif
-
-#ifndef MAX
-#define MAX(x, y) (((x) > (y)) ? (x) : (y))
-#endif
-
 #ifdef __cplusplus
 extern "C" {
 #endif
 
 /* define heap macros */
-#ifndef XMALLOC
+#ifndef MP_MALLOC
 /* default to libc stuff */
-#   define XMALLOC(size)                   malloc(size)
-#   define XREALLOC(mem, oldsize, newsize) realloc(mem, newsize)
-#   define XCALLOC(nmemb, size)            calloc(nmemb, size)
-#   define XFREE(mem, size)                free(mem)
+#   define MP_MALLOC(size)                   malloc(size)
+#   define MP_REALLOC(mem, oldsize, newsize) realloc(mem, newsize)
+#   define MP_CALLOC(nmemb, size)            calloc(nmemb, size)
+#   define MP_FREE(mem, size)                free(mem)
 #else
 /* prototypes for our heap functions */
-extern void *XMALLOC(size_t size);
-extern void *XREALLOC(void *mem, size_t oldsize, size_t newsize);
-extern void *XCALLOC(size_t nmemb, size_t size);
-extern void XFREE(void *mem, size_t size);
+extern void *MP_MALLOC(size_t size);
+extern void *MP_REALLOC(void *mem, size_t oldsize, size_t newsize);
+extern void *MP_CALLOC(size_t nmemb, size_t size);
+extern void MP_FREE(void *mem, size_t size);
 #endif
 
+#define MP_MIN(x, y) (((x) < (y)) ? (x) : (y))
+#define MP_MAX(x, y) (((x) > (y)) ? (x) : (y))
+
 /* ---> Basic Manipulations <--- */
-#define IS_ZERO(a) ((a)->used == 0)
-#define IS_EVEN(a) (((a)->used == 0) || (((a)->dp[0] & 1u) == 0u))
-#define IS_ODD(a)  (((a)->used > 0) && (((a)->dp[0] & 1u) == 1u))
+#define MP_IS_ZERO(a) ((a)->used == 0)
+#define MP_IS_EVEN(a) (((a)->used == 0) || (((a)->dp[0] & 1u) == 0u))
+#define MP_IS_ODD(a)  (((a)->used > 0) && (((a)->dp[0] & 1u) == 1u))
 
 /* lowlevel functions, do not call! */
 int s_mp_add(const mp_int *a, const mp_int *b, mp_int *c);
--- a/tommath_private.h
+++ b/tommath_private.h
@@ -44,23 +44,23 @@ extern void MP_FREE(void *mem, size_t size);
 /* lowlevel functions, do not call! */
 int s_mp_add(const mp_int *a, const mp_int *b, mp_int *c);
 int s_mp_sub(const mp_int *a, const mp_int *b, mp_int *c);
-int fast_s_mp_mul_digs(const mp_int *a, const mp_int *b, mp_int *c, int digs);
+int s_mp_mul_digs_fast(const mp_int *a, const mp_int *b, mp_int *c, int digs);
 int s_mp_mul_digs(const mp_int *a, const mp_int *b, mp_int *c, int digs);
-int fast_s_mp_mul_high_digs(const mp_int *a, const mp_int *b, mp_int *c, int digs);
+int s_mp_mul_high_digs_fast(const mp_int *a, const mp_int *b, mp_int *c, int digs);
 int s_mp_mul_high_digs(const mp_int *a, const mp_int *b, mp_int *c, int digs);
-int fast_s_mp_sqr(const mp_int *a, mp_int *b);
+int s_mp_sqr_fast(const mp_int *a, mp_int *b);
 int s_mp_sqr(const mp_int *a, mp_int *b);
-int mp_balance_mul(const mp_int *a, const mp_int *b, mp_int *c);
-int mp_karatsuba_mul(const mp_int *a, const mp_int *b, mp_int *c);
-int mp_toom_mul(const mp_int *a, const mp_int *b, mp_int *c);
-int mp_karatsuba_sqr(const mp_int *a, mp_int *b);
-int mp_toom_sqr(const mp_int *a, mp_int *b);
-int fast_mp_invmod(const mp_int *a, const mp_int *b, mp_int *c);
-int mp_invmod_slow(const mp_int *a, const mp_int *b, mp_int *c);
-int fast_mp_montgomery_reduce(mp_int *x, const mp_int *n, mp_digit rho);
-int mp_exptmod_fast(const mp_int *G, const mp_int *X, const mp_int *P, mp_int *Y, int redmode);
+int s_mp_balance_mul(const mp_int *a, const mp_int *b, mp_int *c);
+int s_mp_karatsuba_mul(const mp_int *a, const mp_int *b, mp_int *c);
+int s_mp_toom_mul(const mp_int *a, const mp_int *b, mp_int *c);
+int s_mp_karatsuba_sqr(const mp_int *a, mp_int *b);
+int s_mp_toom_sqr(const mp_int *a, mp_int *b);
+int s_mp_invmod_fast(const mp_int *a, const mp_int *b, mp_int *c);
+int s_mp_invmod_slow(const mp_int *a, const mp_int *b, mp_int *c);
+int s_mp_montgomery_reduce_fast(mp_int *x, const mp_int *n, mp_digit rho);
+int s_mp_exptmod_fast(const mp_int *G, const mp_int *X, const mp_int *P, mp_int *Y, int redmode);
 int s_mp_exptmod(const mp_int *G, const mp_int *X, const mp_int *P, mp_int *Y, int redmode);
-void bn_reverse(unsigned char *s, int len);
+void s_mp_reverse(unsigned char *s, int len);
 
 extern const char *const mp_s_rmap;
 extern const uint8_t mp_s_rmap_reverse[];
@@ -91,6 +91,21 @@ int func_name (mp_int * a, type b)                       \
 }
 
 /* deprecated functions */
+MP_DEPRECATED(s_mp_invmod_fast) int fast_mp_invmod(const mp_int *a, const mp_int *b, mp_int *c);
+MP_DEPRECATED(s_mp_montgomery_reduce_fast) int fast_mp_montgomery_reduce(mp_int *x, const mp_int *n, mp_digit rho);
+MP_DEPRECATED(s_mp_mul_digs_fast) int fast_s_mp_mul_digs(const mp_int *a, const mp_int *b, mp_int *c, int digs);
+MP_DEPRECATED(s_mp_mul_high_digs_fast) int fast_s_mp_mul_high_digs(const mp_int *a, const mp_int *b, mp_int *c,
+      int digs);
+MP_DEPRECATED(s_mp_sqr_fast) int fast_s_mp_sqr(const mp_int *a, mp_int *b);
+MP_DEPRECATED(s_mp_balance_mul) int mp_balance_mul(const mp_int *a, const mp_int *b, mp_int *c);
+MP_DEPRECATED(s_mp_exptmod_fast) int mp_exptmod_fast(const mp_int *G, const mp_int *X, const mp_int *P, mp_int *Y,
+      int redmode);
+MP_DEPRECATED(s_mp_invmod_slow) int mp_invmod_slow(const mp_int *a, const mp_int *b, mp_int *c);
+MP_DEPRECATED(s_mp_karatsuba_mul) int mp_karatsuba_mul(const mp_int *a, const mp_int *b, mp_int *c);
+MP_DEPRECATED(s_mp_karatsuba_sqr) int mp_karatsuba_sqr(const mp_int *a, mp_int *b);
+MP_DEPRECATED(s_mp_toom_mul) int mp_toom_mul(const mp_int *a, const mp_int *b, mp_int *c);
+MP_DEPRECATED(s_mp_toom_sqr) int mp_toom_sqr(const mp_int *a, mp_int *b);
+MP_DEPRECATED(s_mp_reverse) void bn_reverse(unsigned char *s, int len);
 
 #ifdef __cplusplus
 }
--- a/tommath_private.h
+++ b/tommath_private.h
@@ -36,7 +36,6 @@ extern void MP_FREE(void *mem, size_t size);
 /* lowlevel functions, do not call! */
 int s_mp_add(const mp_int *a, const mp_int *b, mp_int *c);
 int s_mp_sub(const mp_int *a, const mp_int *b, mp_int *c);
-#define s_mp_mul(a, b, c) s_mp_mul_digs(a, b, c, (a)->used + (b)->used + 1)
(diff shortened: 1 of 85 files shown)


minad:
@sjaeckel You mentioned that the non-hidden symbols in the private header are exposed and are thus available to ld etc. This is true. However they still don't belong to neither the API nor the (stable) ABI. If some library users deliberately access hidden symbols, they must assume that those symbols vanish at some point.

sjaeckel:
This very much collides with 'Such things need to get announced early enough and loud enough and at least two releases in advance.' [1] and therefore I'm not sure what I should do with it _now_.

Also I'm not sure if we shouldn't first make a plan where we want to go, especially in regards to [2], [3] resp. [4], [5] and potentially more...

[1] https://github.com/libtom/libtommath/issues/164#issuecomment-466713658
[2] https://github.com/libtom/libtommath/pull/160#issuecomment-467650547
[3] https://github.com/libtom/libtommath/issues/164#issuecomment-469456398
[4] https://github.com/libtom/libtommath/pull/155
[5] https://github.com/libtom/tomsfastmath/issues/10

minad:
Sorry for repeating myself: Why do you need to announce a change of internals? The possibility of change is implied by the fact that these are marked private.

See for example
http://www.ros.org/reps/rep-0009.html
which explicity allows removing/adding static privates. Well its c++ but it somehow transfers.

sjaeckel:
> which explicity allows removing/adding static privates

but the functions you renamed are not static

BTW regarding the 'two releases in advance': I count minor releases, so I don't see this only changing in 3.0.0

how about adding something like this for all renamed functions?

```
diff --git a/bn_s_mp_exptmod_fast.c b/bn_s_mp_exptmod_fast.c
index c1b6da2..299ed40 100644
--- a/bn_s_mp_exptmod_fast.c
+++ b/bn_s_mp_exptmod_fast.c
@@ -313,2 +313,7 @@ LBL_M:
 }
+
+int mp_exptmod_fast(const mp_int *G, const mp_int *X, const mp_int *P, mp_int *Y, int redmode)
+{
+   return s_mp_exptmod_fast(G, X, P, Y, redmode);
+}
 #endif
diff --git a/tommath_private.h b/tommath_private.h
index f651984..c9a30dd 100644
--- a/tommath_private.h
+++ b/tommath_private.h
@@ -17,2 +17,18 @@
 
+#if defined(__GNUC__) && (__GNUC__ * 100 + __GNUC_MINOR__ >= 301)
+   #define LTM_DEPRECATED __attribute__((deprecated))
+#elif defined(_MSC_VER) && _MSC_VER >= 1500
+   /* supported since Visual Studio 2008 */
+   #define LTM_DEPRECATED __declspec(deprecated)
+#else
+   #define LTM_DEPRECATED
+#endif
+
+/* ToDo: This needs some more work */
+#if __GNUC__ >= 4
+   #define LTM_PRIVATE  __attribute__ ((visibility ("hidden")))
+#else
+   #define LTM_PRIVATE
+#endif
+
 #ifndef MIN
@@ -75,3 +91,3 @@ int s_mp_invmod_slow(const mp_int *a, const mp_int *b, mp_int *c);
 int s_mp_montgomery_reduce_fast(mp_int *x, const mp_int *n, mp_digit rho);
-int s_mp_exptmod_fast(const mp_int *G, const mp_int *X, const mp_int *P, mp_int *Y, int redmode);
+LTM_PRIVATE int s_mp_exptmod_fast(const mp_int *G, const mp_int *X, const mp_int *P, mp_int *Y, int redmode);
 int s_mp_exptmod(const mp_int *G, const mp_int *X, const mp_int *P, mp_int *Y, int redmode);
@@ -79,2 +95,8 @@ void s_mp_reverse(unsigned char *s, int len);
 
+/* Soon to be deprecated functions
+ * These functions will be removed in v2.0.0 and their public visible counterparts are private.
+ * Please make sure to use the appropriate API functions instead of these ones.
+ */
+LTM_DEPRECATED int mp_exptmod_fast(const mp_int *G, const mp_int *X, const mp_int *P, mp_int *Y, int redmode);
+
 extern const char *const mp_s_rmap;
```

like that we don't break existing stuff but tell the users that it will break in the future

sure, if they stick with <1.2.0 until 2.0.0 is out they won't notice and stuff will break, but that's not something we can solve

This definitely needs some more work, but we shouldn't just remove stuff that was there forever.

minad:
I recommend the following plan:

* [1, 3]: #172 can be merged as is since there is no API change
* [2]: #160 floating point enhancements should be merged since it arguably improves portability. There are no downsides except complexity. Also no API problems.
* [4]: #154 should not be merged yet since it is unclear what the reason for the linking error is. Is there some explanation somewhere? I find the patch a bit unsatisfying. We could explore a possibility to avoid introducing these accessor functions, as I suggested before.
* [5]: What does fastmath have to do with this library? Are they going to be merged at some point? If this is the case, this is kind of a longshot and should not impede making changes to the libraries in the meantime.

minad:
@sjaeckel Oh, I just saw you commented. You are wrong about "but the functions you renamed are not static". In C++ private static functions are still linker visible as in our case. Our case with private functions in tommath_private.h transfers exactly.

minad:
Concerning your proposal: 

1. We should probably add LTM_PRIVATE macros, but this is separate work from this PR.
2. I would not add these deprecated wrappers, since they are unneeded and we are not doing an API/ABI change here. This PR is about reducing cruft, not introducing more of it.
3. If you want to go through with this using versioning, we could remove the deprecate wrappers in 2.2?

But I don't see a reason for going through the versioning here. And I have yet to see a good argument. Please let you convince otherwise, even if you and @czurnieden's reaction was initially critical to the change. I argue that you are unnecessarily careful. libtommath already has a good separation between private and public and we can use that.

sjaeckel:
Well TBH I don't care about C++ semantics and how whatnot translates in theory to this project.

If we remove these symbols it will mean 

1. an ABI bump 
2. breakage if someone uses the symbols

Regarding the ABI bump:
Yeah sure, in itself it's no big deal, but 
1. I don't want to end up with ABI version 42 just because... [1] fits here pretty well
2. I don't know about most of the distro guidelines, but if we merge this now as is I don't see an effortless update of ltm in e.g. debian until bullseye. Sure someone could start backporting but we were just in a state where most distros don't need any bigger patches to package this library ... (@dod38fr please correct me if I'm wrong)

[1] https://semver.org/#if-even-the-tiniest-backwards-incompatible-changes-to-the-public-api-require-a-major-version-bump-wont-i-end-up-at-version-4200-very-rapidly

>     3\. If you want to go through with this using versioning, we could remove the deprecate wrappers in 2.2?

in 2.0.0

> I recommend the following plan:
> 
>     * [1, 3]: #172 can be merged as is since there is no API change

see above

>     * [2]: #160 floating point enhancements should be merged since it arguably improves portability. There are no downsides except complexity. Also no API problems.

ACK

>     * [4]: #154 should not be merged yet since it is unclear what the reason for the linking error is. Is there some explanation somewhere? I find the patch a bit unsatisfying. We could explore a possibility to avoid introducing these accessor functions, as I suggested before.

Yes, this needs some work

>     * [5]: What does fastmath have to do with this library? Are they going to be merged at some point? If this is the case, this is kind of a longshot and should not impede making changes to the libraries in the meantime.

Yes, the original idea was to merge ltm and tfm. To prevent further drifting apart it should've been done yesterday, but yeah...

minad:
Ok, I stop arguing here since I disagree with your definition of the public ABI. Note that this PR doesn't have any urgency and can be easily merged at some point in time when you see fit in your versioning scheme by just executing the rename.sh script again.

It seems on the other points of the plan we agree.

> Yes, the original idea was to merge ltm and tfm. To prevent further drifting apart it should've been done yesterday, but yeah...

When I started using ltm, I also checked fastmath but it seemed to have a wildly different API and a different purpose using fixed size integers. I guess I have to check again. Given that there are a few libraries providing arbitrarily sized integers (ltm, gmp, etc), I think fastmath has its own niche and the two libraries should **not** be merged.

sjaeckel:
> Ok, I stop arguing here since I disagree with your definition of the public ABI.

there's nothing to argue about, it's not me who defined this http://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html

czurnieden:
> [4]: #154 should not be merged yet since it is unclear what the reason for the linking error is.

The reason for the linking error lies in the intricacies of the x64-32 ABI. Basically `long` and the pointers are only 32 bit large. That causes trouble with exporting (internal) global variables/functions with shared libraries and PIC relocation. I knew the error from (bad) assembler before but not from C hence my initial befuddlement. Short description here:
https://stackoverflow.com/questions/48369771/relocation-r-x86-64-pc32-against-symbol-when-calling-function-from-inline-assemb

> Is there some explanation somewhere?

Not really. At least no single point of information. I can share some links, but I found none for that exact problem. I could ask at stackoverflow? With a bit of luck we can get somebody who is good at explaining?

I think I already had the following link, hadn't I?
https://bugs.chromium.org/p/webm/issues/detail?id=46

A general link:
http://www.akkadia.org/drepper/dsohowto.pdf

Some thoughts about the future of x64_32
https://lwn.net/Articles/774734/

> I find the patch a bit unsatisfying.

The patch itself is correct although a bit too specific for my taste--needs GCC>4 (which is really old, admitted) or Intel's ICC, and ELF, and an x86-64 (ARM has something similar to the x86 specific x64_32 with their arm64 but with a different ABI, not yet tested).

Another two links showing how others do it:

https://docs.microsoft.com/en-us/cpp/build/exporting-from-a-dll-using-declspec-dllexport?view=vs-2017
https://www.ibm.com/developerworks/aix/library/au-aix-symbol-visibility/index.html 

> We could explore a possibility to avoid introducing these accessor functions, as I suggested before.

It's just for x64_32 wouldn't invest much work into it. You can test for the x64_32 ABI and exclude the whole mess for the rest with the preprocessor. I'm sure it can be done more elegantly as I did it, though.

minad:
> The reason for the linking error lies in the intricacies of the x64-32 ABI. Basically `long` and the pointers are only 32 bit large. That causes trouble with exporting (internal) global variables/functions with shared libraries and PIC relocation. I knew the error from (bad) assembler before but not from C hence my initial befuddlement. Short description here:
> https://stackoverflow.com/questions/48369771/relocation-r-x86-64-pc32-against-symbol-when-calling-function-from-inline-assemb

Is the x64-32 abi broken in this sense? I have no idea what the technical reason is for this. In theory x64-32 should not be much different from x32 + additional 64bit registers available. 

I will check out your links... Thank you for collecting them.

> Not really. At least no single point of information. I can share some links, but I found none for that exact problem. I could ask at stackoverflow? With a bit of luck we can get somebody who is good at explaining?

Maybe it could help. Would you like to ask? I won't do it since I don't have an account there and there are just to many janitors around over there...

> Some thoughts about the future of x64_32
> https://lwn.net/Articles/774734/

I've read that one before. And I hope that it gets dropped since it is a bad idea. For this reason I would also not put much effort in a fix.

> It's just for x64_32 wouldn't invest much work into it. You can test for the x64_32 ABI and exclude the whole mess for the rest with the preprocessor. I'm sure it can be done more elegantly as I did it, though.

I am pretty sure it can be done more elegantly. But as you said, we shouldn't put effort in it. Just drop the patch and wontfix?

minad:
What I don't get or assuming wrongly - in this x64-32 abi, is the virtual address space not artificially restricted such that only 32 bits are used? Are shared objects still placed arbitrarily within the 48 bit address space for ASLR etc? Then these relocation errors can occur obviously when the address does not fit within the 32 bit R_X86_64_PC32 relocation.

minad:
See also this link 
http://eli.thegreenplace.net/2011/11/11/position-independent-code-pic-in-shared-libraries-on-x64/

minad:
@sjaeckel rebased and added deprecated functions

tommath_private.h
```diff
@@ -6,58 +6,61 @@
...
+#  define MP_DEPRECATED(x) __attribute__((deprecated("replaced by " #x)))
+#elif defined(_MSC_VER) && _MSC_VER >= 1500
+#  define MP_DEPRECATED(x) __declspec(deprecated("replaced by " #x))
+#else
+#  define MP_DEPRECATED
+#endif
+
 /* define heap macros */
-#ifndef XMALLOC
+#ifndef MP_MALLOC
```
sjaeckel: I like the idea, but I guess we have to keep these X-versions around until 2.0.0

demo/timing.c
```diff
@@ -10,10 +10,10 @@
 #define SLEEP
 #endif
 
-#ifdef LTM_TIMING_REAL_RAND
-#define LTM_TIMING_RAND_SEED  time(NULL)
+#ifdef LTM_TIMP_MING_REAL_RAND
```
sjaeckel: I've the impression your `MP_MIN` regex went wild ;)

tommath_private.h
```diff
@@ -6,58 +6,61 @@
...
-extern void *XREALLOC(void *mem, size_t oldsize, size_t newsize);
-extern void *XCALLOC(size_t nmemb, size_t size);
-extern void XFREE(void *mem, size_t size);
+extern void *MP_MALLOC(size_t size);
+extern void *MP_REALLOC(void *mem, size_t oldsize, size_t newsize);
+extern void *MP_CALLOC(size_t nmemb, size_t size);
+extern void MP_FREE(void *mem, size_t size);
 #endif
 
+#define MP_MIN(x, y) (((x) < (y)) ? (x) : (y))
```
sjaeckel: The same as for the Xalloc macros applies here

changes.txt
```diff
@@ -24,7 +24,7 @@ Aug 29th, 2017
 v1.0.1
        -- Dmitry Kovalenko provided fixes to mp_add_d() and mp_init_copy()
        -- Matt Johnston contributed some improvements to mp_div_2d(),
-          mp_exptmod_fast(), mp_mod() and mp_mulmod()
+          s_mp_exptmod_fast(), mp_mod() and mp_mulmod()
```
sjaeckel: I'd say we should keep this file as is and only add to it on a new release

demo/timing.c
minad: haha, thanks for checking :)

tommath_private.h
minad: No. You just added CALLOC and removed it etc etc. This is internal API not even ABI. And these are only macros.

tommath_private.h
minad: What applies here? This is not API.

changes.txt
minad: Ok!

tommath_private.h
sjaeckel: Uhm, you're right... we already broke this part in #192 

The removal of `XCALLOC` doesn't break, since if someone overrides those macros with their own version it's simply not called anymore, whereas with #192 the function signatures changed and now even the name.

IMO we have to revert #192 (and part of #209) as well.

minad:
Fixed the MAX/MIN renaming issues.

tommath_private.h
minad: This is not an issue. Please! This is a private header!

minad: I added the deprecated functions to keep the *private* ABI stable (which is already unnecessary in my opinion). Changes to tommath_private.h are a even more minor. Let's change this to move forward :)

sjaeckel: > Let's change this to move forward

what's so important about this change?

sjaeckel: why can't you just keep it for a future PR and move forward instead of discussing such stuff and then complaining that those discussions eat up your time?

btw. I will leave #192 in develop, but will revert it for 1.2.0

minad: We can also keep all the things as is :( However we could also try to improve things as for example in #214. This particular PR is about having a well-defined API/ABI. I consider that important.

sjaeckel: > We can also keep all the things as is :(

oh come on... we simply don't break existing stuff w/o warning first...

I'm fine with #214 if @karel-m is fine with it

> This particular PR is about having a well-defined API/ABI. I consider that important.

I consider that important as well.
But _renaming_ something that was there for ages, only for the sake of consistency, without any other gain, I don't consider as a such important improvement.

minad: > why can't you just keep it for a future PR and move forward instead of discussing such stuff and then complaining that those discussions eat up your time?

Sure I will. But then in the future I have to come back again to the PR and update it.

> btw. I will leave #192 in develop, but will revert it for 1.2.0

Ok.

sjaeckel: > Sure I will. But then in the future I have to come back again to the PR and update it.

if you keep it in this nice script somebody else could also do it ;-)

minad: You know, for that reason I made it as script. Otherwise this would be totally crazy.

minad: > oh come on... we simply don't break existing stuff w/o warning first...

This is not right what you are saying. I am very careful not to break stuff. I adhere to a well defined boundary of what can be changed and what cannot.

> I'm fine with #214 if @karel-m is fine with it

Ok.

> I consider that important as well. But _renaming_ something that was there for ages, only for the sake of consistency, without any other gain, I don't consider as a such important improvement.

I agree. But for example just now, mp_balance_mul was added which is actually a private function and should be called s_mp_balance_mul. By mistake, the API was extended. Such things are to be avoided.

sjaeckel: > This is not right what you are saying. I am very careful not to break stuff. I adhere to a well defined boundary of what can be changed and what cannot.

You're right from many aspects, but from a user perspective you're wrong. This change breaks users' implementations if they rely on the fact that they can override XMALLOC etc by defining XMALLOC at compile time of the library.

This entire stuff has been public for ages, then it has been "private" for some time, now we should make it (as possible) really private but this just needs some time (and at least a 1.2.0 release which announces the deprecation of all this stuff in 2.0.0).

> But for example just now, mp_balance_mul was added which is actually a private function and should be called s_mp_balance_mul. By mistake, the API was extended. Such things are to be avoided.

I'm talking about releases. What's in develop is not yet released. 
Yes this should've been `s_mp_balance_mul`, and we can still change it with very little effort, as it also should be `s_mp_toom_mul` etc. but we have to wait until 2.0.0 to fix this.

Which reminds me that we should start to deprecate parts of the public API as well after this PR is merged.

sjaeckel: @czurnieden @buggywhip what's your opinion on this?

I rethought this entire topic and am now tending towards accepting as proposed by @minad and if someone has overridden these private things, well then he has to repair it...

sjaeckel: regarding the deprecated API functions we're forced to keep them, as otherwise we'd have to bump the ABI

czurnieden: @sjaeckel 

> am now tending towards accepting as proposed

A clean-up is necessary, no question, and we have to start at one point in time and (code)space so why not here and now?

> at least a 1.2.0 release which announces the deprecation of all this stuff in 2.0.0

Sounds good to me. And gives enough time to "round the edges".

minad: @sjaeckel Did you decide to merge this? If you do, ping me and I will update this PR again.

sjaeckel: thanks, yes please rebase

btw. would you be fine with me rebasing your commits if I sign them?

minad: Yes, you can also do the rebase without signing, that's ok. You are the committer.

minad:
@sjaeckel rebased, please merge!

sjaeckel:
> @sjaeckel rebased, please merge!

but not on develop

minad:
@sjaeckel Right I confused a few branches. Now it should be ok.

sjaeckel:
> @sjaeckel Right I confused a few branches. Now it should be ok.

I'll merge as soon as travis is done

buggywhip:
> On 9Apr, 2019, at 4:17 AM, Steffen Jaeckel <notifications@github.com> wrote:
> But renaming something that was there for ages, only for the sake of consistency, without any other gain, I don't consider as a such important improvement.
> 
Agree


> On 9Apr, 2019, at 4:40 AM, Steffen Jaeckel <notifications@github.com> wrote:
> 
> ... This change breaks users' implementations if they rely on the fact that they can override XMALLOC etc by defining XMALLOC at compile time of the library.
> 
> This entire stuff has been public for ages, then it has been "private" for some time, now we should make it (as possible) really private but this just needs some time (and at least a 1.2.0 release which announces the deprecation of all this stuff in 2.0.0).
I concur.  Some consistency across LTM and LTC is desirable, especially for otherwise-common low-level utilities like the MALLOC family.  While I personally don't rely on XMALLOC in this way, I easily could when my situation warranted.  

That said, I suppose I could support both XMALLOC and MP_MALLOC in tommath_private.h.  Then, after a stronger case can be made, perhaps deprecate XMALLOC for later removal?

minad:
@buggywhip I don't think it is easy to support both macros. You have to check if one or the other is defined etc.  Then you have to add some deprecated pragmas as in #223. This is not worth it. People who overwrite the macros of the **private** API, know what they are doing and they will accept if they change (I am overwriting them for example).

sjaeckel:
> Some consistency across LTM and LTC is desirable, especially for otherwise-common low-level utilities like the MALLOC family.

I like that thought! probably it'd be better to call it `LT_MALLOC` etc. and then also rename it in libtomcrypt

> People who overwrite the macros of the **private** API, know what they are doing and they will accept if they change

I think so too, even if they're already there for a long time people who overwrite them will (hopefully) know what they do and well then they have to adapt their sources...

minad:
> I like that thought! probably it'd be better to call it LT_MALLOC etc. and then also rename it in libtomcrypt

I am not very fond of it. I see tommath and crypt as distinct. Right now I am also only using tommath (maybe also crypt in the future). I would prefer if we can unify all the tommath macros and functions under the mp_\* prefix for consistency. I think consistency within the library is more important than consistency between two distinct libraries.

minad:
We could however rename everything to ltm_ and ltc_ and then you can use the lt_ macro, but I think that would really classify as needless renaming.

sjaeckel:
ah I don't know how to name it, but for sure we have to change the signature of the XMALLOC  class of macros in ltc as well to the one that is used here...

minad:
I haven't looked at ltc but I would just use the canonical namespace prefix if there is one?

sjaeckel:
they're called the exact same as they were called here before XMALLOC XFREE etc.

ltm and ltc were once one library and then they got separated... ;-)

buggywhip:
> On 12Apr, 2019, at 2:06 PM, Steffen Jaeckel <notifications@github.com> wrote:
> 
> ah I don't know how to name it


Consistency does not have to be sameness.  You can have consistency at the paradigm level, and perhaps that is all that is needed here.

PR #173, opened 2019-03-07
Proposal of changes to #160

sjaeckel:
This is an update of #160 but I didn't know if that's really what we want, so I opened a separate PR

### DIFF
--- a/demo/test.c
+++ b/demo/test.c
@@ -5,14 +5,17 @@
 #ifndef MIN
 #define MIN(x, y) (((x) < (y)) ? (x) : (y))
 #endif
-static double s_abs(double s_d){
+static double s_abs(double s_d)
+{
    return (s_d < 0.0)?-s_d:s_d;
 }
-static float s_absf(float s_d){
+static float s_absf(float s_d)
+{
    return (s_d < 0.0f)?-s_d:s_d;
 }
 #include <float.h>
-static int s_compare_doubles(double s_a, double s_b){
+static int s_compare_doubles(double s_a, double s_b)
+{
    double abs_a, abs_b, delta;
 
    /* NaN, inf's and subnormals ignored, not needed in this case*/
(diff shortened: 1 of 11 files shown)


demo/test.c
```diff
@@ -515,13 +443,127 @@ static int test_mp_invmod(void) {
...
+static int s_compare_doubles(double s_a, double s_b)
+{
+   double abs_a, abs_b, delta;
+
+   /* NaN, inf's and subnormals ignored, not needed in this case*/
+
+   if (s_a == s_b) {
+      return 1;
+   }
+#ifdef LTM_WARN_X87_EXT_PREC
```
sjaeckel: @czurnieden this define appears nowhere else what should we do with it?

czurnieden:
Thanks for the clean-up!

> This is an update of #160 but I didn't know if that's really what we want

Assuming "that" means #160 : it's not much work to keep it "mergeable" (as long as I don't eff it up again ;-) ) so that the decision can be restricted to "moving it into the main version or not". It helps to avoid arguments like "Yeah, would like it but there is a lot of work to do before we can include it!", which are rarely helpful.

> run `make astyle`

Was sure I told my little editor to run it before saving? Ah, no it isn't able to do it, run a program before saving that is, my misunderstanding, sorry.

demo/test.c
czurnieden: It's a bug: "Functionality not documented."

Didn't want to put it in the makefiles for the rule `test_standalone`, too much output, so I made it optional, as a kind of `--verbose`.

sjaeckel:
> Assuming "that" means #160 

hrhr, "that" meaning "the changes in this PR" :-D 

especially the stuff regarding Valgrind

> Was sure I told my little editor to run it before saving

we could add a pre-commit hook!?

In general: if you think this PR is okay and doesn't need any further changes, simply apply the patches to your branch and we can close this

demo/test.c
sjaeckel: OK

sjaeckel:
You can install the hook with `make install_hooks`

If you try to commit it will run `make astyle` and if that had to format something it will stop the commit so you can fix it up

czurnieden:
> we could add a pre-commit hook!?

Yes, that is a nice one, thanks!

> if you think this PR is okay and doesn't need any further changes, simply apply the patches to your branch

Done, together with a word or two regarding the x87 warning macro in the documentation.
I think that was all by now.

PR #175, opened 2019-03-11
Added valgrind tests to testme.sh and travis.yml

czurnieden:
It has been hinted at by somebody here, that the integration of a memory checker like e.g.:  [Valgrind](http://valgrind.org/) would be appreciated.

It extends the runtime of the Travis check to about 10 minutes per job.

The Travis-timeout is at 50 minutes per job and the total time is not counted (yet), so the jobs can be split further (e.g.: the low-mp one) to reduce the total runtime but the highest amount of parallel jobs I encountered was 6 (six) and just 4 (four) on average.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,35 +1,133 @@
-language: c
+#############################################################################
+#                                                                           #
+#      Travis-ci test-suite for LibTomMath                                  #
+#      (https://github.com/libtom/libtommath.git)                           #
+#                                                                           #
+#############################################################################
 
-install:
-    - sudo apt-get update -qq
-    - sudo apt-get install gcc-multilib
+# Compilation failures are in gcc_errors_*.log
+# Failed tests in test_*.log
+# Files do not exist in case of success
+after_failure:
+  - cat test_*.log
+  - cat gcc_errors_*.log
 
-matrix:
-  fast_finish: true
+# In case of a Travis error a success might get signaled
+# even without any test run. This file also keeps any notes
+# printed from the tests which might come handy from time
+# to time.
+# Valgrid will print its output to stderr which will not show up
+# in test_*.log. testme.sh accepts one additional option to
+# valgrind and "--valgrind-options=--log-fd=1" sends the output
+# of Valgrind to stdout instead.
+after_success:
+  - cat test_*.log
+
+# Tests restricted to the following branches of LTM.
 branches:
   only:
     - master
     - develop
     - /^release\/.*$/
 
-compiler:
-  - gcc
-  - clang
-script:
-  - ./testme.sh --with-cc=$CC ${BUILDOPTIONS}
-env:
-  - |
-    BUILDOPTIONS="--test-vs-mtest=333333"
-  - |
-    BUILDOPTIONS="--test-vs-mtest=333333 --mtest-real-rand"
-  - |
-    BUILDOPTIONS="--with-low-mp"
-  - |
-    BUILDOPTIONS="--with-m64 --with-m32 --with-mx32"
+# Additional installs are Valgrind for the memory-tests
+# and gcc-multilib for the compilation of the different
+# architectures.
+install:
+  - sudo apt-get update -qq
+  - sudo apt-get install valgrind
+  - sudo apt-get install gcc-multilib
 
-after_failure:
-  - cat test_*.log
-  - cat gcc_errors_*.log
+# The language is C and it will load the respective dependencies
+language: c
 
+# The actual workspace. Will run the individual jobs in parallel
+# which also means that the jobs must be able to run in parallel.
+# Either specify sets which will be combined or, as in this case,
+# specify all builds individually. The number of jobs is currently
+# restricted to 200 jobs at most.
+matrix:
+  # Will mark as finished if all of the remaining tests are allowed to fail
+  # or one test has failed already.
+  fast_finish: true
+
+  # The individual jobs
+  include:
+    # The environment given to the programs in the build
+    # We have only one program and the variable $BUILDOPTIONS
+    # has only the options to that program: testme.sh
+
+    # GCC for the 32-bit architecture (no valgrind yet)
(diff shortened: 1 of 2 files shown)


czurnieden:
OK, that needs a bit more work for the `-m32` and `-mx32` avlgrind tests, I'm afraid.

czurnieden:
The ggc build for `--with-low-mp` fails repeatable with a Travis timeout. The runtime of the `MP_8BIT ` clang build is already at 9:41minutes which is pretty close to Travis' 10 minutes timeout.  They have a small tool at hand `travis_wait` to extend it to 20 minutes. Question is where to put it.

czurnieden:
OK. Works.
May I `git rebase -i 8972afe023fd1c4c2584a36d7b19df9149493a46` now?

sjaeckel:
>  May I `git rebase -i 8972afe023fd1c4c2584a36d7b19df9149493a46` now?

IMO the easiest would be

```
$ git branch pr175beforereset
$ git reset --hard 8972afe023fd1c4c2584a36d7b19df9149493a46
$ git merge --squash pr175beforereset 
$ git add -u
$ git commit --amend
```

maybe the changes also need some `'s/[[:blank:]]*$//'` :-)

czurnieden:
Yepp, that took some doing ;-)

What was assumed to be a short half hour hack including the time waiting for Travis to run and the inevitable extra two commits to correct the typos has gone mad and took what, 30 commits?

If I would have known that I would have [installed Travis locally](https://stackoverflow.com/questions/21053657/how-to-run-travis-ci-locally) instead of pestering this pull with commits&mdash;but hindsight is always 20/20, as the saying goes.
Based on that hindsight I would also strongly recommend to do so if you want to change `.travis.yml` if it is more than a correction of a single-character typo.
The Travis documentation is rather basic, a bit spread out and not really up to the latest date. The parser used for .`travis.yml`seems to be [Yaml.rb](https://yaml.org/YAML_for_ruby.html) and  is whitespace-sensitive. A linter would be helpful but Travis' [online linter](https://lint.travis-ci.org/) is not working.
One can install the online-linter locally but it needs a newer Ruby version, more modern (> 2.0)  than I had installed. Wasn't able  to update it, had to use a VM instead, hence the funny email address `czunieden@debian64.debian64` for most of the commits. Found out quickly that this linter was outdated with a lot of false positives, nearly useless. Tried another, more general Yaml linter online that also "tidied" the code which I didn't recognise immediately, needessly causing another 2 or 3 commits.

Still not done?
Still not done.

The `-m32` needs a 32-bit valgrind because the 64-bit valgrind cannot run 32-bit binaries. Multiarch needs a bit more work than just installing the `*:i386` libs and binaries. Not much, at least not with Debian but Travis runs Ubuntu images, which is based on Debian but nevertheless slightly different.
Needs some investigation, hence moved to a later date.

The `-mx32` ([The X32 architecture](https://sites.google.com/site/x32abi/home)) seems to work but not with valgrind. The webpage mentioned above is not very active ([last update](https://sites.google.com/site/x32abi/system/app/pages/recentChanges)). Actual testing would need a [x32-activated linux-kernel](https://wiki.debian.org/X32Port) plus all the x32-programs and libraries. If the Travis kernels are activated is unknown but highly unlikely.
I concur with the rest of the Libtommath group here and move it to a later date. I would even propose to skip the test, because the result is not very conclusive without running it in a X32 environment. The only error it can show in LibTomMath is if somebody assumed 64-bit `long`s on a x86_64.

I would propose three other tests instead of the `-m32`, '-m64', and `-mx32`, namely arm32, arm64, and probably even [arm64_32](https://wiki.debian.org/Arm64ilp32Port). Travis [supports some embed boards](https://docs.travis-ci.com/user/integration/platformio/) by way of [Platform.io](https://platformio.org/platforms). Implementation seems [simple enough](http://docs.platformio.org/en/latest/ci/travis.html).

Travis also offers Windows and MacOS builds. LibTomMath says it builds on Windows and MacOS and  we should test if it does what it says, shouldn't we?

Aaand after some trouble with the editor eating characters (actually a problem with X in the VM) and my inability to get valgrind to eat its options from the `testme.sh` commandline I ended up with 30 commits and a comment commenting that fact that grew waaay too long.

Let's see if I'm able to squash those commits w/o effing it up despite having nothing more to do than to dutifully run @sjaeckel 's recipe.

> maybe the changes also need some 's/[[:blank:]]*$//' :-)

I don't see any reasons they couldn't have made a grammar for the blocks that doesn't rely on "invisible syntax", as I like to call it, but [they did](https://yaml.org/spec/1.2/spec.html#id2769866).
But that's not what you meant, I know. Just felt I had [to mention it](https://beebo.org/haycorn/2015-04-20_tabs-and-makefiles.html) again ;-)

OK, will clean it up, sprinkle some comments for later generations, squash it, and push it and go on a long vacation immediately.
No, wait, let me see if my formatting and commenting broke it first. Highly unlikely but I have some bad experiences in that regard, y'know ;-)

czurnieden:
This thing is really working my last nerve!

czurnieden:
@sjaeckel 
Hu?
The only thing you changed in the last commit is the exit code in case of an erroneous command-line option-parameter for `testme.sh`.
(BTW: What is the reason for line 229?)
Ah, [b90a28a](https://github.com/libtom/libtommath/pull/175/commits/b90a28a825f39fb62c5bb7afd8e8f23634838ae2) brought `_timeout` back from upper-case-land (my bad, sorry!) which is set to 900 seconds, which are 15 minutes and Travis ended after 15-16 minutes runtime. That runtime includes the setup, so it is most likely that timeout. Travis runtimes vary wildly over the day (and week), problems with timeouts may not come up immediately.

What to do?
Either split that thing into several smaller parts (that is in seperate Travis jobs, max is 200) which is a bit of work or set the timout higher, say 30 minutes (or 1800 seconds if that timeout(1) Travis uses is different from the one here which accepts the m, h, d modifiers) which can trigger, too, but I think 1800s should suffice and: it doesn't need any hard work.

Do you want to do it or shall I?

PR #178, opened 2019-03-22
add missing const to cast

minad:


minad:
@czurnieden There was a failure in the prime tests. Has something like that happened before?

czurnieden:
@minad `mp_prime_miller_rabin` doesn't like it when input and witness are the same. Can happen in the random fips rounds.

minad:
Thx!

PR #179, opened 2019-03-23
avoid ctype toupper

minad:
- libc toupper is locale dependent
- removes the dependency on ctype.h

### DIFF
--- a/bn_mp_read_radix.c
+++ b/bn_mp_read_radix.c
@@ -12,6 +12,8 @@
  * SPDX-License-Identifier: Unlicense
  */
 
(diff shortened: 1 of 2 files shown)


minad:
@sjaeckel Are there objections to #176, #177, #178, #179? I split them to make merging and checking easy. Maybe it seems that there is more to it than there is? Do you need some words about the motivation? While these are very minor nitpicks, these are issues I have to fix by patching tommath in a downstream project. I am working on a unikernel project, where there is essentially no libc (no stdio, no ctype etc). tommath compiles well there except for those issues :)

PR #180, opened 2019-03-23
Fixed bug in mp_prime_isprime

czurnieden:
Random witness to Miller-Rabin cannot be equal to input to MIller-Rabin

### DIFF
--- a/bn_mp_prime_is_prime.c
+++ b/bn_mp_prime_is_prime.c
@@ -335,19 +335,19 @@ int mp_prime_is_prime(const mp_int *a, int t, int *result)
           * smaller than or equal to "a"
           */
          len = mp_count_bits(&b);
-         if (len >= size_a) {
-            /* Witness and test subject must not be equal */
-            if( (len == size_a) && (mp_cmp(a, &b) == MP_EQ) ) {
-               len++;
-            }
+         if (len > size_a) {
             len = len - size_a;
             if ((err = mp_div_2d(&b, len, &b, NULL)) != MP_OKAY) {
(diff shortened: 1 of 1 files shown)


minad:
thank you for looking into it so quickly :+1:

czurnieden:
The alternative is based on my first idea. It is much faster but it wastes 1 bit of entropy. On the other hand: it is a very rare event for larger numbers.

czurnieden:
@sjaeckel this is a security fix. Is it grave enough to go into the current version (assuming the patch (ffec6f1) is correct and accepted by the rest of the crew, of course!) instead of `develop`?

sjaeckel:
> @sjaeckel this is a security fix. Is it grave enough to go into the current version (assuming the patch ([ffec6f1](https://github.com/libtom/libtommath/commit/ffec6f1447b532c5d83a6055aa4337fe01c58d89)) is correct and accepted by the rest of the crew, of course!) instead of `develop`?

we can go through the PR's merged since the last release and make a bugfix-release, I see more than this patch in there!

I've created a new milestone [v1.1.1](https://github.com/libtom/libtommath/milestone/4) and will later today go through the PR's since v1.1.0 and assign them to this milestone, feel free to do so as well!

sjaeckel:
> I've created a new milestone [v1.1.1](https://github.com/libtom/libtommath/milestone/4) and will later today go through the PR's since v1.1.0 and assign them to this milestone, feel free to do so as well!

I will then cherry-pick all those PR's to the to-be-created v1.1.1 branch

PR #181, opened 2019-03-24
efficienty enhancements for get/set routimes

nijtmans:
mp_get_bit() doesnt handle well when limb >= a->used:  It should return 0 in that case, not MP_VAL.

The MP_SET_XLONG() macro is quite inefficient, doing only 4 bits at a time in the loop. I rewrote this macro, handling DIGIT_BIT's at a time in each loop iteration.

Also I would like to add some (unsigned long) or (unsigned long long) casts: The lack of those caused some MSVC warnings in my environment.

### DIFF
--- a/bn_fast_s_mp_sqr.c
+++ b/bn_fast_s_mp_sqr.c
@@ -79,7 +79,7 @@ int fast_s_mp_sqr(const mp_int *a, mp_int *b)
       }
 
       /* store it */
-      W[ix] = _W & MP_MASK;
+      W[ix] = (mp_digit)_W & MP_MASK;
 
       /* make next carry */
       W1 = _W >> (mp_word)DIGIT_BIT;
--- a/bn_mp_get_bit.c
+++ b/bn_mp_get_bit.c
@@ -27,18 +27,8 @@ int mp_get_bit(const mp_int *a, int b)
 
    limb = b / DIGIT_BIT;
 
-   /*
-    * Zero is a special value with the member "used" set to zero.
-    * Needs to be tested before the check for the upper boundary
-    * otherwise (limb >= a->used) would be true for a = 0
-    */
-
-   if (IS_ZERO(a)) {
-      return MP_NO;
-   }
-
    if (limb >= a->used) {
-      return MP_VAL;
+      return MP_NO;
    }
 
    bit = (mp_digit)(1) << (b % DIGIT_BIT);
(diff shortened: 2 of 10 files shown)


nijtmans:
Hm... that doesn't work either, now it's even worse ....  Suggestions how to fix this welcome!

czurnieden:
@nijtmans The casting works if and only if `long` is smaller than, or equal to  `mp_digit`. So a check like `if(sizeof(long) < sizeof(mp_digit))...` is necessary. That, more or less, doubles the code because you need to branch very early. A good compiler might notice (`sizeof(type)` is known at compile time) and optimizes the `false` branch away.
Mmh&hellip;
Or you check if `c < b` after `mp_digit c = b`  which only doubles the loop (one version with `c` if `c` is large enough and one with `b` if not) but the compiler, though it can  detect it and utter a warning, would not dare optimizing the branch away automatically.

czurnieden:
@nijtmans The style here at LibTomMath (forgot who introduced it) is to make all implicit casts explicit. So your `mp_digit c = b` should be `mp_digit c = (mp_digit) b`. Otherwise&hellip;how do I make that thumbs-up smiley? Ah, here it is: :+1:

nijtmans:
Thanks, @czurnieden !  Actually, I'm not satisfied myself yet ...   even though it works now. Being continued...

czurnieden:
@nijtmans the other two points (in `bn_mp_get_bit.c` and MSVC'S problem with the casting) are proper bugfixes. Please think about snipping them out (or simply do another commit without the changes in `tommath_private.h`), they would get an immediate stamp of approval, at least from my side.

nijtmans:
Well, thinking about it, the only situation to be handled differently is when sizeof(long)==32 and DIGIT_BIT == 60.    Since sizeof(long long) is guaranteed to be at least 64 bits, I now think I have a better compromize.

@czurnieden, please continue with your good remarks!  I intend to make it as 'perfect' as possible, and the submit for inclusion. That shouldn't take more than a few days, I think.

Again, Thanks!

nijtmans:
Well, the build finally passed. Any more remarks, before it gets full approval?

czurnieden:
@nijtmans I'm OK with it but I'm not the one who makes the final decisions.
It is also possible that not everyone likes the mix of a function macro and a proper function, Basically the same code, twice, in two different places for a minimal gain, they might say.
But I don't care. It works, doesn't change the API, doesn't bloat the binary, doesn't bloat the source too much, and tickles a small gain out of it at the end, so I approve.

nijtmans:
This Pull Request is split into 4 different Pull Requests now, which can be reviewed separately. So, this one can be closed.

PR #182, opened 2019-03-26
bug-fix in bn_mp_get_bit.c: handle overflow correctly

nijtmans:
Split off from "efficienty enhancements for get/set routines", as a separate issue.


nijtmans:
This looks like a usefull addition to the 1.1.1 milestone.

PR #189, opened 2019-04-01
changed seed to make nth-root usable

czurnieden:
Old implementation made it unusable, especially for tests for (perfect) power which test all prime indices up to log_2(n). This patch does not change the API. It increases the binary size by ca. 450 bytes, about 16%.


bn_mp_n_root_ex.c
```diff
@@ -89,14 +129,39 @@ int mp_n_root_ex(const mp_int *a, mp_digit b, mp_int *c, int fast)
...
+          got stuck, so break out of the loop and do the rest manually.
+       */
+      if (ilog2-- == 0) {
+         break;
+      }
    }  while (mp_cmp(&t1, &t2) != MP_EQ);
 
    /* result can be off by a few so check */
+   /* Overshoot by one if root is smaller */
+   for (;;) {
```
sjaeckel: why is this `for` loop there? AFAIU this will never repeat!?

czurnieden: > AFAIU this will never repeat!?

Neither do I think it will, but it is just a conjecture and should not be trusted until proven.
A slightly different case was my first try of adjusting the seed, although with Halley's algorithm instead of Newton's, where I actually measured the errors. I can't find the raw data now but I asked cs.stackexchange including a plot of that data at https://cs.stackexchange.com/questions/56848/numerical-stability-of-halleys-recurrence-for-integer-n-mathrmth-root Yepp, to no avail ;-)

Oh, I just recognised I missed an "s" in the comment.  It should have said
"[The loop underneath ]Overshoot**s** by one if root is smaller"
Will make it a bit clearer if you insist.

sjaeckel: I should've gone to bed already long ago...

sjaeckel:
While rebasing I took the liberty to take the https://github.com/libtom/libtommath/pull/190#issuecomment-479821400 of @minad into account

demo/test.c
```diff
@@ -1362,6 +1363,288 @@ static int test_mp_decr(void)
...
+   mp_int a, c, r;
+   int e;
+   int i, j;
+
+   const char *input[] = {
+      /*
+       * 7695055368100873513971520489282333897557005032980617084180385070\
+       * 39530690836861445204045548919181252491201760744537260787 x
+       */
+      "4n9cbk886QtLQmofprid3l2Q0GD8Yv979Lh8BdZkFE8g2pDUUSMBET/+M/YFyVZ3mBp",
```
minad: Why are you writing the numbers in base64 and add them in decimal as a comment? Either write them only in base64 or just write them in decimal which I would prefer.

demo/test.c
```diff
@@ -1284,7 +1284,8 @@ static int test_mp_incr(void)
       goto LTM_ERR;
    }
    a.sign = MP_ZPOS;
-   if (mp_cmp_d(&a, MP_MASK) != MP_EQ) {puts("DDD");
+   if (mp_cmp_d(&a, MP_MASK) != MP_EQ) {
+      puts("DDD");
```
minad: DDD

sjaeckel: DDDaumen hoch

czurnieden: To paraphrase @sjaeckel : I should have gone to bed a long time before that! ;-)

sjaeckel: > DDDaumen hoch

czurnieden:
> Could you describe a bit more clearly what issue is being solved here? Is there a convergence problem with the old seed?

The old seed makes it abysmally slow. It is just not usable with that seed. See [this "paper"](https://github.com/czurnieden/libtommath/blob/nthroot/doc/new/) for some illustrated details.

demo/test.c
czurnieden: > Why are you writing the numbers in base64 and add them in decimal as a comment? 

For the same reason I left the tracing `puts("DDD")` in it ;-)

I calculated them in Pari/gp and converted them later with an external program to shorten the code lines. It was just a list that I ticked off (the little "x" at the end of the decimal numbers is that tick), not meant as a translation.

> [...] or just write them in decimal which I would prefer.

I just don't like the overly long lines (not every compiler accepts string constants in multiple parts). OCD maybe?

minad: Yes, but you can also split the decimal constants?

minad: Wat? I didn't read. Not every compiler supports multiple string constants?! Those are unsupported. I hardly believe that tommath would compile with such a thing since it is quite common practice to stringize and concatenate in macros.

czurnieden: > Not every compiler supports multiple string constants?

No, " multiple string constants " is not the right nomenclature, I mean things like this, taking `printf`  as an example
```
printf( "this shall mimic a %s long "
          "line, cut in %c\n", "very", '2');
```
I vividly remember all the curses I uttered, although it was a long time ago. Nevertheless:

> Those [compilers] are unsupported.

The documentation has no restrictions in that regard. It includes the old Borland C compiler and even MSVC is supported. We should set some strict limits, e.g (versions #s arbitrary): GCC > 3, clang > 3.5, MSVC > 2012, icc > 19 and so on (mingw, djgpp, &hellip;), document them and test them. The exact version #s are open for discussion, of course. But not here, I think it deservers its own thread, pardon, issue.

czurnieden:
Uhm, can somebody here tell me what Travis was about to tell me but didn't?

sjaeckel:
> Uhm, can somebody here tell me what Travis was about to tell me but didn't?

oops is your confusion caused by me canceling the build job? :)

czurnieden:
> oops is your confusion caused by me canceling the build job? :)

Well, as someone who broke the build not only once&hellip; ;-)

What is it "Conflicting files demo/test.c"? Ah, I see, have already resolved it while applying @minad 's suggestions, just waiting for the local tests to finish to push it.

czurnieden:
@minad 
> Could you describe a bit more clearly what issue is being solved here?

A concrete example is #201 

timings of the isolated mp_expt test which checks the result of `mp_expt` by checking the root.

with orig. n-root:  killed after 75 minutes
with bugfix-n-root: 2.5 seconds
with spaceforspeed[1]: 0.35 seconds

[1] found a bug in that version, fix not yet pushed but used in that test

sjaeckel:
as you're online @czurnieden I'd say we start with this branch to be rebased and then merged next ...or #204 ...whoever is faster

sjaeckel:
can you please squash those fixup commits together and force-push?

the "workaround" commit should stay so it can be reverted in #207

czurnieden:
> those fixup commits together

@sjaeckel  Argh, **only** the fixes you said. Sorry, squashed too much, I'm afraid.

sjaeckel:
> @sjaeckel Argh, **only** the fixes you said. Sorry, squashed too much, I'm afraid.

perfect like that

PR #190, opened 2019-04-03
additional functions mp_incr and mp_decr

czurnieden:
First batch.
The functions are interdependent, they call each other in case of negative input.

### DIFF
--- a/bn_mp_decr.c
+++ b/bn_mp_decr.c
@@ -16,7 +16,7 @@
 /* Decrement "a" by one like "a--". Changes input! */
 int mp_decr(mp_int *a)
 {
-   int e;
+   int e = MP_OKAY;
    if (IS_ZERO(a)) {
       mp_set(a,1uL);
       a->sign = MP_NEG;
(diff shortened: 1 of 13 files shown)


minad:
Looks good except for the whitespace changes.

czurnieden:
> Looks good except for the whitespace changes.

But @sjaeckel was very nice and helped me to install a hook to&hellip;oh, that was only for that branch not for all. My bad, sorry.

I'm currently building the PR for `mp_ilogb` which won't need long (hopefully ;-) ) and clean up when that is done.

sjaeckel:
which whitespace changes?

sjaeckel:
> What is the reason for that restriction?

Tests should only use the public API. For your case there's still `mp_iszero()`... but I can also live with `mp_cmd_d()`

sjaeckel:
> which whitespace changes?

@minad \*ping\*

minad:
@sjaeckel I was on mobile and I think I confused the two PRs by @czurnieden. See here https://github.com/libtom/libtommath/pull/191/files#diff-aa67c18ca27d74fae0446b5cd9376391L14.

PR #191, opened 2019-04-03
New function:  ilogb, integer logarithm to integer base

czurnieden:
Can also be used to decrease runtime for `mp_radixsize` from O(n^2) to O(log n) (`mp_ilogb` uses a simple binary search). Some empirical tests show a cutoff useing `mp_ilogb` at about 100 limbs (MP_64BIT) which is a lot.

### DIFF
--- a/bn_mp_ilogb.c
+++ b/bn_mp_ilogb.c
@@ -1,0 +1,183 @@
+#include "tommath_private.h"
+#ifdef BN_MP_ILOGB_C
+/* LibTomMath, multiple-precision integer library -- Tom St Denis */
+/* SPDX-License-Identifier: Unlicense */
+
+/* Compute log_{base}(a) */
+static mp_word s_pow(mp_word base, mp_word exponent)
+{
+   mp_word result = 1uLL;
+   while (exponent) {
+      if ((exponent & 0x1) == 1) {
+         result *= base;
+      }
+      exponent >>= 1;
+      base *= base;
+   }
+
+   return result;
+}
+
+static mp_digit s_digit_ilogb(mp_digit base, mp_digit n)
+{
+   mp_word bracket_low = 1uLL, bracket_mid, bracket_high, N;
+   mp_digit ret, high = 1uL, low = 0uL, mid;
+
+   if (n < base) {
+      return (mp_digit)0uL;
+   }
+   if (n == base) {
+      return (mp_digit)1uL;
+   }
+
+   bracket_high = (mp_word) base ;
+   N = (mp_word) n;
+
+   while (bracket_high < N) {
+      low = high;
+      bracket_low = bracket_high;
+      high <<= 1;
+      bracket_high *= bracket_high;
+   }
+
+   while (((mp_digit)(high - low)) > 1uL) {
+      mid = (low + high) >> 1;
+      bracket_mid = bracket_low * s_pow(base, mid - low) ;
+
+      if (N < bracket_mid) {
+         high = mid ;
+         bracket_high = bracket_mid ;
+      }
+      if (N > bracket_mid) {
+         low = mid ;
+         bracket_low = bracket_mid ;
+      }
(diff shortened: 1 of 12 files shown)


bn_mp_ilogb.c
```diff
@@ -0,0 +1,192 @@
...
+      }
+      if (cmp == MP_EQ) {
+         mp_set(c, mid);
+         goto LBL_ERR;
+      }
+   }
+
+   if (mp_cmp(&bracket_high, a) == MP_EQ) {
+      mp_set(c, high);
+      goto LBL_ERR;
```
sjaeckel: this can go

bn_mp_ilogb.c
```diff
@@ -0,0 +1,192 @@
...
+   } else {
+      mp_set(c, low);
```
sjaeckel: same

bn_mp_ilogb.c
```diff
@@ -0,0 +1,192 @@
...
+       Idea shamelessly stolen from https://programmingpraxis.com/2010/05/07/integer-logarithms/2/
+       The effect is asymptotic, hence needs benchmarks to test if the Giant-step should be skipped
+       for small n.
+    */
+   while (mp_cmp(&bracket_high, a) == MP_LT) {
+      low = high;
+      if ((err = mp_copy(&bracket_high, &bracket_low)) != MP_OKAY) {
+         goto LBL_ERR;
+      }
+      high <<= 1;
```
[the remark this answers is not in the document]
czurnieden: Can it overflow?
I don't think so but "I dont't think so" is not good enough, of course. Will check it.

bn_mp_ilogb.c
czurnieden: What? Oh, yes, thanks.

bn_mp_ilogb.c
```diff
@@ -0,0 +1,192 @@
...
+   if (bracket_high == N) {
+      ret = (mp_digit) high;
+   } else {
+      ret = (mp_digit) low;
+   }
+
+   return ret;
+}
+
+/* TODO: output could be "int" because the output of mp_radix_size is int, too,
```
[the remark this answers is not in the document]
czurnieden: No, that overflow problem is deep in LTM itself, as the comment describes. and it needs a an API change to solve it: change the return of `mp_radix_size`, `mp_bitcount` and all the other functions counting bits to a type that is larger than `int`.

bn_mp_ilogb.c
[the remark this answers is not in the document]
czurnieden: Changed it to a type larger than `signed int`and added a test if `mp_expt_d` can work with it (that was one of the reasons I wrote `mp_expt`)

bn_mp_ilogb.c
```diff
@@ -0,0 +1,182 @@
...
+         high = mid;
+         mp_exch(&bracket_mid, &bracket_high);
+      }
+      if (cmp == MP_GT) {
+         low = mid;
+         mp_exch(&bracket_mid, &bracket_low);
+      }
+      if (cmp == MP_EQ) {
+         mp_set_int(c, (unsigned long)mid);
+         goto LBL_ERR;
```
sjaeckel: so this `LBL_ERR` is not only used for errors? ;-)

czurnieden: > so this LBL_ERR is not only used for errors? ;-)

Works also very well as a `break` ;-)
Oh my&hellip; \*sigh\*
(probably too much Old Fortran over the last weeks)
Thanks!

> ...if you change it, can you please squash the commits?...

Will do.

Oh, François has "more linting". Seems as if I got the whole broadside alone this time ;-)

czurnieden: Oh, you nearly had me there! ;-)
But changed it nevertheless to `LTM_END` while I was at it.

PR #195, opened 2019-04-04
addition of function mp_pollard_rho

czurnieden:
Another one ripped out of the large packet at #190


minad:

PR #199, opened 2019-04-04
remove mp_set_long specialization

minad:
Taken out from #194


minad:
@sjaeckel I think @czurniden looked already at the old PR. This is just extracted exactly.

PR #201, opened 2019-04-04
addition of mp_expt (all input bigint)

czurnieden:
An exponentiation function like `mp_expt_d` but with all input having the type `mp_int`.
Needed for low-MP.

The test for this function will time-out because it uses the original `mp_n_root_ex` function. Works well with the version in #189

### DIFF
--- a/bn_mp_expt.c
+++ b/bn_mp_expt.c
@@ -1,0 +1,51 @@
+#include "tommath_private.h"
+#ifdef BN_MP_EXPT_C
+/* LibTomMath, multiple-precision integer library -- Tom St Denis */
+/* SPDX-License-Identifier: Unlicense */
+
+/* exponentiate a^b = c with a, b, c big integers */
+int mp_expt(const mp_int *a, const mp_int *b, mp_int *c)
+{
+   mp_int tmp,e;
+   int err;
+
+   /* only positive exponents for now */
+   if (b->sign == MP_NEG) {
+      return MP_VAL;
+   }
+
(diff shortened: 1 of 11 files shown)


czurnieden:
Mmh&hellip;for some reason the function to add the leading and trailing comments to the new files in dep.pl doesn't work anymore? The backslashes for the dollar signs in the end-comment gets printed! Works without the backslashes. WTF?
And who did it? Oh, me. \*grrr\*
OK, will check it out.

czurnieden:
Ah, that was mean! ;-)
Oversaw the [single quotes](https://www.perlmonks.org/?node_id=736989)

minad:
@czurnieden Would it make sense to instead extend mp_expt_d_ex/mp_expt_d to accept larger exponents than mp_digit (maybe call them mp_expt_int_ex/mp_expt_int)? Then we could avoid this function.

czurnieden:
> Would it make sense to instead extend mp_expt_d_ex/mp_expt_d to accept larger exponents than mp_digit (maybe call them mp_expt_int_ex/mp_expt_int)? 

I found no easy way to do so.
And as most people use the "normal" `expt_d` function, the bigint one is more for the low-mp users, especially the 8-bitters,  I made it a separate function instead that you can exclude from the build if you don't need it.
But, as hinted on above: if you find a neat way to merge those two, I would be all in for it.

minad:
@czurnieden This one can be closed right? Instead we want mp_expt_int or sth similar.

PR #205, opened 2019-04-06
Add travis build-job that checks the source-code format

sjaeckel:
This adds a new `--format` option to `testme.sh` which runs all the source-code format checkers and the generators of `callgraph.txt` and `tommath_class.h` and adds a travis job which runs this.

This also fixes all the issues found in the current develop branch.


sjaeckel:
nice... astyle 2.03 and 2.05 format the sources differently...

PR #206, opened 2019-04-06
Multiplication balancing for the Toom-Cook algorithms

czurnieden:
Had been requested a long time ago (#2 ).

### DIFF
--- a/tommath.h
+++ b/tommath.h
@@ -335,6 +335,7 @@ int mp_sub(const mp_int *a, const mp_int *b, mp_int *c);
 
 /* c = a * b */
 int mp_mul(const mp_int *a, const mp_int *b, mp_int *c);
+int mp_balance_mul(const mp_int *a, const mp_int *b, mp_int *c);
 
 /* b = a*a  */
 int mp_sqr(const mp_int *a, mp_int *b);
--- a/tommath_class.h
+++ b/tommath_class.h
@@ -22,6 +22,7 @@
 #   define BN_MP_ADD_D_C
 #   define BN_MP_ADDMOD_C
 #   define BN_MP_AND_C
+#   define BN_MP_BALANCE_MUL_C
 #   define BN_MP_CLAMP_C
 #   define BN_MP_CLEAR_C
 #   define BN_MP_CLEAR_MULTI_C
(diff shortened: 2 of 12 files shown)


tommath_class.h
```diff
@@ -624,9 +636,10 @@
 #endif
 
 #if defined(BN_MP_MUL_C)
+#   define BN_FAST_S_MP_MUL_DIGS_C
```
sjaeckel: lulz looks like that comment mentioning "fast_s_mp_mul_digs" triggers `dep.pl`

czurnieden: Oh, great \*sigh\*
Had merge conflicts with `tommath_class.h`and `callgraph.txt` and thought the simplest way would be to just delete their content and let `make new_file` do its thing.
Which *seemed* to work.
What to do: change my comment or patch dep.pl to ignore comments (which is quite complicated if you want to get all edgecases)?

sjaeckel: >  Which _seemed_ to work.

it worked indeed.

> What to do: change my comment or patch dep.pl to ignore comments

very good question!
@karel-m you're the perl guy I know, is there an easy way to fix this in dep.pl?

tommath.h
```diff
@@ -335,6 +335,7 @@ int mp_sub(const mp_int *a, const mp_int *b, mp_int *c);
 
 /* c = a * b */
 int mp_mul(const mp_int *a, const mp_int *b, mp_int *c);
+int mp_balance_mul(const mp_int *a, const mp_int *b, mp_int *c);
```
minad: @czurnieden Is it intended that mp_balance_mul is exposed in the public api or should this rather go to tommath_private?

czurnieden: > Is it intended that mp_balance_mul is exposed in the public api

But *everything* is exposed?!
I know you fight it tooth and nails but LTM should make it explicit by using the methods offered by modern C-compilers and not by "hiding" it in another header.
But that discussion is for another day and another version.

> should this rather go to tommath_private

You really don't need to tiptoe around me if you want something done, I'm a friendly bear ;-)

But as this PR is already merged and it is only a small, non-essential problem I would like to put it piggyback  (but with a distinct commit for blaming) on one of the other PRs.

minad: @czurnieden I really care about the exposed API to be honest. This is what keeps the library nice and usable. To make this more clear private functions should be hidden symbols or at least be prefixed by s_.

I have some other minor comments incoming in a small PR. But no big deal.

PR #210, opened 2019-04-06
Corrected printing of footer (last_comment) in dep.pl

czurnieden:
As the titel suggests, a correction in the printing of the footer.

### DIFF
--- a/dep.pl
+++ b/dep.pl
@@ -81,9 +81,9 @@ EOS
    if ($apply == 1) {
       print {$out} << 'EOS';
 #endif
-/* ref:         \$Format:\%D$ */
-/* git commit:  \$Format:\%H$ */
-/* commit time: \$Format:\%ai$ */
+/* ref:         $Format:%D$ */
+/* git commit:  $Format:%H$ */
+/* commit time: $Format:%ai$ */
 EOS
    }
    close $src;


dep.pl
```diff
@@ -81,9 +81,9 @@
    if ($apply == 1) {
       print {$out} << 'EOS';
 #endif
-/* ref:         \$Format:\%D$ */
-/* git commit:  \$Format:\%H$ */
-/* commit time: \$Format:\%ai$ */
```
minad: Could we just get rid of these footers? I think these are obsolete in git. With SVN such comments were very popular.

czurnieden: > With SVN such comments were very popular

No, they weren't just popular, they were absolutely needed! ;-)

But you are right, they are not really needed anymore with git. Hands up who is still using them?
\*puts hands in pockets\*

minad: I find them useless and distracting. They are also never expanded in my git checkouts. Probably git must be configured somehow, but I never bothered.

minad: I don't even know when these are meant to be expanded. For releases? If they would be changed every time that would be some unnecessary disk IO.

minad: But then - I wouldn't want release files to have different checksums than the files in git. I have no idea what the tags are for. Maybe I should think more before I hit enter.

minad:
Ok to close?

PR #212, opened 2019-04-07
Better interoperability between MSVC and mingw-w64

nijtmans:
The decision whether to compile with MP_64BIT or MP_32BIT mode should not be dependent on the compiler but on the platform. For example, it should be possible to compile libtommath as a dll, using mingw-w64, but then using that dll in a MSVC application. Currently mingw-w64 uses MP_64BIT as default, but MSVC only operates in MP_32BIT. So they cannot co-operate.

Therefore I propose to make MP_32BIT the default on mingw-w64 too, but make it overridable setting -DMP_64BIT explicitly. This way, applications or dll's compiled with MSVC or mingw-w64 can always operate together, no matter which module/application was compiled with which compiler.

The same change (I suppose) should be made in libtomcrypt as well.

### DIFF
--- a/tommath.h
+++ b/tommath.h
@@ -27,7 +27,7 @@ extern "C" {
 #endif
 
 /* MS Visual C++ doesn't have a 128bit type for words, so fall back to 32bit MPI's (where words are 64bit) */
-#if defined(_MSC_VER) || defined(__LLP64__) || defined(__e2k__) || defined(__LCC__)
+#if (defined(_WIN32) || defined(__LLP64__) || defined(__e2k__) || defined(__LCC__)) && !defined(MP_64BIT)
 #   define MP_32BIT
 #endif
 
@@ -38,7 +38,7 @@ extern "C" {
     defined(__sparcv9) || defined(__sparc_v9__) || defined(__sparc64__) || \
     defined(__ia64) || defined(__ia64__) || defined(__itanium__) || defined(_M_IA64) || \
     defined(__LP64__) || defined(_LP64) || defined(__64BIT__)
-#   if !(defined(MP_32BIT) || defined(MP_16BIT) || defined(MP_8BIT))
+#   if !(defined(MP_64BIT) || defined(MP_32BIT) || defined(MP_16BIT) || defined(MP_8BIT))
 #      if defined(__GNUC__)
 /* we support 128bit integers only via: __attribute__((mode(TI))) */
 #         define MP_64BIT


minad:
I would prefer to have MP_64BIT on _WIN64 platforms. Are you proposing to use MP_32BIT always on windows?

nijtmans:
The most dominant compiler is Visual Studio, and MP_64BIT doesn't work there.  So - unfortunately - mingw-w64 should conform to the 'standard', otherwise dll's generated by mingw-w64 are not compatible with dll's generated with Visual Studio.  That's the problem.

nijtmans:
I want MP_32BIT to be the default, but people who want MP_64BIT (thereby forced to use mingw-w64) should have that possible too.

minad:
So if mingw64 is detected, enable MP_64BIT and in visual studio MP_32BIT? Why would people compile stuff with mingw64 and some other stuff with msvc?

nijtmans:
> So if mingw64 is detected, enable MP_64BIT and in visual studio MP_32BIT? Why would people compile stuff with mingw64 and some other stuff with msvc?

Well, I do!  In a future version of Tcl (9.0?)  I would like to provide libtommath as a dll, but I don't want to force all Tcl extension writers using libtommath functions to abandon MSVC and switch to mingw-w64. Personally I prefer mingw-w64 too, but I cannot convince the whole world, and I also cannot convince MSVC to implement a 128bit integer type. So, I'm stuck in this desire ... :-(

minad:
>  Well, I do! In a future version of Tcl (9.0?) I would like to provide libtommath as a dll, but I don't want to force all Tcl extension writers using libtommath functions to abandon MSVC and switch to mingw-w64. Personally I prefer mingw-w64 too, but I cannot convince the whole world, and I also cannot convince MSVC to implement a 128bit integer type. So, I'm stuck in this desire ... :-(

But then I think choosing the non ideal setting for msvc compatibility is the wrong idea. I would prefer if tommath chooses the ideal settings for the given platform and compiler - and for mingw-w64/_WIN64 this is MP_64BIT. Is it not possible to provide a special compatibility dll for your use case in tcl? 

Furthermore I wonder if tommath MP_64BIT is binary compatible with MP_32BIT as long as the mp_digit internals are not accessed?

nijtmans:
> Furthermore I wonder if tommath MP_64BIT is binary compatible with MP_32BIT as long as the mp_digit internals are not accessed?

Yes, that's true. But - currently - Tcl accesses mp_digit's internals, and it will be a lot of work to fix this (I already fixed it in some places).  So, at this moment, Tcl _MUST_ be compiled with the same compiler as the libtommath dll. Anyway, I presented my arguments, it's up to @sjaeckel to decide (I assume).  That's all I can do.

karel-m:
I think it would be unfotunate to downgrade perfomance of default
mingw/gcc/64bit build just because MSVC doesn't support 128bit wide int
type.

Would it help to change the default dll names to something like
libtommath_mingw64.dll?

minad:
I agree with @karel-m. Maybe provide two different builds libtommath_mingw64.dll and libtommath_msvc.dll where the msvc dll could still be compiled with mingw-w64, but in such a way that it's compatible with msvc. tommath_mingw64.dll should have optimal performance for pure mingw-w64 builds and this should be the default.
Furthermore I don't see why a project like tommath should compromise in favor of an unfree compiler.

czurnieden:
@minad 
> Furthermore I don't see why a project like tommath should compromise in favor of an unfree compiler.

As long as it doesn't cost too much it is ok. And the patch here costs exactly: nothing.
Assuming the documentation gets updated, too, of course! (\*hint hint\* ;-) )

Also: you cannot ignore MSVC and leave millions out there in the cold who cannot install a better compiler. C'mon, show some heart!

minad:
>  As long as it doesn't cost too much it is ok. ...
> Assuming the documentation gets updated, too, of course! (*hint hint* ;-) )

Agree.

> And the patch here costs exactly: nothing.

Not true. This patch puts mingw-w64 users at a disadvantage in favor of msvc-mingw compatibility.

czurnieden:
>> And the patch here costs exactly: nothing.

> Not true.

But for us developers here at LTM? Nothing.

> This patch puts mingw-w64 users at a disadvantage in favor of msvc-mingw compatibility.

That's why it is called a compromise. Everybody hates it but everybody can live with it, too.

Put it into the documentation, maybe `#warning` the Ming-64 users and that's it. Don't try to please everybody, it never works out.
Believe me, I tried it a lot ;-)

Oh, and it is Windows for Kernighan&Ritchie's sake! Its users are well-adjusted to a moderate-to-harsh degree of discomfort and this is nothing more than setting a macro if they want to use a non-standard (in this case MingW ) compiler, not a life-threatening registry surgery.

minad:
@czurniden For me as a user this patch is not good. It basically downgrades mingw to the level of msvc.

If compatibility is the priority, the tommath dll should be compiled with msvc. Then this patch is unnecessary.

karel-m:
> Oh, and it is Windows for Kernighan&Ritchie's sake! Its users are well-adjusted to a moderate-to-harsh degree of discomfort and this is nothing more than setting a macro if they want to use a non-standard (in this case MingW ) compiler, not a life-threatening registry surgery.

Why do we want to support the scenario when a user of "standard compiler" wants to use a DLL built with "non-standard compiler"? The penalty for all "non-standard compiler" users is IMO too high.

As already mentioned I suggest patching `makefile.mingw` like:
```diff
- LIBMAIN_D =libtommath.dll
+ LIBMAIN_D =libtommath_gcc.dll
```
Which should avoid problems caused by copying a random `libtommath.dll` to somewhere and expect it will work with other MSVC compiled binaries.

If the original intention of this PR was: "to be able to build libtommath.dll with gcc so that it is compatible with MSVC-built binaries" we can try to add this option but not by downgrading the performance of default mingw/gcc build.

karel-m:
> If the original intention of this PR was: "to be able to build libtommath.dll with gcc so that it is compatible with MSVC-built binaries" we can try to add this option but not by downgrading the performance of default mingw/gcc build.

In fact we already do support it:
```
gmake -f makefile.mingw CFLAGS="-O2 -DMP_32BIT" libtommath.dll
```

nijtmans:
Don't worry, there's another solution building a "libtommath.dll" which is usable by both MSVC and mingw-w64, and also built with MP_64BIT!    Stay tuned. I'll submit a new PR for that soon.

PR #213, opened 2019-04-08
do not expose mp_balance_mul, add MP_ENABLED feature detection

minad:
ping @czurnieden @sjaeckel

### DIFF
--- a/bn_mp_mul.c
+++ b/bn_mp_mul.c
@@ -6,82 +6,65 @@
 /* high level multiplication (handles sign) */
 int mp_mul(const mp_int *a, const mp_int *b, mp_int *c)
 {
-   int     res, neg;
-#ifdef BN_MP_BALANCE_MUL_C
-   int len_b, len_a;
-#endif
+   int res, neg, min_len, max_len, digs;
+   min_len = MIN(a->used, b->used);
+   max_len = MAX(a->used, b->used);
+   digs = a->used + b->used + 1;
    neg = (a->sign == b->sign) ? MP_ZPOS : MP_NEG;
-#ifdef BN_MP_BALANCE_MUL_C
-   len_a = a->used;
-   len_b = b->used;
 
-   if (len_a == len_b) {
-      goto GO_ON;
+   if (0) {
    }
-   /*
-    * Check sizes. The smaller one needs to be larger than the Karatsuba cut-off.
+
+#ifdef BN_MP_BALANCE_MUL_C
+   /* Check sizes. The smaller one needs to be larger than the Karatsuba cut-off.
     * The bigger one needs to be at least about one KARATSUBA_MUL_CUTOFF bigger
     * to make some sense, but it depends on architecture, OS, position of the
     * stars... so YMMV.
     * Using it to cut the input into slices small enough for fast_s_mp_mul_digs
     * was actually slower on the author's machine, but YMMV.
     */
-   if ((MIN(len_a, len_b) < KARATSUBA_MUL_CUTOFF)
-       || ((MAX(len_a, len_b)) / 2 < KARATSUBA_MUL_CUTOFF)) {
-      goto GO_ON;
+   else if ((min_len >= KARATSUBA_MUL_CUTOFF) &&
+            (max_len / 2 >= KARATSUBA_MUL_CUTOFF) &&
+            (max_len >= (2 * min_len))) { /* Not much effect was observed below a ratio of 1:2, but again: YMMV. */
+      res = mp_balance_mul(a,b,c);
    }
-   /*
-    * Not much effect was observed below a ratio of 1:2, but again: YMMV.
-    */
-   if ((MAX(len_a, len_b) /  MIN(len_a, len_b)) < 2) {
-      goto GO_ON;
-   }
-
-   res = mp_balance_mul(a,b,c);
-   goto END;
-
-GO_ON:
 #endif
 
-   /* use Toom-Cook? */
 #ifdef BN_MP_TOOM_MUL_C
-   if (MIN(a->used, b->used) >= TOOM_MUL_CUTOFF) {
+   else if (min_len >= TOOM_MUL_CUTOFF) {
       res = mp_toom_mul(a, b, c);
-   } else
+   }
 #endif
+
 #ifdef BN_MP_KARATSUBA_MUL_C
-      /* use Karatsuba? */
-      if (MIN(a->used, b->used) >= KARATSUBA_MUL_CUTOFF) {
-         res = mp_karatsuba_mul(a, b, c);
-      } else
+   else if (min_len >= KARATSUBA_MUL_CUTOFF) {
+      res = mp_karatsuba_mul(a, b, c);
+   }
 #endif
-      {
-         /* can we use the fast multiplier?
-          *
-          * The fast multiplier can be used if the output will
-          * have less than MP_WARRAY digits and the number of
-          * digits won't affect carry propagation
-          */
-         int     digs = a->used + b->used + 1;
 
 #ifdef BN_FAST_S_MP_MUL_DIGS_C
-         if ((digs < (int)MP_WARRAY) &&
-             (MIN(a->used, b->used) <=
-              (int)(1u << ((CHAR_BIT * sizeof(mp_word)) - (2u * (size_t)DIGIT_BIT))))) {
-            res = fast_s_mp_mul_digs(a, b, c, digs);
-         } else
+   /* can we use the fast multiplier?
+    *
+    * The fast multiplier can be used if the output will
+    * have less than MP_WARRAY digits and the number of
+    * digits won't affect carry propagation
+    */
+   else if ((digs < (int)MP_WARRAY) &&
+            (min_len <=
+             (int)(1u << ((CHAR_BIT * sizeof(mp_word)) - (2u * (size_t)DIGIT_BIT))))) {
+      res = fast_s_mp_mul_digs(a, b, c, digs);
+   }
 #endif
-         {
+
+   else {
 #ifdef BN_S_MP_MUL_DIGS_C
-            res = s_mp_mul(a, b, c); /* uses s_mp_mul_digs */
+      res = s_mp_mul(a, b, c); /* uses s_mp_mul_digs */
 #else
-            res = MP_VAL;
+      res = MP_VAL;
 #endif
-         }
-      }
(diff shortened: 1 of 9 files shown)


demo/test.c
```diff
@@ -1,5 +1,8 @@
 #include "shared.h"
 
+/* We can also test the private API here */
+#include "tommath_private.h"
```
sjaeckel: we can but should we?

IMO as soon as the library symbols are properly set-up you won't be able to build these tests then against the shared library

minad: Then we need a special test build with exposed s_* symbols. I think it is better if internal functions are tested separately. For example in the mul case we could also test the different code paths by passing the right values. But this is too fragile and I would also like to test edge cases.

This would also solve the issue with exposed CUTOFF variables. We should just make those constant and test karatsuba and toom cook separately.

sjaeckel: > We should just make those constant and test karatsuba and toom cook separately

IIUC they can't be made constant as they vary from platform to platform. Sure they're mostly constant now as nobody's really tweaking them but I guess that's caused by the fact that there's no proper way yet to determine them per platform. (besides running `etc/tune`)

> Then we need a special test build with exposed s_* symbols. I think it is better if internal functions are tested separately.

I'm no big fan of explicit testing of internal functionality.

> For example in the mul case we could also test the different code paths by passing the right values.

I prefer that way to make sure the entire chain works and then determine that everything is called through code coverage.

> But this is too fragile

why is this too fragile?

> I would also like to test edge cases

why would you want to test cases which can't be hit?

that all sounds a bit over the top to me

minad: > I'm no big fan of explicit testing of internal functionality.

I am not saying to test arbitrary internal things. However internal functions with a well defined API should be tested, e.g., mp_balance_mul. @czurnieden also added the balance mul test to the test suite, but he did not call it directly which he should have. If only the external api should be tested, there should not even be a function test_balance_mul.

> I prefer that way to make sure the entire chain works and then determine that everything is called through code coverage.

If you want that - then the test suite must be split such that internal function tests and api function tests can be compiled separately. No big deal. `#if TEST_INTERNAL .... #endif`.
Ensure that stuff is hit via code coverage is good practice, but not enough and it depends on external tooling. Do we have coverage set up?

Besides, if we would just start renaming internal functions to `s_*`, we wouldn't have to tweak the visibility settings in the first step. The `s_*` functions would just be internal and not belong to the public ABI by our own convention. In the next step we could make the symbols hidden.

> why is this too fragile?

The test of the internal function mp_balance_mul should not depend on the criterion at which this function is selected in mp_mul.

>  why would you want to test cases which can't be hit?

Robustness of the functions. I would not extend functions to accept values which cannot be handled fundamentally, but for example if karatsuba also works for small numbers and these should be tested too.

> that all sounds a bit over the top to me

No it is not. If you want to see something which is over the top look at the sqlite test suite. What I am proposing here is good practice and modest. It is also too much work hopefully. Furthermore there are people willing to improve stuff, like me getting this test suite refactoring started.

minad:
> TBH I preferred the way with separate if-statements with comments instead of one big if-statement with in-line comments

Hmm written like this it is more aligned with how the conditions are checked for the other special cases karatsuba and toom. I think it is was more clear like this, but astyle makes a bit of a mess :(

minad:
@sjaeckel Better like this?

bn_mp_mul.c
```diff
@@ -6,82 +6,69 @@
...
-   if ((MIN(len_a, len_b) < KARATSUBA_MUL_CUTOFF)
-       || ((MAX(len_a, len_b)) / 2 < KARATSUBA_MUL_CUTOFF)) {
-      goto GO_ON;
-   }
-   /*
-    * Not much effect was observed below a ratio of 1:2, but again: YMMV.
-    */
-   if ((MAX(len_a, len_b) /  MIN(len_a, len_b)) < 2) {
-      goto GO_ON;
+   if (0) {
```
sjaeckel: srsly?

bn_mp_mul.c
```diff
@@ -6,82 +6,69 @@
...
    }
 
-   res = mp_balance_mul(a,b,c);
-   goto END;
+#ifdef BN_MP_BALANCE_MUL_C
+   else if ((a->used != b->used) &&
+
+            /* Check sizes. The smaller one needs to be larger than the Karatsuba cut-off.
```
sjaeckel: it's about the entire readability and such complex if statements suck

demo/test.c
sjaeckel: > If only the external api should be tested, there should not even be a function test_balance_mul.

why not if this test is specifically crafted to test `mp_balance_mul` within `mp_mul`? 

> Do we have coverage set up?

you can do `make coverage` :)

> Besides, if we would just start renaming internal functions to `s_*`, we wouldn't have to tweak the visibility settings in the first step. The `s_*` functions would just be internal and not belong to the public ABI by our own convention. In the next step we could make the symbols hidden.

fine by me, keep the discussion points of #172 in mind

> The test of the internal function mp_balance_mul should not depend on the criterion at which this function is selected in mp_mul.

what sense does a specific test even make if its counterpart isn't available? as it is now the test's can't even be built w/o `mp_balance_mul`

> if karatsuba also works for small numbers and these should be tested too.

then set the cutoff in the testcase appropriately?

bn_mp_mul.c
minad: Please compare the old code with the new code side by side, don't look at the diff only.
Then I think this PR improves things.

sjaeckel: which old code? pre-206-merged?

minad: Compare
https://github.com/libtom/libtommath/blob/develop/bn_mp_mul.c
with
https://github.com/libtom/libtommath/blob/41d3d73b8e19531fd476cc57355bdb1a015c5eb3/bn_mp_mul.c
and then tell me seriously if you defend the goto version.

sjaeckel: Btw. I've the impression that this is going in the same direction as the rebase discussion.

Linearity of things, where linearity in my opinion leads to easier readability for a human.

minad: I only see superficial similarities besides you disagreeing with me on basic things. And btw goto can be much less linear while if/else etc is linear by construction.

sjaeckel: > besides you disagreeing with me on basic things

please explain

sjaeckel: >  and then tell me seriously if you defend the goto version.

BTW yes, overall it improved

minad: For example we disagreed about what should be tested or not, this is quite basic. Then there was the discussion about the rebasing, then the discussion about about the public ABI.

I would prefer if we wouldn't over discuss too many things to be honest. If there is something to be learned ok but otherwise it just costs time. It happened a few times now - I created a (from my pov) quick and uncontroversial patch and then a longish discussion emerges.  But I am not saying I am innocent in this respect ;)

minad: The length of the patch should be proportional to the length of the discussion :)

minad: > BTW yes, overall it improved

good to hear :sweat_smile:

sjaeckel: > The length of the patch should be proportional to the length of the discussion :)

True, if there are no fundamentals being discussed as e.g. the testing.

> then the discussion about about the public ABI

You mean #172 ?

> I created a (from my pov) quick and uncontroversial patch

I was mostly fine with the original patch, besides the points I commented on. Then you made a v2 which implemented `mp_mul` entirely different and also changed parts that weren't commented on... yes this leads to re-discussion of things and new discussions...

minad:
>     * I don't think `(a->used != b->used)` at line 18 will save enough to be worth the ops.

Fixed.

>     * And either use a variable for both MIN/MAX or for none. Please don't mix them. Many OCDers will thank you for that.

Ok, ok :)

> I can live with all of the above, no problem, with one exception: `if (0) {`

I will not change this one. I like it since it looks like a nice trick.

minad:
Btw, we could introduce an `ENABLED` macro to handle the ifdefs then we could use something much nicer like. However we would then rely on the compilers dead code elimination. I usually prefer that style. If you want I prepare a patch, but this will change quite a few lines all over tommath I guess.

~~~
if (ENABLED(...) && condition) {
    ....
}
~~~

czurnieden:
> . I like it since it looks like a nice trick.

I don't like it since it looks like an abomination.
An alien element.
A leftover you would automatically delete and then wonder why the compiler complains.

Please: either use a goto, or, since there is nothing to cleanup, set the sign and return from the balance branch directly.

Oh, and also: there is more to come. The TC4,5(6) can go below the balancing, no special treatment needed but FFT/NFT need a different kind of massage, must be treated before the balancing and have an immediate out, too. FFT/NFT will need some time (FFT is ready to go but NFT is not even started) but it would be nice if you don't make it too complicated to add them.

bn_mp_mul.c
```diff
@@ -3,6 +3,13 @@
 /* LibTomMath, multiple-precision integer library -- Tom St Denis */
 /* SPDX-License-Identifier: Unlicense */
 
+#define _MP_ENABLED_TEST        ,
+#define _MP_ENABLED_TEST1       ,
+#define MP_ENABLED(x)           _MP_ENABLED1(x)
+#define _MP_ENABLED1(x)         _MP_ENABLED2(_MP_ENABLED_TEST##x)
+#define _MP_ENABLED2(x)         _MP_ENABLED3(x 1, 0)
+#define _MP_ENABLED3(x, y, ...) y
```
minad: @czurnieden @sjaeckel How do you like something like this.

The macro will evaluate to 1 if the macro is either defined as empty or defined as 1.

czurnieden: The original macros were meant to exclude code from compiling, from going into the binary in the first place. This patch keeps the branches in and hopes for the compiler to optimize them away.
Is it the case? Have you checked the dumps of GCC and clang? What is left if compiled with `-Os`?

minad: This will always work. Every compiler optimizes `if (0)` away. For example the Linux kernel does that too. It is actually a quite common practice. The advantage of this is that you always get the typechecking.

minad: The only situation where you have to go back to #ifdefs is if the definition is missing or changes depending on macros.

minad:
superseded by #214

PR #216, opened 2019-04-09
Better interoperability between MSVC and mingw-w64, 2

nijtmans:
Here is another attempt, which succeeded to produce a single "libtommath.dll" (actually 2: win32 and win64) which is usable for both MSVC and mingw-w64.   The dll can be built normally, only an addtional compiler option prevents that it depends on a mingw-w64 specific dll.

I built Tcl with MSVC (both win32 and win64) and with mingw-w64 (also win32 and win64), but replacing the built-in libtommath with "libtommath.dll". It worked. The MSVC import library "tommath.lib" must be generated using a def-file (tommath.def), which is also committed. Ideally, this .def-file should be automatically generated (as is done with the makefiles), but I'm not confident enough in perl to implement that.

Advantage of this approach: MSVC-built application can benefit from DIGIT_BITS=60, even though the compiler cannot build the dll. But it can import the mingw-w64-generated dll, and fully interoperate with it.


sjaeckel:
how's this def file created?

nijtmans:
> how's this def file created?
Manually, just by building first and then using nm to make a list of the exported symbols.  It just contains a list of exported symbols in alphabetical order (not really necessary, actually).  Should be generated by a script, but I'm not confident in perl doing that.

sjaeckel:
> Manually, just by building first and then using nm to make a list of the exported symbols. It just contains a list of exported symbols in alphabetical order (not really necessary, actually). Should be generated by a script, but I'm not confident in perl doing that.

no need to do it in perl, I'd use the same tools as @minad did in `rename.sh` in #172 as I'm also not confident in perl :) 

I think this script needs to do more in the future, but what you described is enough for now.

In the future (2.0.0) it should build the library, then check vs. what is in `tommath.h` (and don't care which functions come out of `tommath_private.h`) and only add those of `tommath.h` to the list.

sjaeckel:
can you please rebase on top of develop and squash all those commits?

nijtmans:
> can you please rebase on top of develop and squash all those commits?

Done @sjaeckel

sjaeckel:
Thx!

@karel-m can you please have a look at these changes, fine by you as well?

minad:
> MSVC-built  application can benefit from DIGIT_BITS=60, even though the compiler  cannot build the dll. But it can import the mingw-w64-generated dll, and  fully interoperate with it.

@nijtmans I had expected that something like this might work. However you mentioned that tcl access the mp_int structs. How will that work?

minad:
Ah, and why do you need a def file? Just asking, I don't know what they are for. Some kind of additional infos for the linker?

nijtmans:
> > MSVC-built  application can benefit from DIGIT_BITS=60, even though the compiler  cannot build the dll. But it can import the mingw-w64-generated dll, and  fully interoperate with it.
> 
> @nijtmans I had expected that something like this might work. However you mentioned that tcl accesses the mp_int structs. How will that work?

Well, it works: [See here](https://travis-ci.org/tcltk/tcl/builds/518632157).   As long as Tcl is compiled with the same MP-mode as the libtommath library, it works. (There were a few fixes needed in the Tcl source-code, but those are done now)

nijtmans:
> Ah, and why do you need a def file? Just asking, I don't know what they are for. Some kind of additional infos for the linker?

A def-file is needed for MSVC-builds, the import-libraries can be generated from it.

tommath.h
```diff
@@ -19,7 +19,7 @@ extern "C" {
 #endif
 
 /* MS Visual C++ doesn't have a 128bit type for words, so fall back to 32bit MPI's (where words are 64bit) */
-#if defined(_MSC_VER) || defined(__LLP64__) || defined(__e2k__) || defined(__LCC__)
+#if (defined(_MSC_VER) || defined(__LLP64__) || defined(__e2k__) || defined(__LCC__)) && !defined(MP_64BIT)
```
minad: I wonder if this whole discussion also in #212 missed the point. Isn't __LLP64__ also defined in mingw?

nijtmans: > I wonder if this whole discussion also in #212 missed the point. Isn't **LLP64** also defined in mingw?

Maybe, but that doesn't matter for this PR:   This PR makes the default overridable, so by defining -DMP_64BIT we can override the default on any platform.

tommath.h
```diff
@@ -66,7 +66,9 @@ typedef uint32_t             mp_word;
 #elif defined(MP_64BIT)
 /* for GCC only on supported platforms */
 typedef uint64_t mp_digit;
+#if defined(__GNUC__)
 typedef unsigned long        mp_word __attribute__((mode(TI)));
+#endif
```
minad: What happens here? Will mp_word just be undefined if MP_UINT64 is defined and we are not using __GNUC__?

nijtmans: Yes, that's it. Using tommath.h with MSVC in MP_64BIT mode makes the mp_word type unusable. But that doesn't matter since mp_word is not used in any public API: It works!

nijtmans:
I'll rebase this one after #267 is merged to trunk.

tommath.h
nijtmans: See #267, I think that explains it all

generate_def.sh
```diff
@@ -0,0 +1,9 @@
+echo "; libtommath" >tommath.def
```
minad: Could we please add this to the existing scripts in order to ensure that things get executed automatically and don't get out of sync?

minad: I merged a few scripts together and helper.pl is now the place where all generation tasks should be done. See #293

nijtmans: Thanks!  I'll try to have a look this weekend

sjaeckel: FYI I've merged #293 so you should be able to finally do the rebase when you have the time!

nijtmans:
Well, I tried to rebase, but somehow that failed horribily

sjaeckel:
> Well, I tried to rebase, but somehow that failed horribily

hmm I tried the rebase of the old HEAD @ fbc0829405ae5bf9e5aed4d16cf40d052d0eb1e9 and there was a minor merge conflict, but otherwise it went smoothly

now that we have `bn_deprecated.c` et.al. your old approach of searching for the files etc also won't work anymore...

sjaeckel:
> hmm I tried the rebase of the old HEAD @ [fbc0829](https://github.com/libtom/libtommath/commit/fbc0829405ae5bf9e5aed4d16cf40d052d0eb1e9) and there was a minor merge conflict, but otherwise it went smoothly

I've pushed the rebased branch to [rebased/interop-mingw-w64-2](https://github.com/libtom/libtommath/tree/rebased/interop-mingw-w64-2)

nijtmans:
Closing in favour of #300

PR #217, opened 2019-04-10
Rename bn_error.c to bn_mp_error_to_string.c

nijtmans:
Title says it all.  Most source file-name are named after the function they implement. This one was an exception, streightforward to fix.


nijtmans:
> Any plans to deprecate the old name for backward compatibility?

The function name is unchanged, only the name of the source-file implementing this function ;-)

sjaeckel:
@buggywhip feel free to comment if you meant the filename! any particular reason you think this would be required?

PR #219, opened 2019-04-10
Resolving problems mentioned in #218

czurnieden:
Added forgotten checks for `mp_set_int` after type-change as found by @fperrad . I hope I included all additional changes from #218 , too. Please check.


minad:
@czurnieden @fperrad The error checking will be unnecessary as soon as #221 is merged, since these calls are guaranteed to never fail. After #221 the error checking for all *set_int* and *set_long_long* calls should be removed.

For now this PR looks ok however.

PR #220, opened 2019-04-12
Add handling of MP_ITER error-code to mp_error_to_string()

nijtmans:
mp_error_to_string() doesn't handle MP_ITER currently, I think it should.

### DIFF
--- a/bn_mp_error_to_string.c
+++ b/bn_mp_error_to_string.c
@@ -3,30 +3,21 @@
 /* LibTomMath, multiple-precision integer library -- Tom St Denis */
 /* SPDX-License-Identifier: Unlicense */
 
-static const struct {
-   int code;
-   const char *msg;
-} msgs[] = {
-   { MP_OKAY, "Successful" },
-   { MP_MEM,  "Out of heap" },
-   { MP_VAL,  "Value out of range" },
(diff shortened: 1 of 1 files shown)


minad:
While you are on it - what about replacing the function with a simple switch statement. With modern compilers you can enable coverage checking then.

sjaeckel:
> **Edit**: Ok we are not using enums :(

but we should

minad:
No, not here, it would also give some kind of compatibility issues. This time in the public API, about which I am much more hesitant. My **should** was more meant as - this is how it should be done in a modern c style.

nijtmans:
I agree with the comments:  It should have been an enum, but that can cause compatibility issues in the public API, so doing that now is a bad idea!

minad:
Haha, @sjaeckel you wrote **we should**. I only thought it :)

czurnieden:
> mp_error_to_string() doesn't handle MP_ITER currently, I think it should.

Yepp, forgot it, my bad.
But at least it is documented in bn.tex ;-)

And it is not really an iteration that doesn't converge in a timely manner but the list of available candidates for the coefficient `a` ended (significant only with MP_8BIT, although the probability is sufficiently low), so `MP_ITER` is another proof that I am really bad at naming things.

PR #221, opened 2019-04-12
No grow necessary in mp_set_int* functions

nijtmans:
In the mp_set_long_long (and related functions), an mp_grow() call is done, but that's only necessary if the already allocated room (MP_PREC) is so small that not even 64 bits fit. There is only one such a situation: -DMP_LOW_MEM in combination with -DMP_8BIT.

Therefore, I propose to increase the minimum MP_PREC from 8 to 16 in this case (allocating 8 bytes is ridicilously small anyway), then the mp_grow() is no longer necessesary in any mode.

This means that the minimum number of _bytes_ allocated are always minimum 16, which is always sufficient to store the 64 bits of a long long.

### DIFF
--- a/bn_mp_grow.c
+++ b/bn_mp_grow.c
@@ -11,8 +11,8 @@ int mp_grow(mp_int *a, int size)
 
    /* if the alloc size is smaller alloc more ram */
    if (a->alloc < size) {
-      /* ensure there are always at least MP_PREC digits extra on top */
-      size += ((MP_PREC * 2) - 1) - ((size + (MP_PREC - 1)) % MP_PREC);
+      /* round up to a multiple of MP_PREC */
+      size += (MP_PREC - 1) - ((size + (MP_PREC - 1)) % MP_PREC);
 
       /* reallocate the array a->dp
        *
--- a/bn_mp_grow.c
+++ b/bn_mp_grow.c
@@ -12,7 +12,7 @@ int mp_grow(mp_int *a, int size)
    /* if the alloc size is smaller alloc more ram */
    if (a->alloc < size) {
       /* ensure there are always at least MP_PREC digits extra on top */
-      size += (MP_PREC * 2) - (size % MP_PREC);
+      size += ((MP_PREC * 2) - 1) - ((size + (MP_PREC - 1)) % MP_PREC);
 
       /* reallocate the array a->dp
        *
--- a/bn_mp_shrink.c
+++ b/bn_mp_shrink.c
@@ -7,9 +7,9 @@
 int mp_shrink(mp_int *a)
 {
    mp_digit *tmp;
-   int alloc = MP_PREC;
+   int alloc = MP_MIN_PREC;
 
-   if (a->used > MP_PREC) {
+   if (a->used > MP_MIN_PREC) {
       alloc = a->used;
       alloc += (MP_PREC - 1) - ((alloc + (MP_PREC - 1)) % MP_PREC);
    }
--- a/bn_mp_shrink.c
+++ b/bn_mp_shrink.c
@@ -7,20 +7,21 @@
 int mp_shrink(mp_int *a)
 {
    mp_digit *tmp;
-   int used = 1;
+   int alloc = MP_PREC;
 
-   if (a->used > 0) {
-      used = a->used;
+   if (a->used > MP_PREC) {
+      alloc = a->used;
+      alloc += (MP_PREC - 1) - ((alloc + (MP_PREC - 1)) % MP_PREC);
    }
 
-   if (a->alloc != used) {
+   if (a->alloc != alloc) {
       if ((tmp = (mp_digit *) MP_REALLOC(a->dp,
                                          (size_t)a->alloc * sizeof(mp_digit),
-                                         (size_t)used * sizeof(mp_digit))) == NULL) {
+                                         (size_t)alloc * sizeof(mp_digit))) == NULL) {
          return MP_MEM;
       }
       a->dp    = tmp;
-      a->alloc = used;
+      a->alloc = alloc;
    }
    return MP_OKAY;
 }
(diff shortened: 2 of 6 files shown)


minad:
Most allocators also don't return less than 16 bytes.

sjaeckel:
thou shalt then never call `mp_shrink` before that :)

minad:
mp_shrink should never shrink below this minimum precision? This should also be included here.

sjaeckel:
I'd say it should be enough to never shrink below the required storage for a `long long` then! otherwise the target of `mp_shrink isn't fulfilled anymore

sjaeckel:
>  @sjaeckel Can it happen that numbers are completely empty or are they always allocated?

they're always allocated

nijtmans:
Yeah, I found mp_shrink() as well ;-(

So mp_shrink() should be modified to always shrink to at least MP_PREC, and also always a multiple of that.

And .. mp_init_size/mp_grow lied: The comment said they always allocote MP_PREC digits extra, but in fact they always allocate MP_PREC+1 digits extra.  That's too much. Fixed that as well ....

sjaeckel:
> So mp_shrink() should be modified to always shrink to at least MP_PREC, and also always a multiple of that.

> I'd say it should be enough to never shrink below the required storage for a `long long` then

not MP_PREC

nijtmans:
Hoping it's better now!

sjaeckel:
can you please fix the commit message?

bn_mp_shrink.c
```diff
@@ -7,20 +7,21 @@
 int mp_shrink(mp_int *a)
 {
    mp_digit *tmp;
-   int used = 1;
+   int alloc = MP_PREC;
```
sjaeckel: this still shrink's only down to `MP_PREC` and not the required storage for a `long long`

nijtmans: I think the requirement of MP_PREC should be that it can store at least a long long.   That's what the change in tommath.h was meant for!   mp_init() also starts with allocating MP_PREC digits.

nijtmans: So, for MP_8BIT, the minimum MP_PREC=16, for MP_16BIT, the minimum MP_PREC=8, for MP_32BIT the minimum MP_PREC=4, for MP_64BIT it is MP_PREC=2.   That should be documented somewhere.

sjaeckel: @czurnieden what do you say?

minad: What about adding a MP_MIN_PREC setting to the minimums given by @nijtmans and changing MP_PREC such that it is always larger than MP_MIN_PREC? 
Furthermore I think MP_PREC is usually too large. But this depends heavly on the use case. Via static_assert it could be checked that the condition always holds if MP_PREC is set by the user.

nijtmans: How about an assert somewhere in the code checking that MP_PREC * sizeof(mp_digit) >= sizeof(long long)   ???

nijtmans: Hm...   I meant MP_PREC * DIGIT_BIT >= sizeof(long long).    That means that MP_PREC for MP_8BIT could even be as low as 10, although that's not a power of 2.

nijtmans: Hm  ....    MP_PREC * DIGIT_BIT >= CHAR_BIT * sizeof(long long)

sjaeckel: > How about an assert somewhere in the code checking that MP_PREC * sizeof(mp_digit) >= sizeof(long long) ???

no we need to make it sure at compile time

I'd make it `#define MP_MIN_PREC ((CHAR_BIT * sizeof(long long))/DIGIT_BIT)` and hope that nobody uses this macro in a `#if` preprocessor statement :)

and please keep the patch to to tommath.h regarding MP_8BIT

czurnieden: > czurnieden what do you say?

A `MP_PREC` that is always of the same bitsize no matter what the `MP_xBIT`is would add nicely to the "rounding edges" part of the LTM update, yes.

bn_mp_grow.c
```diff
@@ -12,7 +12,7 @@ int mp_grow(mp_int *a, int size)
    /* if the alloc size is smaller alloc more ram */
    if (a->alloc < size) {
       /* ensure there are always at least MP_PREC digits extra on top */
-      size += (MP_PREC * 2) - (size % MP_PREC);
+      size += ((MP_PREC * 2) - 1) - ((size + (MP_PREC - 1)) % MP_PREC);
```
czurnieden: Is the "extra" really necessary?
I would prefer that LTM does what I tell it and not something different, as benign as it might be.
That "extra" is also not documented, so if somebody uses that information to reduce the memory they cannot blame LTM if it fails.

minad: I agree with @czurnieden. I would **strongly** advise to not use any extra. This will ensure that our valgrind tests tell if something is going wrong.

nijtmans:
So, here my new attempt. MP_MIN_PREC is a multiple of MP_PREC such that at least a long long fits. So, we can set MP_PREC to 2, or even 1 if desired. All allocations will be a multiple of MP_PREC mp_digit's.

bn_mp_grow.c
nijtmans: Adapted now not to use extra's, but simply round up to a multiple of MP_PREC. Build started, let's see how it goe's

czurnieden:
Yes, looks good!
But be aware that `MP_PREC` can be changed by the user and I think it would be helpful if the size of `MP_PREC` is checked. The preprocessor can do it and utter a `#warning` if e.g.: `MP_PREC < 1`. Typos are easily made and if it is equally easy to check them it should be done.

minad:
@nijtmans @czurnieden What is the status here?

nijtmans:
Well, I'm on vacation, so I can't do anything now. I'll have a look when I'm back

minad:
@nijtmans Ok :+1:

minad:
close this in favor of #253

PR #223, opened 2019-04-12
deprecate some macros in tommath.h

minad:
@sjaeckel Does it make sense to cleanup the public api a bit? This will only give warnings in gcc and probably clang. But I don't see a way to do this more generally.

**Edit**: I moved the MP_DEPRECATED to the public header since we might need it later. Furthermore I added MP_DEPRECATED_PRAGMA which can be used in macro situations.

### DIFF
--- a/tommath.h
+++ b/tommath.h
@@ -165,7 +165,9 @@ typedef int ltm_prime_callback(unsigned char *dst, int len, void *dat);
 #    define MP_DEPRECATED_PRAGMA(s)
 #  endif
 
-#  define DIGIT_BIT   MP_DIGIT_BIT
+/* DIGIT_BIT is used in #ifdefs, therefore we cannot use MP_DEPRECATED_PRAGMA */
+#  define DIGIT_BIT   /* DIGIT_BIT macro is deprecated, use MP_DIGIT_BIT instead */ MP_DIGIT_BIT
+
 #  define USED(m)     (MP_DEPRECATED_PRAGMA("USED macro is deprecated, use z->used instead") (m)->used)
 #  define DIGIT(m, k) (MP_DEPRECATED_PRAGMA("DIGIT macro is deprecated, use z->dp instead") (m)->dp[(k)])
 #  define SIGN(m)     (MP_DEPRECATED_PRAGMA("SIGN macro is deprecated, use z->sign instead") (m)->sign)
--- a/tommath.h
+++ b/tommath.h
@@ -148,10 +148,21 @@ typedef struct  {
 /* callback for mp_prime_random, should fill dst with random bytes and return how many read [upto len] */
 typedef int ltm_prime_callback(unsigned char *dst, int len, void *dat);
 
+#if defined(__GNUC__) && (__GNUC__ * 100 + __GNUC_MINOR__ >= 301)
+#  define MP_DEPRECATED(x) __attribute__((deprecated("replaced by " #x)))
+#  define _MP_DEPRECATED_PRAGMA(s) _Pragma(#s)
+#  define MP_DEPRECATED_PRAGMA(s) _MP_DEPRECATED_PRAGMA(GCC warning s)
+#elif defined(_MSC_VER) && _MSC_VER >= 1500
+#  define MP_DEPRECATED(x) __declspec(deprecated("replaced by " #x))
+#  define MP_DEPRECATED_PRAGMA(s) __pragma(message(s))
+#else
+#  define MP_DEPRECATED
+#  define MP_DEPRECATED_PRAGMA(s)
+#endif
 
-#define USED(m)     ((m)->used)
-#define DIGIT(m, k) ((m)->dp[(k)])
-#define SIGN(m)     ((m)->sign)
+#define USED(m)     (MP_DEPRECATED_PRAGMA("USED macro is deprecated, use z->used instead") (m)->used)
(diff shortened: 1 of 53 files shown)


tommath.h
```diff
@@ -148,10 +148,27 @@ typedef struct  {
...
+#  define MP_DEPRECATED_PRAGMA(s) __pragma(message(s))
+#else
+#  define MP_DEPRECATED
+#  define MP_DEPRECATED_PRAGMA(s)
+#endif
 
-#define USED(m)     ((m)->used)
-#define DIGIT(m, k) ((m)->dp[(k)])
-#define SIGN(m)     ((m)->sign)
+#if defined(__GNUC__) && __GNUC__ >= 3
```
sjaeckel: probably add a define `HAS_MP_DEPRECATED_PRAGMA` and check for that as here you're missing out mscv

minad: you might be looking at the first commit?

minad:
@sjaeckel I squashed the two commits to avoid confusion.

tommath.h
sjaeckel: no, I'm looking at the diff and for `_MSC_VER >= 1500` is `MP_DEPRECATED_PRAGMA` defined but this isn't catched here here.

minad: Sorry I don't understand what do you mean. Please clarify.

minad: ah ups now I got it

minad: Fixed. I should probably go to sleep.

sjaeckel: me too gn8

minad:
@sjaeckel Ready to merge?

minad:
There is only one issue I discovered - if the deprecated macros are used inside an #ifdef, the pragma blows up with a compile error instead of a warning. 

An alternative solution for macro deprecation would be the following:

~~~ c
#ifndef MP_NO_DEPRECATED_MACROS
#  warning deprecated macros enabled, define MP_NO_DEPRECATED_MACROS!
#  define DIGIT_BITS MP_DIGIT_BITS
#endif
~~~

@czurnieden What do you think?

czurnieden:
> the pragma blows up with a compile error instead of a warning

Looked good to me!?

What compiler with what error?

sjaeckel:
> There is only one issue I discovered - if the deprecated macros are used inside an #ifdef, the pragma blows up with a compile error instead of a warning.

ew, that sounds nasty

how about creating a separate header file `tommath_deprecated.h` and include that one in `tommath.h`?

there then use this approach

> ```c
> #ifndef MP_NO_DEPRECATED_MACROS
> #  warning deprecated macros enabled, define MP_NO_DEPRECATED_MACROS!
> #  define DIGIT_BITS MP_DIGIT_BITS
> #endif
> ```

and also protect the header file probably by `#pragma once` for compilers which support it + include guards for the rest!?

if it's in a separate file we can simply throw that file away in 2.0.0

after this is resolved the PR LGTM

minad:
@czurnieden This is a minor thing and a bit of a trade off what library users will see when they use deprecated macros.

Assume a library user does the following:

Alternative 1 (in this PR):
~~~ c
int a = DIGIT_BITS; /* compiles, but gives deprecation warning */
#if DIGIT_BITS < 32 /* compile error, no deprecation warning! */
~~~

Alternative 2 (with MP_NO_DEPRECATED_MACROS):
~~~ c
/* ALWAYS prints warning until user defines MP_NO_DEPRECATED_MACROS */
int a = DIGIT_BITS; /* compiles, no warning */
#if DIGIT_BITS < 32 /* compiles, no warning */
~~~

minad:
@sjaeckel Ok, I try to fix this.

minad:
@sjaeckel See my latest commit introducing the MP_NO_DEPRECATED macro. I didn't add additional headers. Grepping for DEPRECATED will show all the deprecated stuff when we are going to remove it.

sjaeckel:
> The macro is intentionally prefixed by an underscore to emphasise that
> its only for internal usage.

and what do we do with the C standard saying that this means it's UB?

How about calling it `INTERNAL_MP_...` instead?

minad:
> The macro is intentionally prefixed by an underscore to emphasise that its only for internal usage.

Hmm, I don't think there is a problem. There are other places where I used a _ prefix. But I can change this since INTERNAL_ is a bit more clear.

sjaeckel:
> Hmm, I don't think there is a problem. There are other places where I used a _ prefix

It's no real problem AFAICT, just the standard telling us what we're allowed to do and what not...

c.f. https://github.com/libtom/libtomcrypt/issues/448

tommath_private.h
```diff
@@ -33,6 +25,10 @@ extern void *MP_CALLOC(size_t nmemb, size_t size);
 extern void MP_FREE(void *mem, size_t size);
 #endif
 
+/* TODO: Remove PRIVATE_MP_WARRAY as soon as deprecated MP_WARRAY is removed from tommath.h */
```
sjaeckel: :+1:

minad:
@sjaeckel Hmm, I played a bit with the patched PR. I must admit that I don't really like the new solution since it always prints a warning even if nothing deprecated is used. I think this will produce much more annoyance for the users in contrast to Alternative 1, where breakage will only occur if DIGIT_BITS is used in an #ifdef, which is rather improbable.

The other possibility would be to just `#define DIGIT_BITS MP_DIGIT_BITS`, but this would then lead to breakage when DIGIT_BITS is removed at some point.

minad:
closed in favor of #226

PR #225, opened 2019-04-19
Test against multiple specific clang/gcc versions

### LINKED ISSUE #208: Pick a few compiler versions and test against them
Pick a few "supported" compiler versions and test against them on Travis. Right now we are just testing against what is there, whatever that is.

This does not mean that older or other compilers won't work, since tommath is very conservative in that regard. They are just not tested during CI runs.

minad:
Fixes #208

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -112,7 +112,7 @@ matrix:
             - gcc-4.9
 
     # clang for x86-64 architecture (64-bit longs and 64-bit pointers)
-    - env: BUILDOPTIONS='--with-cc=clang-7 --with-m64 --with-valgrind'
+    - env: CONV_WARNINGS=1 BUILDOPTIONS='--with-cc=clang-7 --with-m64 --with-valgrind'
     - env: BUILDOPTIONS='--with-cc=clang-6.0 --with-m64 --with-valgrind'
       addons:
         apt:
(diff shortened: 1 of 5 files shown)


minad:
@sjaeckel I added the Wconversion warnings here - enabled only on recent clang-7 which does not give unneessary false positives.

.travis.yml
```diff
@@ -31,15 +31,15 @@ branches:
...
-# Additional installs are Valgrind for the memory-tests
-# and gcc-multilib for the compilation of the different
-# architectures.
+# Additional installs: Valgrind for memory tests.
 install:
   - sudo apt-get update -qq
   - sudo apt-get install valgrind
-  - sudo apt-get install gcc-multilib
+  - apt-cache search gcc | grep '^gcc-[0-9\.]* '
+  - apt-cache search clang | grep compiler
```
czurnieden: I would install them individually instead. The newer ones won't be a problem but the older ones, like e.g.: gcc-4.7 might. Or set the distribution which would be the safest variant, I think.

minad:
> More castings than in a trauma centre in Aspen/Colorado in high season but otherwise&hellip

But only in the tests suite :)

.travis.yml
minad: The compileds are installed separately. The distro is fixed. These lines just print the available compilers in the package database.

PR #226, opened 2019-04-25
Deprecate macros (2)

minad:
Compare with #223. I think this version is more reasonable.


minad:
@sjaeckel ok to merge this?

PR #229, opened 2019-05-03
Allow setting tuning parameters via -D

nomeata:
Depending on the environment, it might be nicer to be able to set the tuning parameters simply via `-D` than via a defining them in another file that is linked into the program.

### DIFF
--- a/tommath.h
+++ b/tommath.h
@@ -118,10 +118,18 @@ typedef uint64_t             mp_word;
 typedef int           mp_err;
 
 /* you'll have to tune these... */
-extern int KARATSUBA_MUL_CUTOFF,
-       KARATSUBA_SQR_CUTOFF,
-       TOOM_MUL_CUTOFF,
-       TOOM_SQR_CUTOFF;
+#ifndef KARATSUBA_SQR_CUTOFF
+extern int KARATSUBA_MUL_CUTOFF;
+#endif
+#ifndef KARATSUBA_SQR_CUTOFF
+extern int KARATSUBA_SQR_CUTOFF;
+#endif
+#ifndef TOOM_MUL_CUTOFF
+extern int TOOM_MUL_CUTOFF;
+#endif
+#ifndef TOOM_SQR_CUTOFF
+extern int TOOM_SQR_CUTOFF;
+#endif
 
 /* define this to use lower memory usage routines (exptmods mostly) */
 /* #define MP_LOW_MEM */


minad:
I agree and I wanted to have something like this before. However I would like to go a little step further.

See here:
https://github.com/libtom/libtommath/commit/6800a0e4fac80fedfbaa11bf3d14a2a6cf90e034

Ping @czurnieden @sjaeckel

czurnieden:
It would be nice if it is in a separate file than can be easily overwritten by the tuning program.
The problem with default values is, that there are no default values; one-size-fits-all is not only a sartorial myth. Here are the the outputs for the two useful modes for a 64-bit machine (AMD A8-6600K):

MP_64BIT on 64-bit machine

```text
       N     min  median  mode  mean    max      sd      skew     kurt     25%     75%
   km 1000   106   134   134   136.20   162     6.58     0.13     3.54   134.00   141.00
   ks 1000   127   134   134   133.87   134     0.93    -7.25    53.57   134.00   134.00
 tc3m 1000   141   225   225   227.40   260    10.90    -2.73    22.99   225.00   232.00
 tc3s 1000   253   260   260   259.86   260     0.98    -6.86    48.02   260.00   260.00
 tc4m 1000   337   617   617   574.32   645    80.70    -1.54     3.49   610.00   617.00
 tc4s 1000   379   386   386   385.87   386     0.93    -7.25    53.57   386.00   386.00
 tc5m 1000   421   736   736   669.61   771    94.48    -0.93     2.19   596.00   736.00
 tc5s 1000   365   386   386   386.04   400     1.97     2.61    49.03   386.00   386.00
```

MP_32BIT on a 64-bit machine

```text
       N     min  median  mode  mean    max      sd      skew     kurt     25%     75%
   km 1000   260   260   260   260.00   260     0.00     -nan     -nan   260.00   260.00
   ks 1000   127   134   134   133.99   134     0.22   -31.58   998.00   134.00   134.00
 tc3m 1000   505   512   512   511.98   512     0.38   -18.18   331.34   512.00   512.00
 tc3s 1000   253   260   260   259.99   260     0.22   -31.58   998.00   260.00   260.00
 tc4m 1000   694   764   764   763.92   764     2.22   -31.14   978.43   764.00   764.00
 tc4s 1000   379   386   386   385.99   386     0.31   -22.29   498.00   386.00   386.00
 tc5m 1000   974  1023  1023  1022.88  1023     1.86   -21.77   527.00  1023.00  1023.00
 tc5s 1000   505   512   512   511.98   512     0.38   -18.18   331.34   512.00   512.00
```

The `MP_32BIT` version looks like what one would expect but the `MP_64BIT` version says, that TC-4-squaring can be safely replaced with TC-5-squaring&mdash;the software 128-bit integers change a lot here.

So it would be nice if all the cutoff variables are alone and in a single file to be easily adjusted in an automated way (yes, I'm working at it) to get some reasonable default values. The values in this PR are off by a lot, some over 100%, on my machine. They are safe, yes, but not very useful, the tuning is necessary.

The current way via `bncore.c` is simple: just evaluate the values, write `bncore.c`, compile that file and relink. Quick and easy. If you put that stuff into `tommath.h` it would be a bit more complicated to overwrite and, because you changed the main header, you need to recompile the whole lib instead of just a single file.

And yes, I understand the wish to keep the fat-fingered fiddlers from fiddling with the greasy cogs of an installed lib,  and I think this PR will do but I would really appreciate it if you make my work easier by putting it in a separate file.

\* median and mode are the same, because I run the benchmark on an isolated core (cpuset) without touching anything. Without isolation and with watching videos they can differ quite a lot.

minad:
@czurnieden Please note that this change doesn't change the status quo. You can still set the variables directly from the tuner as you do now.

The only thing that changes is that the defaults are not defined in bncore.c but in tommath.h (Now I think tommath_private.h would be better and I will move them there). The values in bncore.c haven't changed many times, therefore I guess it is ok to have them in tommath_private.h. I would like to avoid separate header files.

minad:
Closing this in favor of #231

PR #227, opened 2019-04-27
Toom/FFT multiplication

czurnieden:
Toom-Cook 4, 5 and FFT (radix-2 Hartley transform, nothing fancy)
`etc/tune` has also been updated to tune the additional algorithms, too.
Toom-Cook-3 has been updated to allow to "proof" the algorithm in the same way as Toom-Cook 4 and 5: they contain comments that can be fed into pari/gp.

PR #231, opened 2019-05-03
Tunable cutoff parameters

minad:
Follow up on #229


minad:
@czurnieden 

> The current way via bncore.c is simple: just evaluate the values, write bncore.c, compile that file and relink. Quick and easy. If you put that stuff into tommath.h it would be a bit more complicated to overwrite and, because you changed the main header, you need to recompile the whole lib instead of just a single file.

Why are you using that to do the evaluation instead of changing the variables at runtime?

czurnieden:
> Why are you using that to do the evaluation instead of changing the variables at runtime?

Hu? Oh, no, misunderstanding. That part, writing `bncore.c` that is,  comes *after* the evaluation, of course, it would be the final step. And it would be more complicated and hence error-prone to pick them out of a large file instead of simply writing a new small file with just the cut-offs in it.

It is supposed to be fully automated at the end: just type `make tune` and the magic happens ;-)

minad:
>  Hu? Oh, no, misunderstanding. That part, writing bncore.c that is, comes after the evaluation, of course, it would be the final step. And it would be more complicated and hence error-prone to pick them out of a large file instead of simply writing a new small file with just the cut-offs in it.
It is supposed to be fully automated at the end: just type make tune and the magic happens ;-)

Ok, that sounds cool. In that case we should probably go with the separate file. I think overwriting individual cutoff parameters does not make sense then and things can be simplified here.

When will we get this autotuner?

czurnieden:
> When will we get this autotuner?

When it's done, of course! ;-)

The tuning part is done (what I have so far is all in #227) it is just the writing-to-file-and-recompile-part  that needs work (not yet started, hope the weekend helps a bit with it).

minad:
@nomeata Are you ok with this outcome? The cutoffs can be overwritten manually in the tommath_cutoffs.h file instead of via -D. But we will get the autotuner soon it seems :)

czurnieden:
> The cutoffs can be overwritten manually in the tommath_cutoffs.h file instead of via -D.

What about some `ifdef` brackets in `tommath_cutoffs.h` such that they *can* be overwritten at the commandline? Too messy?

>  But we will get the autotuner soon it seems :)

Allways this pressure! ;-)

minad:
> What about some ifdef brackets in tommath_cutoffs.h such that they can be overwritten at the commandline? Too messy?

Yes I find it a bit too messy and unnecessary given the autotuner. My guess would be that nobody would be interested in overwriting these manually.

> Allways this pressure! ;-)

:grimacing:

czurnieden:
Added a (rudimentary but working) autotuner to #227 
Please take a look and if you are OK with the idea and have this PR merged I can adapt it, document it and make a separate PR `autotune`.

nomeata:
> @nomeata Are you ok with this outcome? The cutoffs can be overwritten manually in the tommath_cutoffs.h file instead of via -D. But we will get the autotuner soon it seems :) Furthermore it is possible compile the library with fixed tuning parameters.

Yes, works for me. I’ll just use `MP_FIXED_CUTOFFS` (I am not iterested in maximum performance in my application, but due to [linker troubles](https://bugs.llvm.org/show_bug.cgi?id=41726) in my environment, I need to avoid sharing global variables between object files). So I’m happy with this, thanks!

minad:
@sjaeckel This one is ready for merging from my side. Are you ok with it?
@czurnieden The autotuner looks good! Let's proceed with it in a separate PR after this one :+1:

czurnieden:
> Let's proceed with it in a separate PR after this one

Oops, to late ;-)

Rebasing and adapting would involve not more than a bit of search&replace, so I already put it up at #232 before it get's lost.

PR #232, opened 2019-05-04
Autotuning functionality for the Toom-Cook cut-offs

czurnieden:
As the name suggests: just do a `make tune` to get well-tempered cut-off values.

I am not able to test the functionality of every makefile\*, so it would be very nice if somebody with that ability would take a look at them, thank you.


etc/statistic_summary_single_column.c
```diff
@@ -0,0 +1,316 @@
...
+   It has been written for a single use case with known input and not for general usage.
+*/
+
+/*
+   Simplified sqrt within 1 ulp, no rounding; good enough to compute the SD
+
+   From a copy of a drafted paper by Prof W. Kahan and K.C. Ng, written in May, 1986.
+   That copy can be found in the file https://www.netlib.org/fdlibm/e_sqrt.c
+*/
+static double s_sqrt(double d)
```
minad: Why not libm sqrt?

etc/statistic_summary_single_column.c
```diff
@@ -0,0 +1,316 @@
...
+   cube = cube * cube * cube;
+   return ((sum/(length)) / cube);
+}
+
+/*
+   Algorithm found in Section 2.3.4 "Finding the Median" in
+   Wirth, Niklaus. "Algorithms + data structures = programs."
+   No. 04; QA76. 6, W56.. 1976.
+*/
+static long k_smallest(long *array, int length, int k)
```
minad: Why not qsort + lookup? Is this performance critical?

etc/tune_it.sh
```diff
@@ -0,0 +1,112 @@
...
+/* SPDX-License-Identifier: Unlicense */
+
+/* Known optimal configurations
+
+ CPU                    /Compiler     /MUL CUTOFF/SQR CUTOFF
+-------------------------------------------------------------
+ Intel P4 Northwood     /GCC v3.4.1   /        88/       128/LTM 0.32 ;-)
+ AMD Athlon64           /GCC v3.4.4   /        80/       120/LTM 0.35
+
+*/
```
minad: Maybe remove those old comments?

etc/tune.c
```diff
@@ -2,141 +2,740 @@
...
+      KARATSUBA_SQR_CUTOFF = INT_MAX;
+      TOOM_MUL_CUTOFF = INT_MAX;
+      TOOM_SQR_CUTOFF = INT_MAX;
+#ifdef BN_S_MP_KARATSUBA_MUL_C
+      /*
+         The influence of the Comba multiplication cannot be
+         eradicated programmatically. It depends on the size
+         of the macro MP_WPARRAY in tommath.h which needs to
+         be changed manually (to 0 (zero)).
+       */
```
minad: Could we do something similar for MP_WPARRAY and define a tunable variable?

etc/tune_it.sh
```diff
@@ -0,0 +1,112 @@
...
+
+
+
+
+
+
+
+
+
+
```
minad: Weird ws

czurnieden: So, you think those characters have mental health problems and should go?

etc/tune.c
czurnieden: Uhm&hellip;no, `MP_WARRAY` is not tunable itself and you cannot set the value to zero easily, that would need a change to how `MP_WARRAY` gets defined.

etc/tune_it.sh
czurnieden: Can do. Shall I?

etc/statistic_summary_single_column.c
czurnieden: No qsort in LTM (grep for "unwanted" in `helper.pl`). It's more of a rule for LTM itself but I had no reason to break with it here.

etc/statistic_summary_single_column.c
czurnieden: No libm in LTM.

minad: You are also using libc printf. This is the tuner and not the lib. Please just use libc for simplicity.
For the lib, I totally agree, there shouldn't be libm and only a bit of libc.

etc/statistic_summary_single_column.c
minad: This rule should be broken. The same argument as above applies.

etc/tune_it.sh
minad: Yes, they don't make sense if the values are updated?

etc/tune.c
minad: Sure. My question is - can we make it tunable?

etc/tune_it.sh
minad: Indeed

etc/statistic_summary_single_column.c
czurnieden: > This is the tuner and not the lib

It is automated now and part of the buildprocess, so the same restrictions as for LTM itself apply.

Using floating point is already questionable here because I only use the medians as the cut-off values, the rest of the statistics is only there to approximate the error which isn't even used yet for evaluation directly, only for manual inspection (answering the question: Can I trust the result?).

I am a bit torn here: throw it all out, keep it as it is (with floating point) or redo the float part with some simple fixed-point math to have both. But the last one has definitely the largest amount of work involved.

etc/statistic_summary_single_column.c
czurnieden: Yes, but as above: it is automated now and part of the buildprocess, so the same restrictions as for LTM itself apply.

etc/tune_it.sh
czurnieden: Not even for historic reasons?
No?
OK.

etc/tune.c
czurnieden: `MP_WPARRAY`? As far as I understand it, the computed value is already at the optimum but I haven't checked the current literature yet.

My problem in `tune.c` was also a different one: I wanted to use the school-multiplication for checking the results, not the Comba multiplication.

> My question is - can we make it tunable?

Who's "we"? ;-)

etc/statistic_summary_single_column.c
minad: What is automated? Please don't replicate libc functionality of it is not strictly necessary. This will only be a maintanance burden. And I am only talking about tools.

Within the lib it is perfectly reasonable to require only a libc subset which is good for embedding.

If restrictions apply they should be lifted.

etc/tune.c
minad: Generic "we" :) It doesn't mean it has to be done. It was just a question if warray would be a potential candidate for the tuner.

etc/statistic_summary_single_column.c
czurnieden: > What is automated?

The evaluation of the TC cut-off points? That's why you called it "autotune"?

Decided to strip all function not necessary to do the evaluation of the TC cut-off points, so no floats anymore. Dependencies are now down to `strtol`, `fgets` and `(f)printf`. which even the most barebone libc should offer.

Tuning must happen on the target-architecture, where, for example in the case of small MCUs not everything what gets offered by their (the IDEs of the manufacturers) libc is, how shall I put it, usable in all and every circumstance, so we should reduce the number of dependencies. Not at all cost, of course, but if it is cheap to have like here, we should do it.

Be conservative in what you do, be liberal in what you accept from others. (Jon Postel)

I was also thinking about exchanging heap for stack to get rid of `malloc` but decided against it, didn't want to implement a sliding-window median.

etc/tune_it.sh
czurnieden: If you are wondering what I had to force-push&hellip;

etc/statistic_summary_single_column.c
minad: > Decided to strip all function not necessary to do the evaluation of the TC cut-off points, so no floats anymore. Dependencies are now down to strtol, fgets and (f)printf. which even the most barebone libc should offer.

fprintf/printf usually does floating point. A barebone libc/libm usually provides sqrt.

> Tuning must happen on the target-architecture, where, for example in the case of small MCUs not everything what gets offered by their (the IDEs of the manufacturers) libc is, how shall I put it, usable in all and every circumstance, so we should reduce the number of dependencies. Not at all cost, of course, but if it is cheap to have like here, we should do it.

I agree that it should be done on the target platform. But you are prematurely assuming certain things, which adds unnecessary code here, which I am against. You assume:

- The presence of floating point in IEEE format
- The presence of fprintf
- The presence of malloc
- That sqrt does not exist

This does not make sense.

I wrote bare metal code before. And the first things I got rid were malloc and fprintf. For sqrt you can just link against a libm. This is not our business. Furthermore the version you added accesses the floating point internals which is not portable and worse than relying on the target libm.

minad: Just to make it clear - I very much appreciate the efforts to only rely on a minimal amount of libc. I am using tommath in a setting without one actually. But we should stay reasonable and code maintainability trumps without-libc-ness. In particular if it is about tooling. The library code is a harder story.

I think the following is overkill for tooling:

* Not relying on libm for tooling
* Fixed point math
* Avoiding malloc or printf

I will go through the list of the unwanted functions.

minad: The following are currently avoided in the lib:

* memcpy, memset, memmove, memcmp: I think these should not be avoided even in the lib since every target platform supports those and they will compile to intrinsics on modern compiler for short copies. However my argument is kind of weak since modern compilers will also recognize our hand written memcpy/memset loops for example. If a platform misses those, simple versions could be provided in like 5 lines of C. The only problem is that we are probably very often not doing bytewise copy, but wordwise. This is were we would get worse with a naive C implementation. But I think memset/memcpy will so or so being present in a binary since other parts of the code are using it. Then we are doing worse here since we generate bloat which would otherwise be unnecessary given a primitive compiler. So what remains in the end is maybe that these functions clearly show the intend (as clearly as a loop?).

* strcmp: Same with strcmp, should not be avoided. Also intrinsic.

* strcpy, strncpy: On these I agree. These are bad functions. They shouldn't be used anywhere.

* clock: Should be avoided obviously, this is a syscall.

* qsort: I agree on that one.

So in my opinion not much remains of the "unwanted function" list :wink:

etc/statistic_summary_single_column.c
minad: I think the median is acceptable since it's a better algorithm than qsort.

etc/statistic_summary_single_column.c
minad: It seems you removed s_sqrt and simplified things a lot. I guess I am more happy now :)

minad:
@czurnieden Could we also run "make tune" on travis just to make sure everything works :)

czurnieden:
Please drop me a note if you want it, so I can squash it into 7f6a791 for easier merge.

minad:
Yes please squash!

minad:
Thanks for simplifying! If you want you can also rebase on top of #231 and close 231 or the other way round since they are related and interdependent. Then @sjaeckel can merge both as one. If you have time, otherwise I can do it tomorrow.

czurnieden:
> If you want you can also rebase on top of #231 and close 231 or the other way round since they are related and interdependent

But that would leave nothing to do for you! ;-)

minad:
Yes but I am afk right now.

sjaeckel:
@czurnieden I rebased and fixed after the merge of #231

czurnieden:
@sjaeckel 
> I rebased and fixed after the merge of #231

Misunderstood as wanting me to do it and not that you already did it (only read the mail) and started to rebase and fix myself. Stopped after the local rebasing and did nothing else, no pushes, no commits, nothing, still have my hands in my pocket. Figuratively, of course ;-)

Do I have to do anything or can I just relax and watch the miracles unfold?

sjaeckel:
> Do I have to do anything or can I just relax and watch the miracles unfold?

https://youtu.be/Q3_2entulkw and gn8 :)

PR #234, opened 2019-05-07
More deprecations

minad:
Some minor cleanups in the public API

### DIFF
--- a/tommath_class.h
+++ b/tommath_class.h
@@ -385,8 +385,6 @@
 
 #if defined(BN_MP_FREAD_C)
 #   define BN_MP_ZERO_C
-#   define BN_MP_S_RMAP_REVERSE_SZ_C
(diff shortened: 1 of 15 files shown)


tommath_class.h
```diff
@@ -468,7 +464,6 @@
 #endif
 
 #if defined(BN_MP_INIT_MULTI_C)
-#   define BN_MP_ERR_C
```
minad: @sjaeckel There were many false positives in the dep.pl script like this one for example. Hopefully my fix removed all of them.

minad: The problem is also that comments are parsed. I should strip them from the source before applying the dep regex.

sjaeckel: looks very good!

minad: and fixed

PR #235, opened 2019-05-07
Tune fixes

minad:
Tune works well, but there a a few minor issues:

* missing break in cmdline parsing
* missing gitignore files


.gitignore
```diff
@@ -1,83 +1,86 @@
-# suppress compiler/linker output
```
minad: diff looks bad due to line ending changes. But you can select "hide whitespace changes" in the gh diff settings.

sjaeckel: ugh they were `\r\n`!?

btw. I added a separate etc/.gitignore
now it's mixed up where the stuff is ignored that is created in etc/ :-)

minad: I don't like multiple .gitignores tbh

.gitignore
```diff
@@ -57,6 +57,7 @@ mpi.c
 
 # ignore file generated by make tune
 tuning_list
+etc/tune
```
minad: @sjaeckel I hope you agree. I find this simpler and it avoids confusion. I generally use single .gitignores and only multiple if there are multiple separate projects in a repository, like a monorepo with multiple vendored projects etc.

.gitignore
sjaeckel: fine by me

.gitignore
sjaeckel: shouldn't this then also be etc/tuning_list ?

minad: Yes, but I executed tune in . and then I also wanted tuning_list to be hidden. tuning_list seems to be a unique enough name to be ignored in the whole project I thought.

czurnieden:
I missed a break? Really? Me? Please, don't tell the union! ;-)
Probably happened while cutting&pasting from`tune.c` of the faster_multiplication branch and I didn't test the "printing" function of the shorter one as thoroughly as it should have been tested, sorry for that.

Oh, and I found another one:
`   fprintf(stderr,"          -t prints comma separated results\n");`
Should read
`   fprintf(stderr,"          -t prints space (0x20) separated results\n");`
Not worth an extra PR, soooo if you wouldn’t mind? Please?

Thanks!

sjaeckel:
>  Not worth an extra PR, soooo if you wouldn’t mind? Please?

done

czurnieden:
> done

Thanks!

PR #236, opened 2019-05-07
support custom random data source via mp_rand_source

minad:
* deprecate MP_PRNG_ENABLE_LTM_RNG
* custom mp_rand_source is used always if set, which should be more aligned with user expectations
* use custom source in tune.c
* don't call random number generator once per digit, which is slow


minad:
The rasons for this PR:

* faster rand, not reading single digits
* user might want to replace random source instead of providing a fallback only, this can also be used in tune.c
* ltm_rng api is a bit weird, since it expects this callback which is passed through etc. I have seen that it's aligned with ltc, but I think the api proposed in this PR is more useful.

sjaeckel:
I really like this PR!

TBH I'd prefer to keep the API signature of the "custom random source" closer to `rng_get_bytes()` i.e. more like `ltm_rng()` is implemented...
... on the other hand I never used the `callback` parameter so probably the API of `rng_get_bytes()` should be updated ...

sjaeckel:
Btw. the reason why the "custom random source" was disabled by default is to prevent the user from shooting himself in the foot ...

minad:
I just pushed again.

> Btw. the reason why the "custom random source" was disabled by default is to prevent the user from shooting himself in the foot ...

Hmm, not really an argument isn't it? The user has to manually explicitly select a custom random source.

I think the lib could also provide multiple sources like:

* mp_rand_source_jenkins (from tune.c)
* mp_rand_source_platform

and those can be passed to mp_rand_source. The default is the safe one and this is important.

sjaeckel:
> Hmm, not really an argument isn't it? The user has to manually explicitly select a custom random source.

nope. now even less where it's not just a global variable assignment but a function call. I like it.



> I think the lib could also provide multiple sources like:
> 
>     * mp_rand_source_jenkins (from tune.c)
> 
>     * mp_rand_source_platform
> 
> 
> and those can be passed to mp_rand_source. The default is the safe one and this is important.

I'd leave this work to the user if he really wants it. `mp_rand_source` is enough

minad:
> TBH I'd prefer to keep the API signature of the "custom random source" closer to rng_get_bytes() i.e. more like ltm_rng() is implemented...
... on the other hand I never used the callback parameter so probably the API of rng_get_bytes() should be updated ...

You mean returning the size instead of an error value? I think it is more consistent returning the error but ymmv. The callback should go away.

sjaeckel:
> You mean returning the size instead of an error value? I think it is more consistent returning the error but ymmv. The callback should go away.

I meant both, but we should most likely update `rng_get_bytes()`. @karel-m what do you think?

etc/tune.c
```diff
@@ -64,19 +63,17 @@ static void s_raninit(uint64_t seed)
...
-static int s_mp_random(mp_int *a, int limbs)
+static int s_ranbuf(void *p, size_t n)
 {
-   int e = MP_OKAY;
-   if ((e = mp_grow(a, limbs + 1)) != MP_OKAY) {
-      goto LTM_ERR;
+   char *q = (char *)p;
+   while (n > 0) {
+      uint64_t x = s_ranval();
+      for (int i = 0; i < 8 && n > 0; ++i, --n) {
```
czurnieden: Has c89/c90 support been officially dropped in LTM now? Not that I would shed a single tear, just asking.

minad: Haha always trying to sneak it in. We should probably set std=c89 to enforce it.

czurnieden:
Now, that the compiler complained, I remember: `ssize_t` is (in Posix, not ISO-C and) explicitly unsigned to be able to return the `-1` from several syscalls. Should have seen that, sorry.

The failure of the test for `is_square` is probably caused by the random number being zero. I'm unsure about the problems with the montgomery_reduce test but because it fails for low_mp it's most likely the same.
Now the question is: reimplement the original "first digit is not zero" or change it to: let the user check?

BTW: I think all of the tests should print their input at failure, especially if it is non-repeatable random input. If feasible, of course, no giant numbers.

minad:
@czurnieden I fixed the issues with the casts. But what is the reason for the other test failures? I don't think the reason is that the random int is zero. Maybe I made a mistake in the random number generator?

minad:
Ah ok, the random number generator was buggy before and returned some number from the stack for digits=0.

minad:
@czurnieden Could you check please? I get errors in mp_is_square and mp_montgomery_reduce. I basically took your s_random implementation from tune and replaced the inefficient version in mp_rand. It seems correct but now we get those failures.

czurnieden:
> I get errors in mp_is_square and mp_montgomery_reduce.

With  low_mp ( 8-bit, 16-bit )?
With what input?

If it is MP_8BIT is most probably a zero but that is easily checked. MP_16BIT on the other side is a bit too large for that.

(you have put `mp_zero(a)` inside `if(digits <= 0)` instead of at the start, is there a special reason for it besides speed?)

czurnieden:
Yes, with MP_8BIT and is_square I get a zero as the input.

minad:
> The way you do it can also cause leading zeros, propose a mp_clamp() at the very end.

Done. Problem persists.

> (you have put mp_zero(a) inside if(digits <= 0) instead of at the start, is there a special reason for it besides speed?)

Moving this in the beginning fixes the issue with montgomery. Is it a guaranteed invariant that unused allocated digits must be zero?

minad:
@czurnieden 

> Propose to reinstall the little loop that guarantees the first digit to be different from zero.

This is a bad idea. I don't want random numbers to have some properties like being nonzero at some bits always. The is_square and montgomery tests should be fixed instead.

czurnieden:
>  I don't want random numbers to have some properties like being nonzero at some bits always.

I understand that but changing it would change the API quite, well, not drastically but not in a small amount, there is a huge difference between "can return zero" or "cannot return zero". It might brake a lot of user programs.

What's with LTC? I don't use it, so I cannot tell quickly but it might rely on "first digit non-zero" somewhere, please check.

> The is_square and montgomery tests should be fixed instead.

There is nothing to fix, they were based on the current API, not your new one.
They need to be changed, though, if we decide to change the API.

Yeah, nitpicking, I know ;-)

minad:
The documentation says `The random number generated with these two functions is cryptographically secure...`. No it is NOT.

The current API is  horribly broken. If programs or tests relying on the broken behavior break it is good.

minad:
@czurnieden @sjaeckel It seems that mp_rand was not meant for cryptographic in the beginning since it relied on rand(). That biased behaviour is a remnant from back then. @sjaeckel make the function safer in #103. The question is what we want here.

czurnieden:
> The question is what we want here.

If LTC has no problems with it: change the API. Well, it's not even the API as you said, the non-zero first digit behaviour is not documented. And if it is not documented, we can change it without much ado.

But may I humbly suppose that we warn the users before we do it?
It took you a moment to get the reason for the failures, didn't it? ;-)

And if LTC relies on it we should give them a chance to get prepared.

minad:
@czurnieden Ok I restored the old behavior for now and opened #237. This PR can then be merged without further changes and the behavior of mp_rand can be discussed separately.

minad:
@sjaeckel Tests are green now. If you agree with the API, it is ready to merge.

PR #242, opened 2019-05-08
deprecate mp_prime_random_ex in favor of mp_prime_rand

minad:
@czurnieden Since we have a good default random source now, we can provide a simpler api for random prime generation, such that the primes are safe for cryptographical use.


czurnieden:
> Since we have a good default random source now,[&hellip;]

Indeed.

Changes of the API need to be well documented. I don't see `doc/bn.tex` included in the list of changed files.
Hey, why the long face? You wrote it, you document it! &gt;;->

minad:
Sure I will. This is not done. I just made this PR since it is faster to tell you by code then by words. So you would generally agree with this PR? @sjaeckel?

czurnieden:
> So you would generally agree with this PR

Yes.

minad:
I guess I have to add some docs then. Long face now.

minad:
docs updated

PR #244, opened 2019-05-09
add warn_unused_result, found one missing check!

minad:
Note that this makes the API much safer and more annoying since you cannot play fast and loose.

However *if* you really want to play you can `#define MP_WUR` before the include.


bn_s_mp_toom_mul.c
```diff
@@ -52,7 +52,9 @@ int s_mp_toom_mul(const mp_int *a, const mp_int *b, mp_int *c)
       goto LBL_ERR;
    }
    mp_rshd(&b1, B);
-   (void)mp_mod_2d(&b1, MP_DIGIT_BIT * B, &b1);
+   if ((res = mp_mod_2d(&b1, MP_DIGIT_BIT * B, &b1)) != MP_OKAY) {
```
minad: this one has been found by adding the warning.

minad:
@czurnieden I added an extended comment which explains MP_WUR better.

minad:
@sjaeckel rebased

Issue #243, opened 2019-05-08
API discussions/overhaul


minad:
Starting from the discussion in #240 with @czurnieden, I thought a bit about the exposed API.

It might make sense to do a bit of an overhaul and make some functions private. In particular `mp_digit` is exposed to often in the public API and it would be more useful to provide functions taking signed ints or longs. I consider `mp_digit` to be more of an implementation detail.

* `mp_cmp_d`, `mp_add_d`, `mp_sub_d`, `mp_div_d`, `mp_mod_d` -> `mp_*_int` or `mp_*_long` (not yet available)
* ~`mp_expt_d` -> `mp_expt_int`~ (**DEPRECATED** in open #304)
* ~`mp_set` -> `mp_set_int`, `mp_set_long`, `mp_set_long_long` (already there)~ (**DONE**)
* ~`mp_init_set` -> `mp_init_set_int` (already there)~ (**DONE**)

Deprecation of the `mp_*_d` functions could be done gracefully since they can be replaced by the more general `mp_*_int` functions. 

~What could also be discussed is going to defined bit sizes since we include `stdint.h` and provide functions like `mp_add_i32` instead of `mp_add_int`/`mp_add_long`. This would make the API platform independent. However this might be a bit of a too large deviation from how things are now.~ (**DONE**)

~Furthermore `mp_set` is not needed anymore as soon as ~#221~ #253 has been implemented. Right now `mp_set` still makes sense since it is guaranteed to not allocate.~ (**KEEP**)

Then there are a few functions which seem to be internal and it might make sense to not expose them.

* ~`mp_exch`~ (**KEEP** since it pretty much harmless)
* ~`mp_rshd`, `mp_lshd` (shifting by digits)~ (**KEEP?**)
* ~`mp_prime_is_divisble` + `ltm_prime_tab` (to special and dependent on MP_PRIME_SIZE, not generally useful?)~ (**DEPRECATED** in #288)
* ~`mp_n_root_ex` (what is this fast parameter for?)~ (**DEPRECATED** in #294)
* ~`mp_n_expt_d_ex` (what is this fast parameter for?)~ (**DEPRECATED** in #294)

~Then there is the rand_digit function which is a remnant of before #236 and could be deprecated.~ (**DEPRECATED**)

~Concerning the bit manipulation functions, there is `mp_get_bit` but no `mp_set_bit`. Either add `mp_set_bit` or internalize `mp_get_bit` as `s_mp_get_bit`?~

[...]

PR #245, opened 2019-05-09
deprecate old macros, deprecate mp_rand_digit

minad:
Some first deprecations in the public API. Desired? See also #243


minad:
I think mp_digit exposure in the public api should be reduced - mp_digit should be seen more as a implementation detail, since it varies greatly depending on the target platform (8, 16, 28, 31, 32, 60 bits etc).

mp_rand_digit is one such function which works with mp_digit and therefore a candidate for deprecation. I added it here since I think it probably doesn't have many consumer under the library users.

Also note: The way I am proposing regarding deprecations always allows us to go back if many downstream users come and tell us that we deprecated something they needed. Just undeprecate and move on in that case instead of removing the deprecated api.

PR #247, opened 2019-05-09
Dont waste rand bits

sjaeckel:
Not sure though if the effort is worth the gain, but at least there's also some tests of `mp_rand` now :)

**Edit**: if you think that we should keep the implementation as is, feel free to do so! But we should at least take the two other commits.

### DIFF
--- a/demo/test.c
+++ b/demo/test.c
@@ -68,6 +68,45 @@ LBL_ERR:
    return EXIT_FAILURE;
 }
 
+static int very_random_source(void *out, size_t size)
+{
+   memset(out, 0xff, size);
+   return MP_OKAY;
+}
+
+static int test_mp_rand(void)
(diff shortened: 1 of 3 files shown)


demo/test.c
```diff
@@ -1856,6 +1896,14 @@ int unit_tests(void)
...
 
    for (i = 0; i < sizeof(test) / sizeof(test[0]); ++i) {
+      if (argc > 1) {
+         for (j = 1; j < argc; ++j) {
+            if (strstr(test[i].name, argv[j]) != NULL) {
+               break;
+            }
+         }
+         if (j == argc) continue;
+      }
```
minad: :+1:

sjaeckel:
Alright, closed in favor of #248

PR #249, opened 2019-05-09
remove some historical artifacts

minad:
* pretty.build: build system is already pretty baroque and we are
  not using this anywhere in CI etc.
* rename.sh: only used for internal renaming. can be retrieved from
  git history if needed again.
* tombc/grammer.txt: ?


sjaeckel:
>     * pretty.build: build system is already pretty baroque and we are
>       not using this anywhere in CI etc.

I forgot that this exists ... very pretty console output indeed

PR #250, opened 2019-05-09
[WIP] Deprecate bitwise functions in favor of two complement versions

minad:
The two complement versions are generalisations for negative numbers, which can be used as replacements.


minad:
@sjaeckel Note: This one should be delayed a bit. I have to think more if it makes sense.

minad:
closed in favor of much better #279

PR #252, opened 2019-05-10
No grow in mp_set_int* (2)

minad:
Based on #221, I tried to simplify some things. No padding to MP_PREC anymore. Ping @nijtmans

### DIFF
--- a/bn_mp_grow.c
+++ b/bn_mp_grow.c
@@ -11,9 +11,6 @@ int mp_grow(mp_int *a, int size)
 
    /* if the alloc size is smaller alloc more ram */
    if (a->alloc < size) {
-      /* round up to a multiple of MP_PREC */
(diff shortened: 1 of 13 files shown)


sjaeckel:
can you please also fix the travis.yml so it doesn't start two builds if the branch starts with travis/ or is that intended?

minad:
It was intended. I wanted to test both commits.

minad:
But I am not sure if travis starts every run twice for PR branches called travis/. I just had this branch as travis/ first for CI to run. It is probably a bad idea to have PR branches called travis.

sjaeckel:
> It is probably a bad idea to have PR branches called travis.

yep, that's what I meant

but I like the idea of having a way to test separate commits!

minad:
> but I like the idea of having a way to test separate commits!

Yes, therefore I called it originally like this.

PR #253, opened 2019-05-10
No grow in mp_set_int (2)

minad:
Based on #221, I tried to simplify some things. No padding to MP_PREC anymore. Ping @nijtmans

(Renamed from #252 to avoid interfering with CI)


minad:
@czurnieden Well, long long is already in the code base ;)

This PR needs careful checking, since it changes the allocation sizes (in particular no rounding to MP_PREC). But since we have valgrind in place I am pretty confident that things are alright.

nijtmans:
Thanks, @minad , for taking this further!

minad:
Thanks for taking a look!

PR #255, opened 2019-05-11
hardening: add MP_ZERO_BUFFER, MP_ZERO_DIGITS

minad:
* (!) Always zero buffer before freeing if MP_NO_ZERO_ON_FREE is not defined
* Add MP_NO_ZERO_ON_FREE to disable hardening
* Add MP_ZERO_BUFFER, MP_ZERO_DIGITS, MP_FREE_BUFFFER, MP_FREE_DIGITS
* Never use MP_FREE directly, use MP_FREE_DIGITS or MP_FREE_BUFFER
* Add MP_USE_MEMSET to use memset instead of loop
* Disable astyle backups which are annonying in the times of git

### DIFF
--- a/tommath_private.h
+++ b/tommath_private.h
@@ -10,6 +10,42 @@
 extern "C" {
 #endif
 
+/* Hardening libtommath
+ * --------------------
+ *
+ * By default memory is zeroed before calling
+ * MP_FREE to avoid leaking data. This is good
+ * practice in cryptographical applications.
+ *
+ * Note however that memory allocators used
+ * in cryptographical applications can often
+ * be configured by itself to clear memory,
+ * rendering the clearing in tommath unnecessary.
+ * See for example https://github.com/GrapheneOS/hardened_malloc
+ * and the option CONFIG_ZERO_ON_FREE.
+ *
+ * Furthermore there are applications which
+ * value performance more and want this
+ * feature to be disabled. For such applications
+ * define MP_NO_ZERO_ON_FREE during compilation.
+ */
+#ifdef MP_NO_ZERO_ON_FREE
+#  define MP_FREE_BUFFER(mem, size)   MP_FREE((mem), (size))
+#  define MP_FREE_DIGITS(mem, digits) MP_FREE((mem), sizeof (mp_digit) * (digits))
+#else
+#  define MP_FREE_BUFFER(mem, size)   do { size_t fs_ = (size); void* fm_ = (mem); if (fm_) { MP_ZERO_BUFFER(fm_, fs_); MP_FREE(fm_, fs_); } } while (0)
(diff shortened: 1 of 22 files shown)


minad:
undid the overengineering, looks reasonably simple now

tommath_private.h
```diff
@@ -10,6 +10,42 @@
...
+ * See for example https://github.com/GrapheneOS/hardened_malloc
+ * and the option CONFIG_ZERO_ON_FREE.
+ *
+ * Furthermore there are applications which
+ * value performance more and want this
+ * feature to be disabled. For such applications
+ * define MP_NO_ZERO_ON_FREE during compilation.
+ */
+#ifdef MP_NO_ZERO_ON_FREE
+#  define MP_FREE_BUFFER(mem, size)   MP_FREE((mem), (size))
```
czurnieden: Second argument to `MP_FREE` is `size_t`. Shouldn't it get at least a cast?

tommath_private.h
```diff
@@ -10,6 +10,42 @@
...
+#  define MP_FREE_DIGITS(mem, digits) MP_FREE((mem), sizeof (mp_digit) * (digits))
+#else
+#  define MP_FREE_BUFFER(mem, size)   do { size_t fs_ = (size); void* fm_ = (mem); if (fm_) { MP_ZERO_BUFFER(fm_, fs_); MP_FREE(fm_, fs_); } } while (0)
+#  define MP_FREE_DIGITS(mem, digits) do { int fd_ = (digits); void* fm_ = (mem); if (fm_) { MP_ZERO_BUFFER(fm_, sizeof (mp_digit) * (size_t)fd_); MP_FREE(fm_, sizeof (mp_digit) * (size_t)fd_); } } while (0)
+#endif
+
+#ifdef MP_USE_MEMSET
+#  include <string.h>
+#  define MP_ZERO_BUFFER(mem, size)   memset((mem), 0, (size))
```
czurnieden: Third argument to `memset` is `size_t`. Shouldn't it get at least a cast?

tommath_private.h
sjaeckel: I'd say it shouldn't get a cast here, but where required in the sources where the invocation of the macro happens.
Otherwise it's possible that something like the following complies without even throwing a warning
```
void* foo = malloc(123);
MP_FREE_BUFFER(foo, foo);
```

czurnieden: Wanted to restrict my otherwise quite verbose comments, and now I'm too shrt.

The last build failed because of a missing cast in `MP_FREE` and the culprits in the code causing it are still missing their rations of plaster-of-paris (e.g.: in `mp_fwrite()`).
Something has been done in the default branch here regarding that problem, so I asked myself why it is in one branch only and then decided to share that question publicly.

minad:
rebased and missing cast added

minad:
@sjaeckel This one should be ready.

minad:
See https://www.usenix.org/system/files/conference/usenixsecurity17/sec17-yang.pdf for issues regarding memset deadstore elim. There is explicit_bzero for example in OpenBSD. Maybe more relevant for LTC.

PR #256, opened 2019-05-12
Use Jenkins rand in test suite

minad:
Motivation for this PR:
We have our own random number gens around (jenkins/platform), just use them were appropriate instead of the bad stdlib prng. Unify them under the mp_rand/mp_rand_source interface. Fix some undefined signed integer overflows in the testsuite on the way. Activate the sanitizer to catch undefined behavior.


demo/test.c
```diff
@@ -1,4 +1,30 @@
...
+   if (s_mp_rand_source(&x, sizeof(x)) != MP_OKAY) {
+      fprintf(stderr, "s_mp_rand_source failed\n");
+      exit(EXIT_FAILURE);
+   }
+   return x;
+}
+
+static unsigned long ulabs(long x)
+{
+   return x > 0 ? (unsigned long)x : -(unsigned long)x;
```
czurnieden: Shouldn't that be 
`return x > 0 ? (unsigned long)x : (unsigned long)(-x);`
?

minad: No exactly not.

czurnieden: So you want it to fold, ok. But it is positive then, the minus makes it negative again, the return casts it implicitly (I think) and folds it again. In code:
`(unsigned long)(-(unsigned long)x)`
Why?
Or, most likely, do I just don't get it?
(Besides that: no implicit casts, please. Thank you.)

minad: No no. It it is different. Should I resolve the puzzle? ;)

czurnieden: > Should I resolve the puzzle?

If you don't mind?

minad: The problem is signed overflow for -(1<<31) or -(1<<63). Therefore we first cast to unsigned long and then do the negation. Probably this still triggers undefined behavior in some weird complement word, but it does not in mine.

We can also be more explicit, catching the special case.

~~~ c
if (x == -(1L<<31))
   return (1UL<<31);
return x < 0 ? (unsigned long)-x : (unsigned long)x;
~~~

czurnieden: Ah, now I got it, you mean  the problem with `LONG_MIN == -LONG_MIN`?

Or as the standard says:
<blockquote>
ISO/IEC 9899:2011 sec. 7.22.6.1
The abs, labs, and llabs functions compute the absolute value of an integer j. If the
result cannot be represented, the behavior is undefined.304)
</blockquote>

Footnote 304 makes note of the obvious:
<blockquote>
304) The absolute value of the most negative number cannot be represented in two's complement.
</blockquote>

I tried it only with the some of my compilers but all return  `(ULONG_MAX >> 1) + 1` for the cast `(unsigned long) LONG_MIN` which they have to otherwise your first trick wouldn't work and they wouldn't be standard compliant.

Quoting the standard again:
<blockquote>
  6.3.1.3 Signed and unsigned integers
[&hellip;]
2 Otherwise, if the new type is unsigned, the value is converted by repeatedly adding or
  subtracting one more than the maximum value that can be represented in the new type
  until the value is in the range of the new type.60)
</blockquote>

So my `return x > 0 ? (unsigned long)x : (unsigned long)(-x);` is indeed correct and, if we would allow for implicit casting, a simple `return x > 0 ? x : -x;` would also suffice.

Or `(unsigned long)labs(x)` (yes, `labs(x)` is in c89, I looked it up ;-) ) which might safe a function call if the compiler is not too stupid.

But if you think you cannot assume standard compliant behaviour for the castings: please check for `LONG_MIN` and return `(ULONG_MIN >> 1) + 1` or something in the line, don't use numbers.
I'm sure you wouldn't do that in production, I'm just soothing myself here ;-)

minad: No. -LONG_MIN already triggers undefined behavior. Btw labs was used before.

czurnieden: > No. -LONG_MIN already triggers undefined behavior

I really should look into the standard instead of just testing.
(Used `clang -Weverything` which normally catches everything, but it is probably too old, but a GCC of the same age complains even without any warnings switched on.)

> Btw labs was used before.

Now that you mention it &hellip; ;-)

minad: > I really should look into the standard instead of just testing.
(Used clang -Weverything which normally catches everything, but it is probably too old, but a GCC of the same age complains even without any warnings switched on.)

I never do that. Therefore I introduced fsanitize here to safe us from madness.

czurnieden: > I never do that.

Yeah, rub it in ;-)

Should I go back to doing [nudes](http://www.flickr.com/photos/deamentiaemundi/32909524605/) in [watercolour](http://www.flickr.com/photos/deamentiaemundi/32623761920/), [badly](http://www.flickr.com/photos/deamentiaemundi/32935789811/)?

(Don't worry, these are safe for work, but not everything there or on my blog is!)

minad: ;) Interesting idea for watercolour? How comes?

PR #258, opened 2019-05-13
Use enums/More explicit types

minad:
* Makes the API more obvious
* Allows us to catch a few implicit conversions
* 100% API/ABI backwards compatible

### DIFF
--- a/tommath_private.h
+++ b/tommath_private.h
@@ -31,7 +31,7 @@ extern "C" {
  */
 #ifdef MP_NO_ZERO_ON_FREE
 #  define MP_FREE_BUFFER(mem, size)   MP_FREE((mem), (size))
-#  define MP_FREE_DIGITS(mem, digits) MP_FREE((mem), sizeof (mp_digit) * (digits))
+#  define MP_FREE_DIGITS(mem, digits) MP_FREE((mem), sizeof (mp_digit) * (size_t)(digits))
 #else
 #  define MP_FREE_BUFFER(mem, size)                     \
 do {                                                    \
--- a/tommath_private.h
+++ b/tommath_private.h
@@ -145,7 +145,7 @@ extern void MP_FREE(void *mem, size_t size);
 extern int (*s_mp_rand_source)(void *out, size_t size);
 
 /* Minimum number of available digits in mp_int, MP_PREC >= MP_MIN_PREC */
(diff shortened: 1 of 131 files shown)


minad:
@czurnieden We could just remove MP_RANGE from the enum. Defining MP_USE_ENUM somehow selects the "modernised" API. Furthermore I will make MP_RANGE deprecated then.

tommath_private.h
```diff
@@ -31,7 +31,7 @@ extern "C" {
  */
 #ifdef MP_NO_ZERO_ON_FREE
 #  define MP_FREE_BUFFER(mem, size)   MP_FREE((mem), (size))
-#  define MP_FREE_DIGITS(mem, digits) MP_FREE((mem), sizeof (mp_digit) * (digits))
+#  define MP_FREE_DIGITS(mem, digits) MP_FREE((mem), sizeof (mp_digit) * (size_t)(digits))
 #else
 #  define MP_FREE_BUFFER(mem, size)   do { size_t fs_ = (size); void* fm_ = (mem); if (fm_) { MP_ZERO_BUFFER(fm_, fs_); MP_FREE(fm_, fs_); } } while (0)
 #  define MP_FREE_DIGITS(mem, digits) do { int fd_ = (digits); void* fm_ = (mem); if (fm_) { MP_ZERO_BUFFER(fm_, sizeof (mp_digit) * (size_t)fd_); MP_FREE(fm_, sizeof (mp_digit) * (size_t)fd_); } } while (0)
```
nijtmans: Noticing that the multiplication "sizeof (mp_digit) * (size_t)fd_"is done twice in this macro. Better do the multiplication, store it in a variable, and them use that variable twice

minad: Ok. But probably this does not matter since compilers perform CSE and similar optimizations.

minad: This has also not much to do with this PR, since it was merged in #255. Maybe change it in #259? @fperrad is improving the macro there.

minad: Ok, well it has something to do since I am also adding the missing cast here ;)
Maybe @fperrad could also cherry-pick that commit from here... It is a bit of a mess if the same code is touched from multiple PRs.

minad:
Thx @nijtmans!

sjaeckel:
> also I appreciate that @fperrad immediately linted the PR, I like that way a lot better and IMO we could keep doing this approach :-)

ah lol, @minad simply rebased on top of #259 also fine by me, but @fperrad how about doing this for future PR's? as soon as they're good to go you put your linting commits on top

PR #259, opened 2019-05-13
more linting

fperrad:


bn_mp_dr_reduce.c
```diff
@@ -58,7 +58,7 @@ int mp_dr_reduce(mp_int *x, const mp_int *n, mp_digit k)
    *tmpx1++ = mu;
 
    /* zero words above m */
-   MP_ZERO_DIGITS(tmpx1, x->used - m - 1);
+   MP_ZERO_DIGITS(tmpx1, (x->used - m) - 1);
```
minad: I mostly agree with this explicit parens, but these are weird. Are you also adding parens to `(a + b) + c`?

fperrad: here the full rational of my linter:
```
       Operator 'Name' followed by operator 'Name' is confusing.  Use
       parentheses. -- Some combinations of operators seem to be
       confusing.  For example

              a = b - c - d;
              a = b - c + d;
              a = b / c / d;
              a = b / c * d;

       tend to befuddle the reader.  To reduce confusion we recommend
       using parentheses to make the association of these operators
       explicit.  For example:

              a = (b - c) - d;
              a = (b - c) + d;
              a = (b / c) / d;
              a = (b / c) * d;

       in place of the above.
```

fperrad:
rebased + more linting

bn_mp_dr_reduce.c
minad: Ok thx! It seems I am more befuddled by the parens. But I am well aware of the relative precedences. If I want parens I use lisp ;)

tommath_private.h
```diff
@@ -33,17 +33,53 @@ extern "C" {
...
+      MP_FREE(fm_, fs_);                                \
+   }                                                    \
+} while (0)
+#  define MP_FREE_DIGITS(mem, digits)                   \
+do {                                                    \
+   int fd_ = (digits);                                  \
+   void* fm_ = (mem);                                   \
+   if (fm_ != NULL) {                                   \
+      MP_ZERO_BUFFER(fm_, sizeof(mp_digit) * (size_t)fd_); \
+      MP_FREE(fm_, sizeof(mp_digit) * (size_t)fd_);     \
```
minad: @nijtmans mentioned an unnecessary duplication of the multiplication in #258. Could we fix it here please such that we don't touch the same code multiple times?

minad: Nevermind, I have rebased by PRs #258 and #260 based on this PR. No need to do anything here.

PR #261, opened 2019-05-13
Don't check for mp_set_int()/mp_set_long() result, since it's always …

nijtmans:
…MP_OK;

### DIFF
--- a/bn_mp_init_set_int.c
+++ b/bn_mp_init_set_int.c
@@ -7,9 +7,9 @@
 int mp_init_set_int(mp_int *a, unsigned long b)
 {
    int err;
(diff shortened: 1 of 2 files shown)


minad:
Right, I already modified almost all of those. These two I intentionally did not modify since control is just transferred as a tail call to mp_set_long and mp_set_int. Not merging this PR should result in better code being generated ;)

But as soon as mp_set_long or the corresponding replacement returns void (see discussion in #243) this PR is good to go.

nijtmans:
Hm.  I didn't think about code generation efficiency, no big deal...   So, I better close this one. Thanks!

minad:
Oh ok. It really should not matter.

PR #263, opened 2019-05-14
more linting

fperrad:


bn_mp_clear.c
```diff
@@ -7,7 +7,7 @@
 void mp_clear(mp_int *a)
 {
    /* only do anything if a hasn't been freed previously */
-   if (a->dp != NULL) {
+   if ((a != NULL) && (a->dp != NULL)) {
```
minad: a==NULL should not be handled here since it is a programming error.

bn_mp_reduce_is_2k.c
```diff
@@ -29,8 +29,10 @@ mp_bool mp_reduce_is_2k(const mp_int *a)
             iz = 1;
          }
       }
+      return MP_YES;
+   } else {
+      return MP_YES;
```
minad: Why?

bn_mp_clear.c
fperrad: in **libtomcrypt**, we could write :
```
LTC_ARGCHK(a != NULL);
```

but sadly **libtommath** have no equivalent.

minad: Yes, we don't have such assertions. However the change you made masks a programming error and is therefore not good. You should revert this change.

minad: If you want to harden the library, what I would recommend - define MP_ASSERT(x) in tommath.h which can be optionally activated in hardened/debugging mode. Then add MP_ASSERTS where it makes sense.

minad: I just checked LTC, and they are doing exactly what I suggested - aborting in the case of assertion failure, but not masking the error and just returning without any message.

fperrad:
commit on `bn_mp_clear.c` removed

PR #265, opened 2019-05-16
faster Toom-Cook 3 algorithms

czurnieden:
Just brought the implementation up to state-of-the-art, nothing else changed.
Halfs the cut-off point and brings it therefore into the expected region. The expected region is a bit less than twice the Karatsuba cut-off point and that is where it is now.

This code was part of #227.

### DIFF
--- a/bn_s_mp_toom_mul.c
+++ b/bn_s_mp_toom_mul.c
@@ -10,263 +10,260 @@
  * only particularly useful on VERY large inputs
  * (we're talking 1000s of digits here...).
 */
+
+/*
+   This file contains code from J. Arndt's book  "Matters Computational"
+   and the accompanying FXT-library with permission of the author.
+*/
+
+/*
+   Setup from
+
+     Chung, Jaewook, and M. Anwar Hasan. "Asymmetric squaring formulae."
+     18th IEEE Symposium on Computer Arithmetic (ARITH'07). IEEE, 2007.
+
+   The interpolation from above needed one temporary variable more
+   than the interpolation here:
+
+     Bodrato, Marco, and Alberto Zanoni. "What about Toom-Cook matrices optimality."
+     Centro Vito Volterra Universita di Roma Tor Vergata (2006)
+*/
+
 mp_err s_mp_toom_mul(const mp_int *a, const mp_int *b, mp_int *c)
 {
-   mp_int w0, w1, w2, w3, w4, tmp1, tmp2, a0, a1, a2, b0, b1, b2;
-   int B;
-   mp_err err;
+   mp_int S1, S2, T1, a0, a1, a2, b0, b1, b2;
+   int err, B, count;
 
    /* init temps */
-   if ((err = mp_init_multi(&w0, &w1, &w2, &w3, &w4,
-                            &a0, &a1, &a2, &b0, &b1,
-                            &b2, &tmp1, &tmp2, NULL)) != MP_OKAY) {
+   if ((err = mp_init_multi(&S1, &S2, &T1, NULL)) != MP_OKAY) {
       return err;
    }
 
    /* B */
    B = MP_MIN(a->used, b->used) / 3;
 
-   /* a = a2 * B**2 + a1 * B + a0 */
-   if ((err = mp_mod_2d(a, MP_DIGIT_BIT * B, &a0)) != MP_OKAY) {
-      goto LBL_ERR;
+   /** a = a2 * x^2 + a1 * x + a0; */
+   if ((err = mp_init_size(&a0, B)) != MP_OKAY) {
+      goto LTM_ERRa0;
+   }
+   for (count = 0; count < B; count++) {
+      a0.dp[count] = a->dp[count];
+      a0.used++;
+   }
+   mp_clamp(&a0);
+   if ((err = mp_init_size(&a1, B)) != MP_OKAY) {
+      goto LTM_ERRa1;
+   }
+   for (; count < (2 * B); count++) {
+      a1.dp[count - B] = a->dp[count];
+      a1.used++;
+   }
+   mp_clamp(&a1);
+   if ((err = mp_init_size(&a2, B + (a->used - (3 * B)))) != MP_OKAY) {
+      goto LTM_ERRa2;
+   }
+   for (; count < a->used; count++) {
+      a2.dp[count - 2 * B] = a->dp[count];
+      a2.used++;
    }
 
-   if ((err = mp_copy(a, &a1)) != MP_OKAY) {
-      goto LBL_ERR;
+   /** b = b2 * x^2 + b1 * x + b0; */
+   if ((err = mp_init_size(&b0, B)) != MP_OKAY) {
+      goto LTM_ERRb0;
    }
-   mp_rshd(&a1, B);
-   if ((err = mp_mod_2d(&a1, MP_DIGIT_BIT * B, &a1)) != MP_OKAY) {
-      goto LBL_ERR;
+   for (count = 0; count < B; count++) {
+      b0.dp[count] = b->dp[count];
+      b0.used++;
    }
-
-   if ((err = mp_copy(a, &a2)) != MP_OKAY) {
-      goto LBL_ERR;
+   mp_clamp(&b0);
+   if ((err = mp_init_size(&b1, B)) != MP_OKAY) {
+      goto LTM_ERRb1;
    }
-   mp_rshd(&a2, B*2);
-
-   /* b = b2 * B**2 + b1 * B + b0 */
-   if ((err = mp_mod_2d(b, MP_DIGIT_BIT * B, &b0)) != MP_OKAY) {
-      goto LBL_ERR;
+   for (; count < (2 * B); count++) {
+      b1.dp[count - B] = b->dp[count];
+      b1.used++;
    }
-
-   if ((err = mp_copy(b, &b1)) != MP_OKAY) {
-      goto LBL_ERR;
+   mp_clamp(&b1);
+   if ((err = mp_init_size(&b2, B + (b->used - (3 * B)))) != MP_OKAY) {
+      goto LTM_ERRb2;
    }
-   mp_rshd(&b1, B);
-   if ((err = mp_mod_2d(&b1, MP_DIGIT_BIT * B, &b1)) != MP_OKAY) {
-      goto LBL_ERR;
-   }
-
-   if ((err = mp_copy(b, &b2)) != MP_OKAY) {
-      goto LBL_ERR;
-   }
-   mp_rshd(&b2, B*2);
-
-   /* w0 = a0*b0 */
-   if ((err = mp_mul(&a0, &b0, &w0)) != MP_OKAY) {
-      goto LBL_ERR;
+   for (; count < b->used; count++) {
+      b2.dp[count - 2 * B] = b->dp[count];
+      b2.used++;
    }
 
-   /* w4 = a2 * b2 */
-   if ((err = mp_mul(&a2, &b2, &w4)) != MP_OKAY) {
-      goto LBL_ERR;
+   /** \\ S1 = (a2+a1+a0) * (b2+b1+b0); */
+   /** T1 = a2 + a1; */
+   if ((err = mp_add(&a2, &a1, &T1)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   /** S2 = T1 + a0; */
+   if ((err = mp_add(&T1, &a0, &S2)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   /** c = b2 + b1; */
+   if ((err = mp_add(&b2, &b1, c)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   /** S1 = c + b0; */
+   if ((err = mp_add(c, &b0, &S1)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   /** S1 = S1 * S2; */
+   if ((err = mp_mul(&S1, &S2, &S1)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   /** \\S2 = (4*a2+2*a1+a0) * (4*b2+2*b1+b0); */
+   /** T1 = T1 + a2; */
+   if ((err = mp_add(&T1, &a2, &T1)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   /** T1 = T1 << 1; */
+   if ((err = mp_mul_2(&T1, &T1)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   /** T1 = T1 + a0; */
+   if ((err = mp_add(&T1, &a0, &T1)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   /** c = c + b2; */
+   if ((err = mp_add(c, &b2, c)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   /** c = c << 1; */
+   if ((err = mp_mul_2(c, c)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   /** c = c + b0; */
+   if ((err = mp_add(c, &b0, c)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   /** S2 = T1 * c; */
+   if ((err = mp_mul(&T1, c, &S2)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   /** \\S3 = (a2-a1+a0) * (b2-b1+b0); */
+   /** a1 = a2 - a1; */
+   if ((err = mp_sub(&a2, &a1, &a1)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   /** a1 = a1 + a0; */
+   if ((err = mp_add(&a1, &a0, &a1)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   /** b1 = b2 - b1; */
+   if ((err = mp_sub(&b2, &b1, &b1)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   /** b1 = b1 + b0; */
+   if ((err = mp_add(&b1, &b0, &b1)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   /** a1 = a1 * b1; */
+   if ((err = mp_mul(&a1, &b1, &a1)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   /** b1 = a2 * b2; */
+   if ((err = mp_mul(&a2, &b2, &b1)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   /** \\S2 = (S2 - S3)/3; */
+   /** S2 = S2 - a1; */
+   if ((err = mp_sub(&S2, &a1, &S2)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   /** S2 = S2 / 3; \\ this is an exact division  */
+   if ((err = mp_div_3(&S2, &S2, NULL)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   /** a1 = S1 - a1; */
+   if ((err = mp_sub(&S1, &a1, &a1)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   /** a1 = a1 >> 1; */
+   if ((err = mp_div_2(&a1, &a1)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   /** a0 = a0 * b0; */
+   if ((err = mp_mul(&a0, &b0, &a0)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   /** S1 = S1 - a0; */
+   if ((err = mp_sub(&S1, &a0, &S1)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   /** S2 = S2 - S1; */
+   if ((err = mp_sub(&S2, &S1, &S2)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   /** S2 = S2 >> 1; */
+   if ((err = mp_div_2(&S2, &S2)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   /** S1 = S1 - a1; */
+   if ((err = mp_sub(&S1, &a1, &S1)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   /** S1 = S1 - b1; */
+   if ((err = mp_sub(&S1, &b1, &S1)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   /** T1 = b1 << 1; */
+   if ((err = mp_mul_2(&b1, &T1)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   /** S2 = S2 - T1; */
+   if ((err = mp_sub(&S2, &T1, &S2)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   /** a1 = a1 - S2; */
+   if ((err = mp_sub(&a1, &S2, &a1)) != MP_OKAY) {
+      goto LTM_ERR;
    }
 
-   /* w1 = (a2 + 2(a1 + 2a0))(b2 + 2(b1 + 2b0)) */
-   if ((err = mp_mul_2(&a0, &tmp1)) != MP_OKAY) {
-      goto LBL_ERR;
+   /** P = b1*x^4+ S2*x^3+ S1*x^2+ a1*x + a0; */
+   if ((err = mp_lshd(&b1, 4 * B)) != MP_OKAY) {
+      goto LTM_ERR;
    }
-   if ((err = mp_add(&tmp1, &a1, &tmp1)) != MP_OKAY) {
-      goto LBL_ERR;
+   if ((err = mp_lshd(&S2, 3 * B)) != MP_OKAY) {
+      goto LTM_ERR;
    }
-   if ((err = mp_mul_2(&tmp1, &tmp1)) != MP_OKAY) {
-      goto LBL_ERR;
+   if ((err = mp_add(&b1, &S2, &b1)) != MP_OKAY) {
+      goto LTM_ERR;
    }
-   if ((err = mp_add(&tmp1, &a2, &tmp1)) != MP_OKAY) {
-      goto LBL_ERR;
+   if ((err = mp_lshd(&S1, 2 * B)) != MP_OKAY) {
+      goto LTM_ERR;
    }
-
-   if ((err = mp_mul_2(&b0, &tmp2)) != MP_OKAY) {
-      goto LBL_ERR;
+   if ((err = mp_add(&b1, &S1, &b1)) != MP_OKAY) {
+      goto LTM_ERR;
    }
-   if ((err = mp_add(&tmp2, &b1, &tmp2)) != MP_OKAY) {
-      goto LBL_ERR;
+   if ((err = mp_lshd(&a1, 1 * B)) != MP_OKAY) {
+      goto LTM_ERR;
    }
-   if ((err = mp_mul_2(&tmp2, &tmp2)) != MP_OKAY) {
-      goto LBL_ERR;
+   if ((err = mp_add(&b1, &a1, &b1)) != MP_OKAY) {
+      goto LTM_ERR;
    }
-   if ((err = mp_add(&tmp2, &b2, &tmp2)) != MP_OKAY) {
-      goto LBL_ERR;
-   }
-
-   if ((err = mp_mul(&tmp1, &tmp2, &w1)) != MP_OKAY) {
-      goto LBL_ERR;
+   if ((err = mp_add(&b1, &a0, c)) != MP_OKAY) {
+      goto LTM_ERR;
    }
 
-   /* w3 = (a0 + 2(a1 + 2a2))(b0 + 2(b1 + 2b2)) */
-   if ((err = mp_mul_2(&a2, &tmp1)) != MP_OKAY) {
-      goto LBL_ERR;
-   }
-   if ((err = mp_add(&tmp1, &a1, &tmp1)) != MP_OKAY) {
-      goto LBL_ERR;
-   }
-   if ((err = mp_mul_2(&tmp1, &tmp1)) != MP_OKAY) {
-      goto LBL_ERR;
-   }
-   if ((err = mp_add(&tmp1, &a0, &tmp1)) != MP_OKAY) {
-      goto LBL_ERR;
-   }
-
-   if ((err = mp_mul_2(&b2, &tmp2)) != MP_OKAY) {
-      goto LBL_ERR;
-   }
-   if ((err = mp_add(&tmp2, &b1, &tmp2)) != MP_OKAY) {
-      goto LBL_ERR;
-   }
-   if ((err = mp_mul_2(&tmp2, &tmp2)) != MP_OKAY) {
-      goto LBL_ERR;
-   }
-   if ((err = mp_add(&tmp2, &b0, &tmp2)) != MP_OKAY) {
-      goto LBL_ERR;
-   }
-
-   if ((err = mp_mul(&tmp1, &tmp2, &w3)) != MP_OKAY) {
-      goto LBL_ERR;
-   }
+   /** a * b - P */
 
 
-   /* w2 = (a2 + a1 + a0)(b2 + b1 + b0) */
-   if ((err = mp_add(&a2, &a1, &tmp1)) != MP_OKAY) {
-      goto LBL_ERR;
-   }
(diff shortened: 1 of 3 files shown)


bn_s_mp_toom_sqr.c
```diff
@@ -4,209 +4,177 @@
...
+   if ((e = mp_add(&tmp, &S0, &tmp)) != MP_OKAY) {
+      goto LTM_ERR;
    }
+   /** a^2 - P  */
+   mp_exch(&tmp, b);
 
-LBL_ERR:
-   mp_clear_multi(&w0, &w1, &w2, &w3, &w4, &a0, &a1, &a2, &tmp1, NULL);
-   return res;
+LTM_ERR:
```
minad: I preferred the old version here for simplicity ;)

minad:
If what you are saying about the cutoff points, this is good :+1:

How is test coverage? Maybe add a tiny test specifically calling s_mul_karatsuba and s_mul_toom and comparing for very large numbers?

czurnieden:
> If what you are saying about the cutoff points, this is good

I have updated my home-machine, if you remember and since then, some things work&hellip;a bit differently, to say the least.
The multiplication algorithms work as expected but the cut-off points of the squaring algorithms went through the roof (Tom's original ones and mine).
First observation: no swapping; the swap partition stays empty and I mean *empty*, zero bytes.
So Linux Mint 19.1 is optimized for SSDs. Will need to check if Ubuntu Bionic has the same defaults and might need to act accordingly.

And no, I don't know yet why it is squaring only but I have some suspicions.

Can you do me the favour and, if you have access to Bionic (virtual won't work here), run `tune`?

There is also the patch for this CPU branching problem (Meltdown, spectre, or whatever they are called) but Mint's repository offers no kernel without the patches, have to build one myself and that will take some time. But I doubt it will have a large effect here.

> Maybe add a tiny test specifically calling s_mul_karatsuba and s_mul_toom and comparing for very large numbers?

`tune.c` has testing abilities build in (just a `mp_cmp` against already generated values). I would like to call it for that purpose instead of mirroring the relevant code in `test.c` but I need to take it apart for it.

Was already working at it ( it's all in `main`, it's quite some spaghetti code and in urgent need of refactoring ;-) ) when I stumbled over the problem described above.
Quite a WTF moment, I can tell you!

czurnieden:
Got to something acceptable by reducing the amount of variables to one, but it is still not known what the precise cause is.

New version:
```text
> summary(data)
       km               ks             tc3m            tc3s      
 Min.   : 85.00   Min.   :111.0   Min.   :167.0   Min.   :162.0  
 1st Qu.: 92.00   1st Qu.:126.0   1st Qu.:175.0   1st Qu.:200.8  
 Median : 93.00   Median :128.0   Median :178.0   Median :213.0  
 Mean   : 93.83   Mean   :125.8   Mean   :178.7   Mean   :213.0  
 3rd Qu.: 97.00   3rd Qu.:128.0   3rd Qu.:182.0   3rd Qu.:220.0  
 Max.   :104.00   Max.   :128.0   Max.   :191.0   Max.   :253.0  
> apply(data, 2, sd)
       km        ks      tc3m      tc3s 
 3.544964  4.493261  5.179710 14.389895
```

minad:
@czurnieden This is also still work in progress? I added a few labels to the PRs. I hope you don't mind. Feel free to remove them if you think you are done!

minad:
@czurnieden I saw you mentioned gprof - I recommend perf these days, which doesn't require recompilation and has other advantages afaik.

czurnieden:
> This is also still work in progress?

I'm waiting for the approval&merging of #280 to rebase on that.
What's the problem with #280?
Oh.
\*blushes\*

>  I saw you mentioned gprof - I recommend perf these days

Am just used to it. Don't need more than a short glimpse at the output to see what the problem/bottleneck is. And recompiling? Don't use it very often, once or twice a year max, so that doesn't matter.

minad:
> Am just used to it. Don't need more than a short glimpse at the output to see what the problem/bottleneck is. And recompiling? Don't use it very often, once or twice a year max, so that doesn't matter.

Ah ok. I profile much more often and like it to have these cpu perf counter infos. I like to have timings per asm instruction from time to time ;)

czurnieden:
With `git merge` it is `-Xours`  but with `git rebase` it is `-Xtheirs`. I hope I've chosen the right one ;-)

bn_s_mp_toom_sqr.c
sjaeckel: me too

bn_s_mp_toom_mul.c
```diff
@@ -10,263 +10,260 @@
...
-   }
-   if ((err = mp_add(&tmp1, c, c)) != MP_OKAY) {
-      goto LBL_ERR;
-   }
 
-LBL_ERR:
-   mp_clear_multi(&w0, &w1, &w2, &w3, &w4,
-                  &a0, &a1, &a2, &b0, &b1,
-                  &b2, &tmp1, &tmp2, NULL);
+LTM_ERR:
```
sjaeckel: same here

minad:
> With `git merge` it is `-Xours` but with `git rebase` it is `-Xtheirs`. I hope I've chosen the right one ;-)
> 
> Oh, that dreaded `callgraph.txt` again *grrr*

czurnieden:
> I thought before about removing [`callgraph.txt`]

It is a "nice to have" but I'm pretty sure any of the better IDEs around are able to it and do it much better. Should be generated on demand only.

I'll take a look at it later.

minad:
> It is a "nice to have" but I'm pretty sure any of the better IDEs around are able to it and do it much better. Should be generated on demand only.

I would keep generating it and just gitignore it and remove it from git. But we also have tommath_class.h which is also bad, but not as bad.

czurnieden:
> I would keep generating it and just gitignore it and remove it from git

Was quickly done, see #283 

> But we also have tommath_class.h which is also bad, but not as bad.

Yeah, true.
But that needs a bit more work to get rid off.

sjaeckel:
@minad fine for you to merge even with the multi-clears instead of a single one?

czurnieden:
> even with the multi-clears instead of a single one?

There is no `mp_init_size_multi`?

sjaeckel:
> There is no `mp_init_size_multi`?

probably `mp_init_multi()` and `mp_grow()`?

czurnieden:
> probably mp_init_multi() and mp_grow()?

That was 5 seconds later, max! ;-)
OK \*sigh\*

minad:
> @minad fine for you to merge even with the multi-clears instead of a single one?

I really don't like it since this is very error prone if something is changed. What is the problem in changing it?

I don't want to slow things down here, but if you want to merge things - there are quite a few PRs which are ready, e.g. #279, #276, #271, #275.

minad:
#266 seems also ready. Not that you misunderstand me - I fully approve this PR, but no hurry - until two hours ago  @czurnieden still seemed to be doing performance optimization and profiling.

czurnieden:
> until two hours ago

That was two hours ago and just rebasing. Performance optimization, or better: make it work at all under the new memory managment, has been done yesterday.

> What is the problem in changing it?

It's work! ;-)
(I'm already at it, may need a minute or two)

minad:
Sorry @czurnieden - already corrected. It still didn't feel ready, but I wasn't really involved here so I cannot judge.

czurnieden:
> What is the problem in changing it?

`mp_grow` is about 10% slower

```text
With mp_grow
#define MP_DEFAULT_KARATSUBA_MUL_CUTOFF 95
#define MP_DEFAULT_KARATSUBA_SQR_CUTOFF 127
#define MP_DEFAULT_TOOM_MUL_CUTOFF      180
#define MP_DEFAULT_TOOM_SQR_CUTOFF      238

With mp_init_size
#define MP_DEFAULT_KARATSUBA_MUL_CUTOFF 96
#define MP_DEFAULT_KARATSUBA_SQR_CUTOFF 125
#define MP_DEFAULT_TOOM_MUL_CUTOFF      158
#define MP_DEFAULT_TOOM_SQR_CUTOFF      216
```

Still change to init_mult/grow?

sjaeckel:
> Still change to init_mult/grow?

IMO no

minad:
> Still change to init_mult/grow?

Just a technical question - I wonder about that. What is causing the slowdown? More allocations? If you access the malloc cache I wouldn't expect such a big difference. I had expect that all the time is eaten up by the numerics code.

czurnieden:
> >   Still change to init_mult/grow?

> IMO no

I know that realloc is a bit slower than a fresh allocation, but it is quite a lot, indeed!

minad:
Well realloc can be much faster but if you really realloc, it is as slow as free+malloc. However with modern allocators usually allocations are taken from a "hot" cache and not that slow then. In particular in your tuner, you are doing many times the same thing. Therefore there should be plenty of touched memory ready.

minad:
If allocations are really an issue and mp_init_size_multi would solve that, why not consider that?

sjaeckel:
> Just a technical question - I wonder about that. What is causing the slowdown? More allocations? If you access the malloc cache I wouldn't expect such a big difference. I had expect that all the time is eaten up by the numerics code.

> I know that realloc is a bit slower than a fresh allocation, but it is quite a lot, indeed!

okay, if both of you are wondering about this I'll merge later

minad:
I haven't really looked I am only speculating (I can look later). If you allocate a lot of memory before you free stuff again, it will get very slow since the OS has to supply memory. But in this case you have many alloc/free pairs, right? How much memory are we talking btw? Are your blocks larger than page size or maybe multiple times larger? Then malloc could directly do mmap/munmap. In particular munmap is unbelievably slow on linux (I had a case where I did munmapping in a background thread for that reason). Did you strace?

sjaeckel:
this change brought the tune build-job seriously down from >7min to <2min?

that's pretty impressive!

czurnieden:
> Did you strace?

I had no time yet to do anything to look into it, most of it is based on educated guesses.
But it's interesting, so I *will* dig deeper, no worry.

I also need to read through the documentations because I just updated from the old LTS to the new one, that is I skipped one LTS release. New kernel, new LibC, new all, I'll need a couple of days to get sufficiently up-to-date with all the gory details and there are a *lot* of entrails to dig through ;-)

> this change brought the tune build-job seriously down from >7min to <2min?

You complained, I delivered ;-)

minad:
> this change brought the tune build-job seriously down from >7min to <2min?
>> You complained, I delivered ;-)

Oh no. Now we have to reorder the travis yamml again

sjaeckel:
should I merge or leave this open until the investigation by @czurnieden is finished?

minad:
My only complaint are those many labels and mp_clear. But I don't want to stand in the way here. The investigation and maybe other improvements could still happen after merging.

sjaeckel:
@czurnieden you can decide whether you want to put in more effort or if it's fine like this for you, please rebase and remove the WIP label if you think it's ready :) 

> maybe other improvements could still happen after merging

and that's also true

czurnieden:
> you can decide whether you want to put in more effort or if it's fine like this for you
It works, it is fast, it is ugly.
Two out of three is good enough for me ;-)

> please rebase

But I just reba&hellip;oh.
I'm really envious of the large amount of free time y'all seem to have ;-)
Rebased, will wait until Travis shows all green to remove the label.


Re. investigation: it seems as if the new GLibC is indeed under a new memory-management. Will compile an old one (glibc-2.23) for a short test and dig through the sources if that test supports my suspicions. That will take a couple of days; compiling the GlibC alone is not a small task.

But the result of that investigation won't be of any large relevance for my code (would make me wonder if any at all) and if it does: "maybe other improvements could still happen after merging". The main branch here has not been named "develop" without a reason ;-)

minad:
@czurnieden I did an experiment concerning allocations and running your tuner. I managed to get rid of all allocations in mp_init and all allocations smaller than MP_MIN_PREC. With allocations 13.172s, without allocations 12.446s. So all these (often needless) small allocations we are doing during init cost as around 5%.

minad:
Then I did another experiment where I replaced malloc with a small shim which takes allocations smaller than MP_PREC from a simple cache (singly linked list). With that I got at one run 12.6s.

minad:
Jemalloc 11.8s
Jemalloc+MP_STATIC_PREC 11.7s

But take the numbers with a grain of salt. No statistics, just for a rough picture. Jemalloc it seems has a very good cache. Better at least than my naive one and the MP_STATIC_PREC gives nearly no advantage.

czurnieden:
The Toom-Cook algorithms do not need a lot of small allocation, the numbers are quite large, so your 5% is probably all you can get without a specialized memory management.
GlibC's one got quite good over the last 10 years, I would have guessed that you would gain much more with a speed optimized malloc implementation.

Tomfastmath could make much better use of an inbuild memory management, but there is already a discussion about that if I remember it correctly.

But I did as you told me and run some tracing programs. The most useful for C&P'ing the results here is probably `ltrace`, the others (strace and valgrind with the tools callgrind and cachegrind) have either too little or way too much information. There are more programs, I know, but I'm still in the "bracketing" stage, reducing the problem set, so ltrace is sufficient.

The first four lines, the ones with the `#define`s, are the output of `make tune`, ltrace gets a fixed-rounds `tune` to look at.

Single rounds with the version of tc3mul and tc3qsr I started this branch with.
Optimization -O3
```text
#define MP_DEFAULT_KARATSUBA_MUL_CUTOFF 97
#define MP_DEFAULT_KARATSUBA_SQR_CUTOFF 126
#define MP_DEFAULT_TOOM_MUL_CUTOFF      189
#define MP_DEFAULT_TOOM_SQR_CUTOFF      892 <

$ ltrace -S -c ./tune -L 3 -r 10 -M 100 -m 11
% time     seconds  usecs/call     calls      function
------ ----------- ----------- --------- --------------------
 26.92    0.266543          63      4228 memset
 24.36    0.241186          63      3792 free
 24.04    0.238011          62      3792 calloc
 16.93    0.167600          63      2620 realloc
  6.66    0.065979          62      1050 memcpy
  0.92    0.009141          63       144 clock_gettime
  0.05    0.000482         120         4 __printf_chk
  0.03    0.000262          65         4 __errno_location
  0.03    0.000262          65         4 strtol
  0.01    0.000112          18         6 SYS_brk
  0.01    0.000103          20         5 SYS_mmap
  0.01    0.000089          22         4 SYS_write
  0.01    0.000081          20         4 SYS_mprotect
  0.01    0.000061          20         3 SYS_access
  0.00    0.000048          16         3 SYS_fstat
  0.00    0.000047          23         2 SYS_openat
  0.00    0.000029          14         2 SYS_close
  0.00    0.000026          26         1 SYS_munmap
  0.00    0.000019          19         1 SYS_read
  0.00    0.000014          14         1 SYS_arch_prctl
------ ----------- ----------- --------- --------------------
100.00    0.990095                 15670 total
```
As you can see: TC3-sqr is quite an outlier

Single rounds with the version of tc3mul and tc3qsr I started this branch with
Optimization -O2
(The timings are not really relevant here, the number and kind of calls is)

```text
#define MP_DEFAULT_KARATSUBA_MUL_CUTOFF 160 <
#define MP_DEFAULT_KARATSUBA_SQR_CUTOFF 128
#define MP_DEFAULT_TOOM_MUL_CUTOFF      228
#define MP_DEFAULT_TOOM_SQR_CUTOFF      255

$ ltrace -S -c ./tune -L 3 -r 10 -M 100 -m 11
% time     seconds  usecs/call     calls      function
------ ----------- ----------- --------- --------------------
 36.76    0.232124          61      3792 calloc
 36.46    0.230235          60      3792 free
 25.12    0.158633          60      2620 realloc
  1.41    0.008935          62       144 clock_gettime
  0.07    0.000463         115         4 __printf_chk
  0.04    0.000271          67         4 __errno_location
  0.04    0.000265          66         4 strtol
  0.02    0.000116          19         6 SYS_brk
  0.01    0.000092          18         5 SYS_mmap
  0.01    0.000086          21         4 SYS_write
  0.01    0.000074          18         4 SYS_mprotect
  0.01    0.000063          21         3 SYS_access
  0.01    0.000048          24         2 SYS_openat
  0.01    0.000045          15         3 SYS_fstat
  0.00    0.000027          13         2 SYS_close
  0.00    0.000024          24         1 SYS_munmap
  0.00    0.000018          18         1 SYS_read
  0.00    0.000014          14         1 SYS_arch_prctl
------ ----------- ----------- --------- --------------------
100.00    0.631533                 10392 total
```
Here Kara-mul is the outlier. With the exception of Kara-mul it is also the result I got with my old system. (got me a Glibc 2.23 compiled \*pooh\*, no difference. Will try it with an old kernel later)

With the current versions and `-O3`

Single rounds mp_init_size

```text
$ ltrace -S -c  ./tune -L 3 -r 10 -M 100 -m 11
% time     seconds  usecs/call     calls      function
------ ----------- ----------- --------- --------------------
 30.62    0.313655          67      4615 memset
 21.00    0.215102          68      3145 realloc
 20.16    0.206486          67      3072 free
 20.14    0.206261          67      3072 calloc
  6.93    0.070965          67      1050 memcpy
  0.96    0.009820          68       144 clock_gettime
  0.05    0.000508         127         4 __printf_chk
  0.04    0.000402         100         4 __errno_location
  0.04    0.000387          96         4 strtol
  0.01    0.000121          20         6 SYS_brk
  0.01    0.000108          21         5 SYS_mmap
  0.01    0.000089          22         4 SYS_mprotect
  0.01    0.000086          21         4 SYS_write
  0.01    0.000067          22         3 SYS_access
  0.01    0.000055          18         3 SYS_fstat
  0.00    0.000046          23         2 SYS_openat
  0.00    0.000035          17         2 SYS_close
  0.00    0.000029          29         1 SYS_munmap
  0.00    0.000019          19         1 SYS_read
  0.00    0.000019          19         1 SYS_arch_prctl
------ ----------- ----------- --------- --------------------
100.00    1.024260                 15142 total
```

And with mp_grow
```text

$ ltrace -S -c ./tune -L 3 -r 10 -M 100 -m 11
% time     seconds  usecs/call     calls      function
------ ----------- ----------- --------- --------------------
 28.81    0.264157          66      3945 memset
 21.88    0.200644          65      3072 free
 21.85    0.200405          65      3072 calloc
 18.54    0.170003          68      2475 realloc
  7.66    0.070236          66      1050 memcpy
  1.04    0.009559          66       144 clock_gettime
  0.05    0.000501         125         4 __printf_chk
  0.05    0.000431         107         4 __errno_location
  0.04    0.000388          97         4 strtol
  0.01    0.000124          20         6 SYS_brk
  0.01    0.000120          24         5 SYS_mmap
  0.01    0.000096          24         4 SYS_mprotect
  0.01    0.000079          19         4 SYS_write
  0.01    0.000074          24         3 SYS_access
  0.01    0.000054          18         3 SYS_fstat
  0.01    0.000053          26         2 SYS_openat
  0.00    0.000036          18         2 SYS_close
  0.00    0.000032          32         1 SYS_munmap
  0.00    0.000023          23         1 SYS_read
  0.00    0.000018          18         1 SYS_arch_prctl
------ ----------- ----------- --------- --------------------
100.00    0.917033                 13802 total
```

Current version with LTM -O2, tune -O3

```text
#define MP_DEFAULT_KARATSUBA_MUL_CUTOFF 162 <
#define MP_DEFAULT_KARATSUBA_SQR_CUTOFF 128
#define MP_DEFAULT_TOOM_MUL_CUTOFF      191 <
#define MP_DEFAULT_TOOM_SQR_CUTOFF      202

% time     seconds  usecs/call     calls      function
------ ----------- ----------- --------- --------------------
 33.86    0.209098          66      3145 realloc
 32.20    0.198796          64      3072 calloc
 32.18    0.198707          64      3072 free
  1.49    0.009174          63       144 clock_gettime
  0.08    0.000482         120         4 __printf_chk
  0.06    0.000355          88         4 __errno_location
  0.05    0.000295          73         4 strtol
  0.02    0.000102          17         6 SYS_brk
  0.01    0.000087          21         4 SYS_write
  0.01    0.000080          16         5 SYS_mmap
  0.01    0.000067          16         4 SYS_mprotect
  0.01    0.000049          16         3 SYS_access
  0.01    0.000043          14         3 SYS_fstat
  0.01    0.000033          16         2 SYS_openat
  0.01    0.000033          16         2 SYS_close
  0.00    0.000023          23         1 SYS_munmap
  0.00    0.000014          14         1 SYS_read
  0.00    0.000013          13         1 SYS_arch_prctl
------ ----------- ----------- --------- --------------------
100.00    0.617451                  9477 total
```
Same with  -O2 (both)

```text
#define MP_DEFAULT_KARATSUBA_MUL_CUTOFF 156 <
#define MP_DEFAULT_KARATSUBA_SQR_CUTOFF 128
#define MP_DEFAULT_TOOM_MUL_CUTOFF      193 <
#define MP_DEFAULT_TOOM_SQR_CUTOFF      195

% time     seconds  usecs/call     calls      function
------ ----------- ----------- --------- --------------------
 33.26    0.207328          65      3145 realloc
 32.57    0.203006          66      3072 calloc
 32.40    0.201967          65      3072 free
  1.48    0.009204          63       144 clock_gettime
  0.08    0.000483         120         4 __printf_chk
  0.06    0.000377          94         4 __errno_location
  0.05    0.000306          76         4 strtol
  0.02    0.000109          18         6 SYS_brk
  0.02    0.000101          20         5 SYS_mmap
  0.01    0.000083          20         4 SYS_mprotect
  0.01    0.000077          19         4 SYS_write
  0.01    0.000062          20         3 SYS_access
  0.01    0.000049          16         3 SYS_fstat
  0.01    0.000044          22         2 SYS_openat
  0.01    0.000032          16         2 SYS_close
  0.00    0.000028          28         1 SYS_munmap
  0.00    0.000017          17         1 SYS_read
  0.00    0.000017          17         1 SYS_arch_prctl
------ ----------- ----------- --------- --------------------
100.00    0.623290                  9477 total
```
No difference between [LTM -O2, tune -O3] and both -O2, all well inside the error-bars. Would have made me wonder if not but now I have it that `tune` itself does not influence the result significantly.

The most significant outlier is TC3sqr within the very first measurement, so I'll start from there.

minad:
> The Toom-Cook algorithms do not need a lot of small  allocation, the numbers are quite large, so your 5% is probably all you  can get without a specialized memory management. GlibC's one got quite good over the last 10 years, I would have guessed  that you would gain much more with a speed optimized malloc  implementation.

I don't understand what you are saying here. I just tested somehow specialised memory managment and I tested jemalloc (probably the best allocator around) and got between 5% to 10% speedup. At some point it is not possible to win more since what is expensive is not the allocation but touching the memory (stalling the cpu and in the worst case mapping pages which have not been touched before).

And interestingly this is aligned with what you are measuring. There is the "memset" call which costs a lot of time and for some reason it does not appear anymore in the last version. We are zeroing the upper digits quite often and this could be a problem. At least some of the memsetting can be disabled now by disabling #255.

czurnieden:
>  I just tested somehow specialised memory managment

With "specialized" I meant an ingrained MM, carefully tailored to LTM where LTM can even give feedback. Feedback would be useful in the example of Toom-Cook where the (max) size of the variables is known in advance and they could tell the MM to keep their memory until the end, overwrite instead of doing deep copies et cetera p. p. And the other way around, too, rewrite LTM  to make the MM's life easier. I would expect a gain of 30% and more in the case of Toom-Cook.
Having full control over the MM also means that you can tailor the size of the buckets, which can give the many-small-allocs use case a good kick. You can also get rid of all the code for the safety-features and edge-cases you need to implement when you do not know your input (more chances for the unsuspecting user to get holes in their feet ;-) ).

There are most likely a lot of even sillier things possible but I'm not as up-to-date with the current state re MM as you are, it is a long time ago (10, 15 years?) that I had to write my own malloc because what was available at that time was not usable for my application.

> I tested jemalloc (probably the best allocator around) 

Now that I had the time to look it up, there are others that claim to be better ;-)

But all of these benchmarks you can find with Google's help only answer the question "Better for what?" and that's the crux of the biscuit here.
As a general purpose allocator on the other side jemalloc is really good!

> There is the "memset" call which costs a lot of time and for some reason it does not appear anymore in the last version.

It is the `-O3` optimization that uses `memset`, `-O2` does not. I should have put some lines in between the listings, sorry.

`lstrace` is not the most verbose but others, e.g.: valgrind's `massif` honour their name well and have a massive amount of  output, way too much for a post here.

> We are zeroing the upper digits quite often and this could be a problem.

Yeah, but LTM is used for cryptography and security does not come without some cost. We can't allow data to linger in memory for longer than absolutely necessary so that's not an option.
Or make it an option?

> But if I recall correctly, I never saw any such outliers as you do.

That is a brutal one and I'm pretty sure you would recall it ;-)

The rest is all explainable, the amount of changes in speed are expected and I can even explain the difference between `init_size` and `mp_grow` with `-O3` now (it is the time needed for the change from the smallest bucket to an extra large one for the latter) but that outlier?  More than 300% slower?
It could even be a bug in the optimizer (Outlier with `-O3`, normal with `-O2`) but I tested with a couple GCCs and clangs.
Haven't tried the Intel compiler yet, will give it a chance tonight to check that assumption.

I could spend a whole day testing everything and the kitchen-sink but I think it's time now for the assembler-dump.
\*sigh\*

minad:
>  With "specialized" I meant an ingrained MM, carefully tailored to LTM where LTM can even give feedback. Feedback would be useful in the example of Toom-Cook where the (max) size of the variables is known in advance and they could tell the MM to keep their memory until the end, overwrite instead of doing deep copies et cetera p. p. And the other way around, too, rewrite LTM to make the MM's life easier. I would expect a gain of 30% and more in the case of Toom-Cook.

What one could try is some kind of region allocator which just reserves a region in the beginning of some bigger operation and never frees. Everything is freed in the end of the operation. This is the only optimization I can imagine which could give an advantage.

But instead of optimizing allocation behavior the lowest hanging fruit I am seeing is reducing the amount of MP_ZERO_DIGITSs.

> Having full control over the MM also means that you can tailor the size of the buckets, which can give the many-small-allocs use case a good kick. You can also get rid of all the code for the safety-features and edge-cases you need to implement when you do not know your input (more chances for the unsuspecting user to get holes in their feet ;-) ).

I am pretty sure, this won't help us much. The mallocs usually already use fine grained buckets and you might even be able to tune them via some settings. The only thing you could win is inlining some kind of fast path code, accessing the bucket if there is something in it and in the slow path calling into the allocator. I think the linux kernel slab allocator does something like this.

>  Now that I had the time to look it up, there are others that claim to be better ;-) But all of these benchmarks you can find with Google's help only answer the question "Better for what?" and that's the crux of the biscuit here. As a general purpose allocator on the other side jemalloc is really good!

Sure it always depends on the use case. But according to my recent experience these general purpose allocators are already so good in selecting and returing a chunk of memory that accessing the memory is the far more expensive operation. But good allocators ensure that you allocate recently used chunks first since they are already hot in the cache etc.

Where these allocators usually differ is in their fragmentation, long runtime behavior and if there are many threads etc. Not relevant to our microbenchmarks ;)

> Yeah, but LTM is used for cryptography and security does not come without some cost. We can't allow data to linger in memory for longer than absolutely necessary so that's not an option.
Or make it an option?

We could at least make it an option. It is like a 5% cost which is basically added to every fundamental operation. From the crypto/mitigation point of view, I think it would also be ok to just let the upper digits uninitialized or let them contain garbage, since the mp_ints probably live only for a short period of time and the lower digits still contain the probably much more valuable data. What is important is zeroing the memory before free (and this is what we are doing).

But if we would want to make it configurable - it is a bit hard. I tried it but I am not familar with how most algorithms work. I disabled some ZERO_DIGITS, checked valgrind for uninitialized accesses, looked a bit around the control flow, while trying not to break the testsuite.

> It is the -O3 optimization that uses memset, -O2 does not. I should have put some lines in between the listings, sorry.

Maybe enable MP_USE_MEMSET for benchmarking? This makes it clear where time is just eaten up by zeroing. For this reason I introduced this option.

> I could spend a whole day testing everything and the kitchen-sink but I think it's time now for the assembler-dump.

If you are digging around further, give perf a try :)
This subsumes lstrace/strace and per machine instruction profiling. You can even look at the call stack inside the kernel...

czurnieden:
> I am pretty sure, this won't help us much.

The list was just an example of what can be done if one is willing to go to the extreme, not as suggestions for LTM ;-)

> I think it would also be ok to just [&hellip;]

No, LTM is not only meant to, it actually gets used in cryptographic programs, so no fiddling here, at least not with the default behaviour. The cost involved for that kind of security is normally accepted as unavoidable if it is not too much and 5-10% is  "not too much" for most people. (Thanks to MS? ;-) )

It is tempting, yes, but I wouldn't touch cryptographic code without a **very** good reason.

We can only make it optional and offer some kind of `make non-crypto` or something in that direction but as you rightfully noted:

> it is a bit hard

Yeah, a lot of work.

> Maybe enable MP_USE_MEMSET for benchmarking?

For the autotuner only:

```text
No zeroing of buffers and digits

MP_DEFAULT_KARATSUBA_MUL_CUTOFF 92
MP_DEFAULT_KARATSUBA_SQR_CUTOFF 128
MP_DEFAULT_TOOM_MUL_CUTOFF      151 <--
MP_DEFAULT_TOOM_SQR_CUTOFF      200 <--

=================================================

No memset(1) for the zeroing of buffers and digits

MP_DEFAULT_KARATSUBA_MUL_CUTOFF 95
MP_DEFAULT_KARATSUBA_SQR_CUTOFF 127
MP_DEFAULT_TOOM_MUL_CUTOFF      167
MP_DEFAULT_TOOM_SQR_CUTOFF      215

=================================================

With memset(1)  for the zeroing of buffers and digits

MP_DEFAULT_KARATSUBA_MUL_CUTOFF 98
MP_DEFAULT_KARATSUBA_SQR_CUTOFF 128
MP_DEFAULT_TOOM_MUL_CUTOFF      165
MP_DEFAULT_TOOM_SQR_CUTOFF      247 <--
```

For the testsuite (run 10 times)

```text
make test_standalone
time for i in {1..100};do ./test;done

real	33m41,921s
user	33m40,540s
sys	0m0,668s

=================================================

CFLAGS=" -DMP_USE_MEMSET " make test_standalone
time for i in {1..100};do ./test;done

real	33m48,759s
user	33m47,180s
sys	0m0,780s
```
A bit worse with system memset if the numbers are very large, but drops into insignificance if we assume that the testsuite comes closer to normal use than the autotune. You have the overhead of the function call and the checks&balances but memset is also highly optimized.
Did you have different results?

minad:
> No, LTM is not only meant to, it actually gets used in cryptographic programs, so no fiddling here, at least not with the default behaviour. The cost involved for that kind of security is normally accepted as unavoidable if it is not too much and 5-10% is "not too much" for most people. (Thanks to MS? ;-) )
It is tempting, yes, but I wouldn't touch cryptographic code without a very good reason.
We can only make it optional and offer some kind of make non-crypto or something in that direction but as you rightfully noted:

For sure I would only make it optional. However I would actually like to know what the reason is to do this zeroing. What kind of attacks are prevented by doing that? If an attacker has the possibility to read leaked data from the upper digits, I assume the probability is high that they also have the possibility to read leaked data from the lower digits.

The usual mitigation of zeroing before free (or even that the hardened malloc zeros everything always) is in place in order to prevent leaking data from a critical cryptographical and hopefully safe part of an application to an potentially unsafe part of the application. Hardened allocators randomize the allocated addresses or segregate allocations in some kind of pools in order to prevent leaking data from one part to the other.

Zeroing the memory reduces the time in which the critical data is around. This is helpful if the data could leak in other ways (swapping or some more subtle channels?). But in the lower digits we still have the finished computation lying around, so this argument seems kind of moot. And because of swapping you should use locked memory for example. We are also not doing that by default.

For all these reasons I concluded that zeroing the upper digits is kind of useless from a security pov in contrast to zeroing before free.

> A bit worse with system memset if the numbers are very large, but drops into insignificance if we assume that the testsuite comes closer to normal use than the autotune. You have the overhead of the function call and the checks&balances but memset is also highly optimized.
Did you have different results?

I don't think there will be a big difference. For small memsets we lose, for big ones we win using memset. But during profiling, using a memset call helps to see the associated costs. Otherwise the cost will just be assigned to the function doing the memset manually. This is what I meant. However as we've seen before at O3 loops are replaced by memset and vice versa (or by inline rep stosb or sth).

czurnieden:
> However I would actually like to know what the reason is to do this zeroing.

Because we have absolutely no information about, safe influence on the environment, we have to assume the worst (within reasonable limits, of course)

> But in the lower digits we still have the finished computation lying around, so this argument seems kind of moot

If you have an operation that tries to remove information, parts of that information can still reside in the upper digits.

Oversimplified example: 123456789 % 12345701 = 12345480 (overwriting: `rem(a,b,a)`) but the last 9 is still in memory and if that was a key or has information about the key you already got one ninth of it.
Yes, it is way more complicated in real life, I know.

> But during profiling, using a memset call helps to see the associated costs.

Ah, now I see what you meant.

minad:
> If you have an operation that tries to remove information, parts of that information can still reside in the upper digits.

Yes, probably this is a strong enough argument. But given the other facts about memory allocation, unlocked memory it might still not be worth it. But as I said, if we would remove zero_digits, we would make it optional so or so. But I won't do it any time soon since I don't think it is a pressing issue and the potential for breaking stuff is to great.

PR #266, opened 2019-05-17
more linting

fperrad:


tommath.h
```diff
@@ -229,7 +229,7 @@ typedef struct  {
 } mp_int;
 
 /* callback for mp_prime_random, should fill dst with random bytes and return how many read [upto len] */
-typedef int private_mp_prime_callback(unsigned char *dst, int len, void *dat);
+typedef int private_mp_prime_callback(unsigned char *dst, int len, const void *dat);
```
minad: This is wrong since it changes the signature of the public (but deprecated) function type ltm_prime_callback.

tommath_private.h
```diff
@@ -139,7 +139,7 @@ extern void MP_FREE(void *mem, size_t size);
 #define MP_IS_EVEN(a) (((a)->used == 0) || (((a)->dp[0] & 1u) == 0u))
 #define MP_IS_ODD(a)  (((a)->used > 0) && (((a)->dp[0] & 1u) == 1u))
 
-#define MP_SIZEOF_BITS(type)    (CHAR_BIT * sizeof(type))
+#define MP_SIZEOF_BITS(type)    ((size_t)CHAR_BIT * sizeof(type))
```
minad: Are these casts necessary? I would prefer if you only add casts where they are strictly necessary. Otherwise they just obfuscate the code.

tommath_private.h
```diff
@@ -69,8 +69,8 @@ do {                                                    \
 do {                                                    \
    size_t zs_ = (size);                                 \
    char* zm_ = (char*)(mem);                            \
-   while (zs_-- > 0) {                                  \
-      *zm_++ = 0;                                       \
+   while (zs_-- > 0u) {                                 \
+      *zm_++ = (char)0;                                 \
```
minad: Unnecessary cast.

tommath.h
fperrad: so, this commit must wait for the removal of `ltm_prime_callback` which is currently only deprecated.

minad: No. The callback will go away completely in the end.

fperrad:
commit concerning `ltm_prime_callback` was removed

fperrad:
updated + refactored

tommath_private.h
sjaeckel: @minad fine like that?

minad: Not a fan of the cast and I don't think it is necessary, but not really worth discussing...

minad:
@sjaeckel This one is ready too btw. It was ready before #274, which you merged earlier today ;)

fperrad:
rebased

PR #267, opened 2019-05-17
make mp_word private

minad:
mp_word is an internal type and it is problematic if it is exposed
in the public api. See for example #216 - MSVC does not support 128 bit
mp_words. But it is perfectly ok to use those internally in the library,
as long as the library is compiled with GCC.

Ping @nijtmans - does it make sense to hide the mp_word type from the public api in general?


minad:
@sjaeckel What do we have to check here? Do you see any downsides of this?

PR #268, opened 2019-05-18
improve coverage

sjaeckel:
I've noticed that the test-case for balanced multiplication wasn't hit anymore (or probably never? :-D ).

While looking at the coverage results I added some trivial ones ~~and I think I found an inconsistency... nothing bad but it still stood out while looking at the code...~~

~~[mp_prime_rabin_miller_trials:38](https://github.com/libtom/libtommath/blob/develop/bn_mp_prime_rabin_miller_trials.c#L38) leads to 2 MR trials for numbers > 4096 bits~~

~~@czurnieden is that change fine that not only 4096bit sized numbers only have to do 1 MR but also numbers >4096bits?~~ 

moved to #269

### DIFF
--- a/demo/test.c
+++ b/demo/test.c
@@ -29,9 +29,11 @@ static unsigned long ulabs(long x)
 static int test_trivial_stuff(void)
 {
    mp_int a, b, c, d;
-   if (mp_init_multi(&a, &b, &c, &d, NULL)!= MP_OKAY) {
+   mp_err e;
+   if ((e = mp_init_multi(&a, &b, &c, &d, NULL)) != MP_OKAY) {
       return EXIT_FAILURE;
    }
+   (void)mp_error_to_string(e);
 
    /* a: 0->5 */
    mp_set_int(&a, 5uL);
@@ -87,10 +89,45 @@ static int test_trivial_stuff(void)
    mp_set_int(&c, 7uL);
    mp_exptmod(&a, &b, &c, &d);
 
+   if (mp_iseven(&a) == mp_isodd(&a)) {
+      goto LBL_ERR;
+   }
+
    mp_clear_multi(&a, &b, &c, &d, NULL);
    return EXIT_SUCCESS;
 LBL_ERR:
    mp_clear_multi(&a, &b, &c, &d, NULL);
+   return EXIT_FAILURE;
+}
+
+static int test_mp_fread_fwrite(void)
+{
+   mp_int a, b;
+   mp_err e;
+   FILE *tmp = NULL;
+   if ((e = mp_init_multi(&a, &b, NULL)) != MP_OKAY) {
+      return EXIT_FAILURE;
+   }
+
+   mp_set_int(&a, 123456uL);
+   tmp = tmpfile();
+   if ((e = mp_fwrite(&a, 64, tmp)) != MP_OKAY) {
+      goto LBL_ERR;
+   }
+   rewind(tmp);
+   if ((e = mp_fread(&b, 64, tmp)) != MP_OKAY) {
+      goto LBL_ERR;
+   }
+   if (mp_get_int(&b) != 123456uL) {
+      goto LBL_ERR;
+   }
+   fclose(tmp);
+
+   mp_clear_multi(&a, &b, NULL);
+   return EXIT_SUCCESS;
+LBL_ERR:
+   if (tmp != NULL) fclose(tmp);
+   mp_clear_multi(&a, &b, NULL);
    return EXIT_FAILURE;
 }
 
@@ -841,7 +878,7 @@ static int test_mp_prime_is_prime(void)
 {
    int ix;
    mp_err err;
-   mp_bool cnt;
+   mp_bool cnt, fu;
 
    mp_int a, b;
    if (mp_init_multi(&a, &b, NULL)!= MP_OKAY) {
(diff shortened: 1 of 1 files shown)


demo/test.c
```diff
@@ -1822,10 +1848,13 @@ static int test_s_mp_balance_mul(void)
    if ((e = mp_read_radix(&b, nb, 64)) != MP_OKAY) {
       goto LTM_ERR;
    }
+   kara = MP_KARATSUBA_MUL_CUTOFF;
+   MP_KARATSUBA_MUL_CUTOFF = b.used-2;
 
    if ((e = mp_mul(&a, &b, &c)) != MP_OKAY) {
```
minad: Could you please just call s_mp_balance_mul here (instead of changing the cutoffs)? This coverage issue you discovered is exactly what I meant, when I said the tests are not robust written like this.

And now that we have tommath_private.h included,  if we call s_mp_balance_mul directly, the issue is just gone :)

demo/test.c
```diff
@@ -87,6 +90,24 @@ static int test_trivial_stuff(void)
    mp_set_int(&c, 7uL);
    mp_exptmod(&a, &b, &c, &d);
 
+   mp_set_int(&a, 123456uL);
+   tmp = tmpfile();
+   if ((e = mp_fwrite(&a, 64, tmp)) != MP_OKAY) {
+      goto LBL_ERR;
+   }
+   rewind(tmp);
+   if ((e = mp_fread(&b, 64, tmp)) != MP_OKAY) {
```
minad: Even if these are trivial tests, could you move them to their own test_* functions please?

bn_mp_prime_rabin_miller_trials.c
```diff
@@ -35,7 +35,7 @@ int mp_prime_rabin_miller_trials(int size)
          return (x == 0) ? sizes[0].t : sizes[x - 1].t;
       }
    }
-   return sizes[x-1].t + 1;
+   return sizes[x-1].t;
```
minad: Is this is the inconsistency? I cannot judge this since I never looked into those prime tests. Ping @czurnieden?

demo/test.c
sjaeckel: > the issue is just gone

until we have proper symbol hiding activated and then we have an entirely new problem

demo/test.c
```diff
@@ -902,6 +923,10 @@ static int test_mp_prime_is_prime(void)
       if (cnt == MP_NO) {
          printf("\nsub is not prime!\n");
       }
+      mp_prime_frobenius_underwood(&b, &fu);
+      if (fu == MP_NO) {
+         printf("\nfrobenius-underwood says sub is not prime!\n");
```
minad: Uuh or are the different tests inconsistent?

sjaeckel: nope, not that I've noticed, it's just that `mp_prime_frobenius_underwood()` was never called in the default configuration

minad:
@sjaeckel Maybe split this PR in two if there is an issue with the prime tests, which needs more work? 
balance_mul and the testing stuff looks good modulo nitpicks.

demo/test.c
minad: For now we don't have that issue. I will fix the building issues when the symbols are hidden, by doing a special test library with unhidden symbols. The test suite already accesses s_functions (jenkins).

demo/test.c
sjaeckel: fine like that?

minad: Not that it matters, but there are missing fcloses.

minad: Besides that :+1:

sjaeckel: better?

sjaeckel:
split into two PR's as they're independent

czurnieden:
`mp_prime_rabin_miller_trials` returns `39` in that special case which is too much for 8-bit. The test for `t > MP_PRIME_SIZE` fails because `MP_PRIME_SIZE` is only `31` for `MP_8BIT`.

There are many places available to repair/work-around that. Any preferences?

minad:
@sjaeckel Why is this failing? Maybe move all the prime sieve stuff to #269. Then balanced mul and fread/fwrite tests can already be merged.

czurnieden:
> Why is this failing?

I tried to explain it?

minad:
@czurnieden Sorry, I didn't read your explanation first because I thought you commented here instead of in #269. I thought @sjaeckel moved all the "critical" stuff to #269.

czurnieden:
> Sorry, I didn't read your explanation first because I thought you commented here instead of in #269

I think I slowly grasp github's UI/UX but you're right, of course, to be still sceptical ;-)

But now that I have you: do you think it is enough to just delete the first test (could be done in this PR) or get rid of the second part, too (needs more work, including change of documentation, hence its own PR)?

minad:
> But now that I have you: do you think it is enough to just delete the first test (could be done in this PR) or get rid of the second part, too (needs more work, including change of documentation, hence its own PR)?

Hmm, I am not sure I really got you. Do you want to get rid of this ltm_primetab division test and replace it with your sieve? Then the specific ltm_primetab tests can go to (what you wrote under Edit II).

sjaeckel:
> I thought @sjaeckel moved all the "critical" stuff to #269.

Sorry, I thought so too :-) ... fixed now

czurnieden:
> Hmm, I am not sure I really got you.

Then let me assure you: you didn't&mdash;it seems as if I was not able to explain it with words.
So let me express myself in code.
Two PRs, one to repair the failure <del>here</del>in #269, one explaining what I meant.
(may take some time, am on an old laptop now that isn't the fastest)

minad:
>  Then let me assure you: you didn't—it seems as if I was not able to explain it with words. So let me express myself in code. Two PRs, one to repair the failure herein #269, one explaining what I meant. (may take some time, am on an old laptop now that isn't the fastest)

Thanks! That's better. I'll add a comment in #272.

PR #269, opened 2019-05-18
do 2 MR rounds for numbers >=2048bits

sjaeckel:
While looking at the coverage results in #268 I think I found an inconsistency... nothing bad but it still stood out while looking at the code...

[mp_prime_rabin_miller_trials:38](https://github.com/libtom/libtommath/blob/develop/bn_mp_prime_rabin_miller_trials.c#L38) leads to 2 MR trials for numbers > 4096 bits

@czurnieden is that change fine that not only 4096bit sized numbers only have to do 1 MR but also numbers >4096bits?

**Edit**: changed to do always at least 2 Rounds

### DIFF
--- a/bn_mp_prime_is_prime.c
+++ b/bn_mp_prime_is_prime.c
@@ -25,11 +25,6 @@ mp_err mp_prime_is_prime(const mp_int *a, int t, mp_bool *result)
    /* default to no */
    *result = MP_NO;
 
-   /* valid value of t? */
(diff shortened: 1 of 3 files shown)


czurnieden:
> is that change fine that not only 4096bit sized numbers only have to do 1 MR but also numbers >4096bits?

Yes.

But now that I think about it: that is the theoretical value. Carefully calculated but still only a theoretical value.
Could we set the minimum to (2) two to sooth my conscience?

PR #270, opened 2019-05-19
Macro fixes

minad:
Some minor things regarding macros in deprecated.c etc.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -86,3 +86,5 @@ test_*.txt
 *.asc
 *.tar.xz
 *.zip
+
+doc/pics/*.ps
--- a/bn_deprecated.c
+++ b/bn_deprecated.c
@@ -116,7 +116,7 @@ mp_err mp_toom_sqr(const mp_int *a, mp_int *b)
    return s_mp_toom_sqr(a, b);
 }
 #endif
-#ifdef BN_REVERSE_C
+#ifdef S_MP_REVERSE_C
 void bn_reverse(unsigned char *s, int len)
 {
    s_mp_reverse(s, len);
--- a/bn_deprecated.c
+++ b/bn_deprecated.c
@@ -2,14 +2,8 @@
 #ifdef BN_DEPRECATED_C
 /* LibTomMath, multiple-precision integer library -- Tom St Denis */
 /* SPDX-License-Identifier: Unlicense */
-/* LibTomMath, multiple-precision integer library -- Tom St Denis */
 
-/* SPDX-License-Identifier: Unlicense */
-#include <tommath_private.h>
 #ifdef BN_MP_GET_BIT_C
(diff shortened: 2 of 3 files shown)


minad:
@sjaeckel Do you agree with the changes here. This is miscellaneous stuff I couldn't fit anywhere else. But some of the BN_* macros in deprecated were wrong before.

doc/tommath.src
```diff
@@ -5986,8 +5949,6 @@ are congruent to one modulo four, otherwise it evaluates to negative one.
 By step nine if $a'$ does not equal one a recursion is required.  Step 9.1 computes $p' \equiv p \mbox{ (mod }a'\mbox{)}$ and will recurse to compute
 $\left ( {p' \over a'} \right )$ which is multiplied against the current Jacobi product.
 
-EXAM,bn_mp_jacobi.c
```
sjaeckel: I really don't like how the book is torn apart... we should probably remove it entirely since it won't be updated... another option would be to fix those parts, but I'm not sure if it's worth the effort.

minad: Yes, I would probably remove it and freeze it. It is a nice thing but it seems tommath diverged quite a bit already from what it was in the beginning. It is still useful as a reference.

minad: The thing is - the functions I was touching until now were quite trivial. But this more complicated number theoretic stuff @czurnieden is doing really needs docs.

minad: @sjaeckel I removed the doc changes for now.

sjaeckel: > The thing is - the functions I was touching until now were quite trivial

this applies to my changes as well...

> But this more complicated number theoretic stuff @czurnieden is doing really needs docs.

@czurnieden already updates the regular documentation... I don't know if this book should be updated as well... probably @czurnieden can answer this?

minad: I don't get what you mean by regular docs vs book. I think the book is the docs - at least this is what I touched for example in #279 too, when @czurnieden asked for docs.

sjaeckel: > You mean bn.tex vs tommath.src?

yep

minad: Let's make this compromise:

* We don't touch the book tommath.src/tex except for making sure that it compiles (Right now it does not)
* We update bn.tex

sjaeckel: I would say

* We remove the book, since a compiling, but not up-to-date book doesn't help
* We update bn.tex

I'd go then and make a build of the latest up-to-date version of the book and add that one on the libtommath page

minad: That's also ok I guess.

sjaeckel: > That's also ok I guess.

deal!

@czurnieden should be fine with this as well, as he was also updating bn.tex all the time

PR #279, opened 2019-05-20
deprecate mp_tc_(and|or|xor) in favor of mp_(and|or|xor)

minad:
deprecate mp_tc_(and|or|xor) in favor of mp_(and|or|xor)

* same behavior for positive numbers
* generalisation for negative numbers, treating them as two complement
* improve algorithm, iterate once over the digits, manually perform two complement
* simplify mp_add_d, mp_sub_d
* functions are safe in case of a==c or b==c
* renamed mp_tc_div_2d to mp_signed_rsh (signed right shift)


minad:
@czurnieden we could also deprecate mp_tc_x in favor of mp_x with changed behavior for negative mp_ints. I think that would be ok since no one uses mp_x on negative numbers.
I would also prefer that to be honest.

minad:
Another option would be to name all the bitwise functions mp_bit_x?

minad:
@czurnieden I changed the PR here - all bitwise functions are called mp_bit_x now. I think the names are much nicer like this. Also `mp_tc_div_2d` didn't make much sense as a name. It is a sign extending right shift. Furthermore I updated the docs. Do you agree with these changes?

sjaeckel:
I also ask myself if there's even one user who depends on the old functions' behavior of negative numbers... I can't imagine as that was pretty useless, so I'd say we should use the old names!

minad:
Ok good! I will change everything to the old names then. No unnecessary _bit_ in the name.

minad:
@sjaeckel @nijtmans updated to old names

PR #282, opened 2019-05-21
Restrict running Valgrind

minad:
* Activate on the develop branch
* Activate on branches containing the word v-algrind (without hyphen)
* Activate if the commit message contains v-algrind (without hyphen)

With valgrind the tests take about 15 minutes, without valgrind about 1 minute. We should not do that. I think it is bad for us and it is kind of abusive towards travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,6 +60,13 @@ matrix:
     # We have only one program and the variable $BUILDOPTIONS
     # has only the options to that program: testme.sh
 
+    # Run always with valgrind (no sanitizer, but debug info)
+    - env: COMPILE_DEBUG=1 BUILDOPTIONS='--with-cc=gcc-4.9 --with-m64 --with-valgrind'
+      addons:
+        apt:
+          packages:
+            - gcc-4.9
+
     # Check source code format
     - env: BUILDOPTIONS='--format'
       addons:
@@ -94,7 +101,7 @@ matrix:
     - env: SANITIZER=1 BUILDOPTIONS='--with-cc=clang-7 --with-travis-valgrind --make-option=tune'
 
     # clang for x86-64 architecture (64-bit longs and 64-bit pointers)
-    - env: COMPILE_DEBUG=1 SANITIZER=1 CONV_WARNINGS=relaxed BUILDOPTIONS='--with-cc=clang-7 --with-m64 --with-valgrind'
+    - env: SANITIZER=1 CONV_WARNINGS=relaxed BUILDOPTIONS='--with-cc=clang-7 --with-m64 --with-travis-valgrind'
     - env: SANITIZER=1 CONV_WARNINGS=strict BUILDOPTIONS='--with-cc=clang-7 --with-m64 --with-travis-valgrind'
     - env: SANITIZER=1 BUILDOPTIONS='--with-cc=clang-6.0 --with-m64 --with-travis-valgrind'
       addons:
@@ -134,11 +141,6 @@ matrix:
         apt:
           packages:
             - gcc-4.8
-    - env: BUILDOPTIONS='--with-cc=gcc-4.9 --with-m64 --with-travis-valgrind'
-      addons:
-        apt:
-          packages:
-            - gcc-4.9
 
     # GCC for the x86-64 architecture with restricted limb sizes
     # formerly started with the option "--with-low-mp" to testme.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,6 +83,35 @@ matrix:
             - libc6-dev-i386
             - gcc-multilib
 
+    # Test "autotuning", the automatic evaluation and setting of the Toom-Cook cut-offs.
+    #- env: SANITIZER=1 BUILDOPTIONS='--with-cc=gcc-5 --cflags=-DMP_8BIT  --with-travis-valgrind --make-option=tune'
+    #- env: SANITIZER=1 BUILDOPTIONS='--with-cc=gcc-5 --cflags=-DMP_16BIT --with-travis-valgrind --make-option=tune'
+    #- env: SANITIZER=1 BUILDOPTIONS='--with-cc=gcc-5 --cflags=-DMP_32BIT --with-travis-valgrind --make-option=tune'
+    #- env: SANITIZER=1 BUILDOPTIONS='--with-cc=gcc-5 --with-travis-valgrind --make-option=tune'
+    #- env: SANITIZER=1 BUILDOPTIONS='--with-cc=clang-7 --cflags=-DMP_8BIT  --with-travis-valgrind --make-option=tune'
+    #- env: SANITIZER=1 BUILDOPTIONS='--with-cc=clang-7 --cflags=-DMP_16BIT --with-travis-valgrind --make-option=tune'
+    #- env: SANITIZER=1 BUILDOPTIONS='--with-cc=clang-7 --cflags=-DMP_32BIT --with-travis-valgrind --make-option=tune'
+    - env: SANITIZER=1 BUILDOPTIONS='--with-cc=clang-7 --with-travis-valgrind --make-option=tune'
+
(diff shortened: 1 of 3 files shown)


minad:
@sjaeckel I added a --travis-valgrind option which has the travis specific behavior. Please do us a favor and merge this, update all PRs on top of develop and be done with all of it in two minutes ;)

minad:
Ah, I see you already approved. But I think you might like --with-travis-valgrind better.

sjaeckel:
> I think you might like --with-travis-valgrind better

indeed I do

> Please do us a favor and merge this, update all PRs on top of develop and be done with all of it in two minutes

that's why I already killed all other PR's running besides that run that is nearly finished ;-)
I'll merge then that one, rebase this and we'll see how long it will take :)

minad:
@sjaeckel For some reason it is still running valgrind. Travis says "branch develop" but this is a PR. Let me check.

minad:
Urks. TRAVIS_BRANCH: for builds triggered by a pull request this is the name of the branch targeted by the pull request.

Fix incoming...

minad:
@sjaeckel Works now. We will see if it still valgrinds on develop after you merged it ;)

sjaeckel:
We could probably add a single valgrind build as one of the first so we're sure that at least the default config is valgrinded before merging a PR!?

minad:
I also thought about it and I think it is unnecessary. But if you want it no problem, only one is much more acceptable imho.

sjaeckel:
> I think it is unnecessary

I think it won't be needed very often, but I'd prefer to have for those few cases

> only one is much more acceptable imho

I also think so! that's how we do it for libtomcrypt as well, only the default build gets valgrinded

minad:
Ah I rejected the idea since we would need valgrind on MP_8BIT MP_16BIT etc to be safe. And then it is again open ended.

minad:
the most restrictive build is with CONV_WARNINGS=strict but this one also enables enum which is not yet the default ABI

sjaeckel:
> we would need valgrind on MP_8BIT MP_16BIT etc to be safe

I don't think that's required! 
1. especially since 8 and 16bit shouldn't be treated as "production use" 
2. there's not that much different code for those configurations

.travis.yml
```diff
@@ -94,37 +94,37 @@ matrix:
...
-    - env: BUILDOPTIONS='--with-cc=gcc-4.9 --with-m64 --with-valgrind'
+    - env: BUILDOPTIONS='--with-cc=gcc-4.9 --with-m64 --with-travis-valgrind'
       addons:
         apt:
           packages:
             - gcc-4.9
 
     # clang for x86-64 architecture (64-bit longs and 64-bit pointers)
-    - env: SANITIZER=1 CONV_WARNINGS=strict BUILDOPTIONS='--with-cc=clang-7 --with-m64 --with-valgrind'
     - env: SANITIZER=1 CONV_WARNINGS=relaxed BUILDOPTIONS='--with-cc=clang-7 --with-m64 --with-valgrind'
```
sjaeckel: :+1:

minad:
@sjaeckel Could we run the tuner only for 64 bit and not for all the cases? It also needs some time. We should run it at least once to make sure that nothing breaks by accident. Otherwise the tuner is probably low churn and we don't have to test in in all configs.

sjaeckel:
> Could we run the tuner only for 64 bit and not for all the cases?

perfectly fine by me!

sjaeckel:
btw. I've just discovered https://travis-ci.org/libtom/libtommath/jobs/535171987 by hazard... should we have error'ed on this? I can't reproduce it

minad:
You mean the malloc reachable memory? At first this is not really a bug and it happens in the dynamic loader. I've seen such things quite often, ld.so and dlopen often leaks memory in some way but this has nothing to do with us.

sjaeckel:
> We should probably compile with debug data enable.

good idea for the valgrind build!

czurnieden:
> especially since 8 and 16bit shouldn't be treated as "production use"

It needs to treated as such as long as it is in LTM.

And the 16-bit version is quite usable with a lot of 16-bit MCUs, says Mouser.

The 8-bit MCUs on the other side have rarely enough memory available for even the most reduced version of LTM. Only some of the 8-bit "classicals" come close.

> there's not that much different code for those configurations

There is no different code for `MP_32BIT` (only setting the types for mp_digit/mp_word) 

There is no different code for `MP_16BIT` (only testing/setting the types) with the exception of `bn_mp_montgomery_setup.c` which is just some math and can be safely assumed as working after all these years of testing.

There **is** different code for `MP_8BIT` and some of it is in cryptographically relevant code, so it still needs testing of function and memory use.
Or get rid of the 8-bit branch completely.
I wouldn't shed a single tear if you would, it's a real pain in the behind.

And honestly: such large test-suites are not really useful for the average commit. One compiler with all hardening options switched on, (with and) without memory tests for 8(, 16, 32,) and 64 bit (and `--format` ;-) ) is sufficient for daily use.

Maybe another option like the one switching memory-tests on/off  to switch the whole test for that environment on/off?

minad:
@czurnieden 

> And honestly: such large test-suites are not really useful for the average commit. One compiler with all hardening options switched on, (with and) without memory tests for 8(, 16, 32,) and 64 bit (and --format ;-) ) is sufficient for daily use.

I think what we have here is a good compromise. Average PRs are valgrinded only once and develop is fully valgrinded. I like it like this!

sjaeckel:
>  I like it like this!

I like it as well

and now that this single valgrind build takes longer than all the other builds together we can also leave all the others enabled :D

minad:
> and now that this single valgrind build takes longer than all the other builds together we can also leave all the others enabled :D

Except for the tuner. It is too slow, @czurnieden! :D

sjaeckel:
> Except for the tuner. It is too slow, @czurnieden! :D

I can live with the speed of the tuner :-) as long as it's faster than the valgrind build it's fine ;-)

minad:
@sjaeckel COMPILE_DEBUG needlessly disables optimizations. This was written in former times when -O2/O3 was not compatible with -g. I have to change this. The valgrind build takes 30 minutes now :angry:

minad:
@sjaeckel Now it works, but the sanitized clang build is probably not the best idea for valgrind - instead I activate valgrind on the gcc 4.9 build without sanitizer and run it first. Do you agree? The sanitized clang build takes around 25 minutes, the gcc 4.9 build 17 minutes.

If we want to optimize more we could also disable installing valgrind everytime etc. But this is probably not worth it. We got the long hanging fruits now.

sjaeckel:
> We got the long hanging fruits now

:banana:

minad:
Now I am done. I won't touch this anymore. I promise.

czurnieden:
> Except for the tuner. It is too slow, @czurnieden! :D

With the new #265 

```text
czurnieden ~/GITHUB/libtommath/etc  X (faster_toom_cook_three)$ time make tune
cc -Wall -W -Wextra -Wshadow -O3 -I../   -c -o tune.o tune.c
cc -Wall -W -Wextra -Wshadow -O3 -I../ tune.o ../libtommath.a -o tune
./tune_it.sh
You might like to watch the numbers go up to 100 but it will take a long time!
100Writing cut-off values to "../tommath_cutoffs.h".
In case of failure: a copy of "../tommath_cutoffs.h" is in "../tommath_cutoffs.h.orig"
'../tommath_cutoffs.h' -> '../tommath_cutoffs.h.orig'
#define MP_DEFAULT_KARATSUBA_MUL_CUTOFF 94
#define MP_DEFAULT_KARATSUBA_SQR_CUTOFF 127
#define MP_DEFAULT_TOOM_MUL_CUTOFF      165
#define MP_DEFAULT_TOOM_SQR_CUTOFF      216

real	0m13,760s
user	0m13,432s
sys	0m0,056s
```
Better? ;-)

PR #283, opened 2019-05-21
remove callgraph.txt from repository

czurnieden:
In the times of free good IDEs it is not needed anymore.
If you still want it run `./dep.pl --print-calgraph`


sjaeckel:
please also put it into gitignore

minad:
Why change anything in the scripts? Just put it in the gitignore!

sjaeckel:
> Why change anything in the scripts? Just put it in the gitignore! But I don't mind...

that's true :D 

delete it and put it in gitignore is enough

PR #273, opened 2019-05-19
prime_is_prime: remove obsolete restriction on PRIME_SIZE

czurnieden:
Removed the non-deterministic M-R-only tests.

PR #285, opened 2019-05-21
introduce mp_set_sint and mp_set_uint functions with precise types

minad:
This is a bit of a compromise now.

* The API is biased toward 32 bit (mp_get_sint, mp_get_uint, ...).
* 64 bit variants are provided (mp_get_sint64)
* Unsigned getters return two complement for negative numbers
* There are an unsigned getters for the magnitudes only (mp_get_mag, mp_get_mag64) - these are just like the current mp_get_long mp_get_long_long
* There are aliases for long and unsigned long (mp_get_ulong, mp_get_slong, ...) since long is problematic if porting between 32/64 bit platforms. Furthermore since the current API uses long, this would make things easier. But maybe these should also go to not again introduce legacy cruft.
* mp_set and mp_init_set are NOT deprecated in order to not penalize programs on platforms with int < 32 bit. This is still the most efficient function.

Why are these changes needed?

* signed setters/getters are a new addition
* unsigned setters return two complement such that round tripping and casting works seemless with c integers
* the new setters never fail, the current setters return mp_err
* the new functions use precise types (the api is robust under platform changes)

~~~ c

/**** NEW ****/
/* get integer, set integer and init with integer (int32_t) */
int32_t mp_get_sint(const mp_int *a) MP_WUR;
void mp_set_sint(mp_int *a, int32_t b);
mp_err mp_init_sint(mp_int *a, int32_t b) MP_WUR;

/**** NEW ****/
/* get integer, set integer and init with integer (int64_t) */
int64_t mp_get_sint64(const mp_int *a) MP_WUR;
void mp_set_sint64(mp_int *a, int64_t b);

/**** CHANGED API: More precise replacements for mp_set_int, mp_set_long, mp_get_int, two complement ****/
/* get integer, set integer and init with integer, behaves like two complement for negative numbers (uint32_t) */
#define mp_get_uint(a) ((uint32_t)mp_get_sint(a))
void mp_set_uint(mp_int *a, uint32_t b);
mp_err mp_init_uint(mp_int *a, uint32_t b) MP_WUR;

/**** CHANGED API: More precise replacements for mp_set_int, mp_set_long, mp_get_int, two complement ****/
/* get integer, set integer and init with integer, behaves like two complement for negative numbers (uint64_t) */
#define mp_get_uint64(a) ((uint64_t)mp_get_sint64(a))
void mp_set_uint64(mp_int *a, uint64_t b);

/***** CHANGED API: More precise replacement for mp_get_long/long_long ****/
/* get magnitude */
uint32_t mp_get_mag(const mp_int *a) MP_WUR;
uint64_t mp_get_mag64(const mp_int *a) MP_WUR;

/**** COMPATIBILITY API, needed? ****/
/* get integer, set integer (long) */
#define mp_get_slong(a)        (sizeof (long) == 8 ? (long)mp_get_sint64(a) : (long)mp_get_sint(a))
#define mp_set_slong(a, b)     (sizeof (long) == 8 ? mp_set_sint64((a), (b)) : mp_set_sint((a), (int32_t)(b)))

/* get integer, set integer (unsigned long) */
#define mp_get_ulong(a)       (sizeof (long) == 8 ? (unsigned long)mp_get_uint64(a) : (unsigned long)mp_get_ulong(a))
#define mp_get_magl(a)     (sizeof (long) == 8 ? (unsigned long)mp_get_mag64(a) : (unsigned long)mp_get_mag(a))
#define mp_set_ulong(a, b)    (sizeof (long) == 8 ? mp_set_uint64((a), (b)) : mp_set_uint((a), (uint32_t)(b)))

/* set to single unsigned digit, only 8 bit guaranteed */
void mp_set(mp_int *a, mp_digit b);
mp_err mp_init_set(mp_int *a, mp_digit b) MP_WUR;

/* get integer, set integer and init with integer (deprecated) */
MP_DEPRECATED(mp_get_mag/mp_get_uint) unsigned long mp_get_int(const mp_int *a) MP_WUR;
MP_DEPRECATED(mp_get_magl/mp_get_ulong) unsigned long mp_get_long(const mp_int *a) MP_WUR;
MP_DEPRECATED(mp_get_mag64/mp_get_uint64) unsigned long long mp_get_long_long(const mp_int *a) MP_WUR;
MP_DEPRECATED(mp_set_uint) mp_err mp_set_int(mp_int *a, unsigned long b);
MP_DEPRECATED(mp_set_ulong) mp_err mp_set_long(mp_int *a, unsigned long b);
MP_DEPRECATED(mp_set_uint64) mp_err mp_set_long_long(mp_int *a, unsigned long long b);
MP_DEPRECATED(mp_init_uint) mp_err mp_init_set_int(mp_int *a, unsigned long b) MP_WUR;
~~~


minad:
Tests are also missing. Maybe I have time later today to fix those things.

minad:
@czurnieden Updated docs + tests

doc/bn.tex
```diff
@@ -820,28 +820,32 @@ \subsection{Single Digit}
 \}
 \end{alltt} \end{small}
 
-\subsection{Long Constants}
+\subsection{Int32 and Int64 Constants}
 
-To set a constant that is the size of an ISO C ``unsigned long'' and larger than a single digit the following function
-can be used.
+To set a constant with 32 or 64 bits the following function can be used.
```
czurnieden: "the following functions" (plural)

I know I am guilty of it myself, but these type of constructs are a bit awkward. Keep it simple, e.g.:
"The following functions set a constant with a size of 32 or 64 bits."
Or a bit more open ended with:
"The functions to set a constant with a size of 32 or 64 bits are:"

doc/bn.tex
```diff
@@ -820,28 +820,32 @@ \subsection{Single Digit}
...
+void mp_set_sint64 (mp_int * a, int64_t b);
+void mp_set_uint64 (mp_int * a, uint64_t b);
 \end{alltt}
 
-This will assign the value of the 32-bit variable $b$ to the mp\_int $a$.  Unlike mp\_set() this function will always
-accept a 32-bit input regardless of the size of a single digit.  However, since the value may span several digits
-this function can fail if it runs out of heap memory.
-
-To get the ``unsigned long'' copy of an mp\_int the following function can be used.
+This will assign the value of the variable $b$ to the mp\_int $a$. The unsigned variants set a positive value, the signed variants set the sign. The value can be obtained again by calling the following function.
```
czurnieden: I would keep it shorter, just: "This will assign the sign and value of the input \texttt{b} to \texttt{mp\_int a}."

"the following functions" (the plural "s" has been cut of in your text again)

doc/bn.tex
```diff
@@ -820,28 +820,32 @@ \subsection{Single Digit}
...
+int32_t mp_get_sint (mp_int * a);
+uint32_t mp_get_uint (mp_int * a);
+uint32_t mp_get_mag (mp_int * a);
+int64_t mp_get_sint64 (mp_int * a);
+uint64_t mp_get_uint64 (mp_int * a);
+uint64_t mp_get_mag64 (mp_int * a);
 \end{alltt}
 
-This will return the 32 least significant bits of the mp\_int $a$.
+This will return either 32 or 64 least significant bits of the mp\_int $a$. The unsigned functions
```
czurnieden: There is more than one function listed and the return only one specific type, not several.

"These functions return the 32 and 64 least significant bits of \texttt{mp\_int a} respectively."

I used "\texttt{}" because it is code but it is quite a mess in the whole documentation, we should standardize it a bit.
But that is work and on top of it: it is documentation! I know ;-)

doc/bn.tex
```diff
@@ -820,28 +820,32 @@ \subsection{Single Digit}
...
+return negative values in two complement representation. If you only want to access the magnitude, use the mp\_get\_mag functions.
```
czurnieden: "in a two complement representation" or "in a twos complement representation". I would prefer the latter as does [Google](https://books.google.com/ngrams/graph?content=twos+complement%2Ctwos-complement&case_insensitive=on&year_start=1800&year_end=2000&corpus=15&smoothing=3&share=&direct_url=t4%3B%2Ctwos%20complement%3B%2Cc0%3B%2Cs0%3B%3Btwos%20complement%3B%2Cc0%3B%3BTwos%20complement%3B%2Cc0%3B%3BTwos%20Complement%3B%2Cc0%3B%3BTWOS%20COMPLEMENT%3B%2Cc0%3B.t4%3B%2Ctwos%20-%20complement%3B%2Cc0%3B%2Cs0%3B%3Btwos%20-%20complement%3B%2Cc0%3B%3BTwos%20-%20complement%3B%2Cc0%3B%3BTwos%20-%20Complement%3B%2Cc0).

The *mag* functions return the magnitude which is the absolute value. Don't forget that documentation get's searched and glimpsed over quickly  and hence should contain more keywords than actually necessary.
E.g.:
"The \texttt{mp\_get\_mag} functions return the magnitude, the absolute value of the input."

doc/bn.tex
minad: fixed

doc/bn.tex
minad: fixed

doc/bn.tex
minad: done

doc/bn.tex
minad: fixed

doc/bn.tex
```diff
@@ -875,46 +880,31 @@ \subsection{Long Constants}
...
-
-\index{mp\_set\_long\_long}
-\begin{alltt}
-int mp_set_long_long (mp_int * a, unsigned long long b);
-\end{alltt}
-
-This will assign the value of the 64-bit variable $b$ to the mp\_int $a$.
-
-To get the ``unsigned long long'' copy of an mp\_int the following function can be used.
+To retrieve the value, the following function can be used.
```
czurnieden: "the following functions", plural, you made two out of The One That Was Before.
Didn't see it the first time, sorry.

doc/bn.tex
```diff
@@ -875,46 +880,31 @@ \subsection{Long Constants}
...
-\index{mp\_get\_long\_long}
+\index{mp\_get\_ulong}
 \begin{alltt}
-unsigned long long mp_get_long_long (mp_int * a);
+long mp_get_slong (mp_int * a);
+unsigned long mp_get_ulong (mp_int * a);
 \end{alltt}
 
-This will return the 64 least significant bits of the mp\_int $a$.
+This will return the least significant bits of the mp\_int $a$ that fit into an ``long''.
```
czurnieden: You might insist on using "into" instead of "in" because you "move the bits into a bucket named 'long'" but it's still "a long", no vowel or mute H at the start of the word "long".

Are there really no native speakers left in this group?

minad: Sorry, I can do better than that. I removed the unsigned but didn't remove the "an".

It is just that I didn't took much care and I am not a big fan of using TeX for API docs. Generally I am a big TeX fan but not for this.

minad: fixed

doc/bn.tex
minad: fixed

minad:
@czurnieden Ok now?

bn_conversion.c
```diff
@@ -0,0 +1,95 @@
...
+#ifdef BN_MP_SET_UINT_C
+MP_SET_UNSIGNED(mp_set_uint, 32)
+#endif
+
+#ifdef BN_MP_SET_UINT64_C
+MP_SET_UNSIGNED(mp_set_uint64, 64)
+#endif
+
+#ifdef BN_MP_SET_SINT_C
+MP_SET_SIGNED(mp_set_sint, mp_set_uint, 32)
```
sjaeckel: why `sint` and not `int`? IMO it looks and feels wrong

minad: Because int is in use already :((( This is the whole point of this PR. 

However I actually started to like sint, since it makes it very obvious that the sign is taken into account!

minad: Maybe I should add - before that I used the convention `mp_set_i` and `mp_set_u` , which I think are much uglier.

sjaeckel: > `mp_set_i` and `mp_set_u` , which I think are much uglier.

that's very true

> Because int is in use already

oops, right, but wait... did we already talk about `mp_set_int32` as possibility? that would also align better with the 64 variant...

> However I actually started to like sint...

first symptoms of the stockholm syndrom? :P

minad: > oops, right, but wait... did we already talk about mp_set_int32 as possibility? that would also align better with the 64 variant...

Yes we did. I named things like this to make things feel a bit less heavy during updating. `mp_set_int --> mp_set_uint` instead of `mp_set_int --> mp_set_uint32`. int32 is somehow the "default" and the longer int64 gets the suffix

I could change it and remove the bias towards 32 bits this way if all of you prefer that.

minad: What I should say - I somehow tested the API change when I converted test.c and the code base itself. And the names I am using now felt less heavy but still somehow consistent.

The more precise and even more consistent variant would be to add the 32 suffix for sure.

sjaeckel: > The more precise and even more consistent variant would be to add the 32 suffix for sure.

I think we should go for it

sjaeckel: @czurnieden ?

minad: Fortunately this change is just one sed away

minad: I think @czurnieden does not bother.

minad:
@sjaeckel There you go

tommath.h
```diff
@@ -270,38 +270,53 @@ mp_bool mp_isodd(const mp_int *a) MP_WUR;
...
-/* get a 32-bit value */
-unsigned long mp_get_int(const mp_int *a) MP_WUR;
+/* get magnitude */
+uint32_t mp_get_mag32(const mp_int *a) MP_WUR;
+uint64_t mp_get_mag64(const mp_int *a) MP_WUR;
 
-/* get a platform dependent unsigned long value */
-unsigned long mp_get_long(const mp_int *a) MP_WUR;
+/* get integer, set integer (long) */
+#define mp_get_slong(a)        (sizeof (long) == 8 ? (long)mp_get_int64(a) : (long)mp_get_int32(a))
```
sjaeckel: `mp_get_long`?

tommath.h
```diff
@@ -270,38 +270,53 @@ mp_bool mp_isodd(const mp_int *a) MP_WUR;
...
+#define mp_set_llong(a, b)     (sizeof (long) == 8 ? mp_set_int64((a), (b)) : mp_set_int32((a), (int32_t)(b)))
```
sjaeckel: What's this? I couldn't find documentation
or just another typo?

tommath.h
```diff
@@ -270,38 +270,53 @@ mp_bool mp_isodd(const mp_int *a) MP_WUR;
...
 
-/* get a platform dependent unsigned long long value */
-unsigned long long mp_get_long_long(const mp_int *a) MP_WUR;
+/* get integer, set integer (unsigned long) */
+#define mp_get_ulong(a)       (sizeof (long) == 8 ? (unsigned long)mp_get_uint64(a) : (unsigned long)mp_get_uint32(a))
+#define mp_get_maglong(a)     (sizeof (long) == 8 ? (unsigned long)mp_get_mag64(a) : (unsigned long)mp_get_mag32(a))
```
sjaeckel: `mp_get_mag_long`?

bn_conversion.c
czurnieden: > I think we should go for it

Then we should go for it!

> I think @czurnieden does not bother.

If it isn't way too, uhm, odd I do not care, that is correct.
But you are also right that I would prefer the precise variants, too.

tommath.h
minad: Oh shit. The issue is that get, set long is already taken. I should have stayed in stockholm instead of giving in to you :-P

minad: Before I had get_ul get_sl and then I changed it to get_ulong get_slong when I changed to get_sint etc

tommath.h
minad: Ok

tommath.h
minad: Maybe we should stay with sint32 uint32 slong ulong etc, this way we are consistent and ugly in every dimension.

minad: Or we just revert to what I proposed before.

sjaeckel: it's too much, my head explodes, I have to go to bed, sorry.

just leave it for now, we'll find a good way :)

minad: I updated the latest commit now, it uses int32, uint32 etc but keeps slong for now. gn8

tommath.h
minad: mp_get_long is taken

minad:
@sjaeckel I made another proposal, `_l` and `_ul` suffix for the long macros.

sjaeckel:
that looks very good now! I like it

minad:
Thanks

minad:
If you have time to merge other stuff, I consider #273, #288, #290 and #291 as ready

PR #286, opened 2019-05-22
pure annotation

fperrad:
see https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#Common-Function-Attributes


minad:
And maybe I should add - as I investigated this before and came to the conclusion I just gave. If you make a good argument and show me where these attributes are a good idea, I would be really eager to learn!

fperrad:
When playing with `pure`, I found https://github.com/libtom/libtommath/commit/3cf0c188837d465744396c6b637b02efbeee5c32

It is already a good result.

PR #287, opened 2019-05-22
some linting

fperrad:


minad:
@sjaeckel Ops, hit the wrong button.

PR #292, opened 2019-05-24
some linting

fperrad:


tommath_private.h
```diff
@@ -160,6 +160,7 @@ typedef private_mp_word mp_word;
 #define MP_IS_ODD(a)  (((a)->used > 0) && (((a)->dp[0] & 1u) == 1u))
 
 #define MP_SIZEOF_BITS(type)    ((size_t)CHAR_BIT * sizeof(type))
+#define MP_SSIZEOF_BITS(type)   (CHAR_BIT * (ssize_t)sizeof(type))
```
minad: ssize_t is often not available and we don't use it in the codebase. I also think this macro does not add significant value. Furthermore I would prefer if we move to unsigned indices/sizes where possible instead of adding another signed variant.

Please revert.

minad:
rest lgtm

tommath_private.h
fperrad: useless after the merge of #285

PR #293, opened 2019-05-24
Simplify scripts

minad:
* remove updatemakes
* ~start to remove dep.pl (should be part of helper.pl)~
* dep.pl is now part of helper.pl


minad:
@sjaeckel I hope we can avoid adding new scripts and put all the generation tasks inside helper.pl. dep.pl is a legacy script which is now also part of helper.pl

sjaeckel:
there's one thing that breaks... no `./updatemakes.sh -c` anymore ;)

minad:
> there's one thing that breaks... no ./updatemakes.sh -c anymore ;)

Yes, that's intentional. I think it is better if the scripts don't mess with git and I never used this. Agree?

minad:
Haha, the data agrees with me.

~~~
$ git log --oneline | grep 'Update makefiles$' | wc
0 0 0
~~~

Issue #240, opened 2019-05-08
Feature request: mp_expt


nomeata:
```
~/build/libtommath $ grep expt tommath.h 
int mp_2expt(mp_int *a, int b);
int mp_expt_d(const mp_int *a, mp_digit b, mp_int *c);
int mp_expt_d_ex(const mp_int *a, mp_digit b, mp_int *c, int fast);
int mp_exptmod(const mp_int *G, const mp_int *X, const mp_int *P, mp_int *Y);
```

I am missing
```
int mp_expt(const mp_int *a, const mp_int *b, mp_int *c);
```
I know that unless a is -1, 0 or 1, a power that does not fit in a `mp_digit` will definitely go out-of-memory. But as a user of this library it would still be nice to have libtommath take care of handling the three bases where it does work, and also to provide a nice uniform interface.

minad:
See here https://github.com/libtom/libtommath/pull/201. I am not a fan honestly. This function won't work in most cases. I would rather like to have mp_expt_int or mp_expt_long which doesn't depend on the possibly small size of mp_digit (MP_8BIT => sizeof(mp_digit)==1).

czurnieden:
>  I am not a fan honestly. 

Neither am I but something is definitely needed for the low-mp users.

But another *small* data type besides `mp_digit` does not make much sense. In that case I would prefer to deprecate `int mp_expt_d(const mp_int *a, mp_digit b, mp_int *c);` in favor of `int mp_expt_d(const mp_int *a, unsigned long b, mp_int *c);`.

Adapting to that API is nothing more than a cast for the users (`unsigned long` is large enough to hold all `MP_DIGIT_BIT` sizes). Not much work, I think.

Also not much work would it be to implement it.
Again, not much work to add a macro/function to do the aforementioned casting for the users, so no API change, just additional functionality-

minad:
> Neither am I but something is definitely needed for the low-mp users.

Why? What is the advantage in contrast to mp_expt_int or mp_expt_long? 

I agree that mp_expt_d should be deprecated in favor of mp_expt_int then. I generally don't like the exposure of mp_digit in the public api. Functions should take int/uint instead.

czurnieden:
> Why? What is the advantage in contrast to mp_expt_int or mp_expt_long?

[...]

PR #294, opened 2019-05-25
deprecate mp_n_root_ex and mp_expt_d_ex

minad:
These functions were introduced to give some timing guarantees.
However the guarantees are too weak to be useful.
The functions seem to be essentially unused by downstream users.

I checked https://github.com/heimdal/heimdal and ltc.

See the discussion in #243. See also #202 which will give us a better root algorithm.


czurnieden:
I overlooked that you removed the constant time version of expt_d instead of repairing it. Intentional?

minad:
Yes, I think most users don't want the constant time version. It is not even constant time I think. But even if it was, I've not seen it being used in the crypto applications ltc and heimdal.

We could however also keep expt_d as "constant time" and add another non constant time function instead of the _ex version. But I think it is the wrong default and we shouldn't make false promises. I certainly would not vouch for expt_d being side channel free...

minad:
@czurnieden If you have a better idea, please let me know. What I thought - we could maybe also deprecate both mp_expt_d and mp_expt_d_ex and introduce another mp_expt_int + mp_expt (as in #240) with the prototypes:

~~~
mp_err mp_expt(const mp_int*, const mp_int*, mp_int*);
mp_err mp_expt_int(const mp_int*, uint32_t, mp_int*);
~~~

This way we would avoid any complaints regarding changed behavior and would end up with a good api after the deprecated stuff has been removed.

czurnieden:
> [slow expt_d] is not even constant time I think. 

It isn't as it is but can be made at least constant-branching by adding the necessary multiplications (just multiply by one) but I don't think it would make a large difference.
Hence you're right in getting rid of it.

Have you changed the documentation? >;->

czurnieden:
(Sometimes the javascript on this page works, sometimes not, feels as if MS has finally taken over ;-) )

`mp_expt_int` is a bit off of your naming scheme?

I would support it but others here think that a full bigint `mp_expt` would be too much of a niche product and that is not a bad argument. But we have user who likes to have it, if I rember it correctly?

minad:
> mp_expt_int is a bit off of your naming scheme?

Yes, we can also use a _u32 name.

> I would support it but others here think that a full bigint mp_expt would be too much of a niche product and that is not a bad argument. But we have a user who likes to have it, if I rember it correctly?

As discussed before - I would not support a full bigint mp_expt, but only a reduced version which extracts the low bits from the bigint exponent for special cases. There was at least an argument for that in #240. But I am not really convinced tbh.

I would personally happy with mp_expt_u32 only. And I would prefer that over mp_expt_d since this gives us api consistency across platforms.

minad:
@czurnieden @sjaeckel Here we can check for soon to be deprecated functions being in use - https://codesearch.debian.net/search?q=mp_expt_d_ex

PR #224, opened 2019-04-18
[WIP] add windows builds to travis

sjaeckel:

PR #262, opened 2019-05-14
add feature detection macro MP_HAS (2)

minad:
See #214. I wanted to try again if I can make this work without using variadic macros, this time only relying on C89.

@karel-m Could you please try again your compiler suite?

@sjaeckel @czurnieden This PR is simpler and better than the one before. Hopefully it also works on older compilers. This helps to clean up control flow. Another advantage is that the compiler will check all branches. Dead code elimination will make sure that the dead branches are not compiled.

PR #295, opened 2019-05-25
fix invalid OBJECTS=... definition in makefile.msvc 

karel-m:
Related to #224 

Fixes `fatal error U1034: syntax error : separator missing` mentioned in the following comment:

> I tried quickly using appveyor and at least I managed to start "nmake -f makefile.msvc all" but then I got a parse error (seperator : missing or sth)?! Can some windows user please try this?


minad:
Thank you @karel-m! I will try this! Could you also quickly check #262 since I have no possibility to test those various compilers.

minad:
I was looking for a missing backslash but I didn't see it :see_no_evil:

minad:
superseded by #297

PR #296, opened 2019-05-26
minor improvements&fixes

czurnieden:
The path to the output of the tuner `test_*.log` was wrong, it pointed to the parent directory.

[Edit] 
* Tests of `mp_shrink()` are added
* The path of `etc/tune_it.sh` is correctly determined, so you can also run e.g `./etc/tune_it.sh` from the project root


sjaeckel:
I took the liberty to add two more commits to this PR which are also minor changes/corrections

PR #297, opened 2019-05-26
Msvc fixes (appveyor works!)

minad:


minad:
@sjaeckel I think we should just go with appveyor for win. Look how simple the yamml is! It also works (in my personal ltm) - I still have to set it up here.

sjaeckel:
I would've expected that the build works like that, but it seems that's not the case... did you do any special configuration in your appveyor project?
...also I couldn't find how to put the build under the libtom organisation...

PR #300, opened 2019-05-28
Better interoperability between MSVC and mingw-w64, 3

nijtmans:
Another try, sucessor of the -2 branch. I don't know how to integrate generate_def.sh with helper.pl.

The .def-file doesn't include deprecated symbols, that's on purpose: I don't really care about those ;-)


nijtmans:
B.T.W: the current  makefile.mingw in develop is wrong: it contains too many lines with duplicate .o-files. Fixed in this branch!

minad:
> ...we have to fix the dependency system...

@czurnieden and I talked about improving the configuration system before. Maybe I find time to work on this at some point and propose a PR.

nijtmans:
> I generally approve but I think we have to fix the dependency system to handle bn_conversion.c and bn_deprecated.c. Furthermore the def generation should be part of helper.pl later.

Agreed!   Actually, I don't care about bn_deprecated.c, but the bn_conversion.c symbols - indeed - must be handled separately. I don't know how to do that better, without actually deriving the symbols from a built library. Yes, it should be part of helper.pl later!

minad:
> I don't know how to do that better, without actually deriving the symbols from a built library.

Either we split bn_deprecated/bn_conversion again per function or we improve the regex "parser" in helper.pl which generates tommath_class etc. The def file should also be generated from that data then.

nijtmans:
I think this one is ready to be merged to develop. Further improvements/cleanups in generating the .def-file can be done in a separate PR.

minad:
Agree, there is already something going on in #301

minad:
@sjaeckel I think this is ready too?

nijtmans:
@sjaeckel    ping ???

PR #274, opened 2019-05-19
some refactoring

fperrad:

PR #302, opened 2019-05-29
some refactoring

fperrad:


fperrad:
Like #274, this serie improves the code consistency.
Except the commit `avoid assignment inside condition` which removes an idiom.

sjaeckel:
I guess it's pretty clear that the 4 commits besides "avoid assignment inside condition" aren't worth a discussion.

I added you guys as reviewers to comment on the advantages and disadvantages of keeping as is vs. merging the "avoid assignment inside condition" change.

minad:
I would not merge the assignment commit. Besides that, it is fine with me.

minad:
Well, the code style is with assignments as of now. There is no harm as it is. Why change it? I think I said it before - while I like consistency, I don't like pure style changes. They obfuscate the git history. The other commits in this PR are better, for example scope or error handling with goto - these improve the code.

sjaeckel:
@fperrad can you please rebase and remove the 'avoid assignment inside condition' commit

PR #310, opened 2019-06-05
Formatting strange definitions of pointers to functions returning mp_err

nijtmans:
Noting that function pointer definitions, where the function returns mp_err, are strangely formatted: They look like function calls. If the function returns int or long, formatting is OK.

This commit fixes this strange formatting.

### DIFF
--- a/bn_mp_rand.c
+++ b/bn_mp_rand.c
@@ -3,9 +3,9 @@
 /* LibTomMath, multiple-precision integer library -- Tom St Denis */
 /* SPDX-License-Identifier: Unlicense */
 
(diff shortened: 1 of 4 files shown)


sjaeckel:
I cancelled the build as the first step failed... auto-formatting the code. It seems like what you want to have fixed is an issue of astyle, but I didn't investigate any further

nijtmans:
> I cancelled the build as the first step failed... auto-formatting the code. It seems like what you want to have fixed is an issue of astyle, but I didn't investigate any further

Yes, it looks like an astyle issue.

minad:
Close. Instead astylerc should be fixed?

Issue #309, opened 2019-06-05
The new approach used in bn_conversion.c makes it impossible to keep no-stdint-h


karel-m:
Some of you might know about my branch `no-stdint-h` that avoids ulitising `stdint.h` which is not available on some compilers.

In v1.1.0 it was as simple as this: https://github.com/libtom/libtommath/commit/602e36ec9ecd97d6af517c462929b523cf26844f

However, since commit https://github.com/libtom/libtommath/commit/3289c959 especially with macro-magic introduced in `bn_conversion.c` it is not possible anymore to maintain `no-stdint-h` branch.

I do not agree with changes introduced by https://github.com/libtom/libtommath/pull/285 What exactly was the trouble this PR fixes?

czurnieden:
> What exactly was the trouble this PR fixes?

On point is the lack of signed alternatives, but that is only a minor one. The main problems are the headaches caused by the needs to check if e.g.: and `int` is 16 bit of 32 bit large, or if the `long` has 64 bits or only 32, the need to check what of the five possible sizes `mp_digit` has. Most of the single digit `*_d` functions have a very restricted usability with low-mp, especially with 8-bit.

That eats a lot of  of developing time, needs extended testings, and the necessary maintenance for all of the possible variations, too.

The functions in `bn_conversion.c` were introduced to get rid of some of the problems listed above which eases maintenance and development with LTM.

We were aware that it comes with the cost of killing backwards-compatibility for some use-cases.
Your problem can be solved by taking `bn_conversion.c` apart such that you can do your search&replace.

I took the liberty to suggest #311 (A quick C&P from #301, hope it went well).

minad:
@karel-m We (@sjaeckel, @czurnieden and me) decided to use precise types since this increases portability and reduces the maintenance effort. If your compiler does not provide a stdint.h, just add one yourself. It only contains a handful of typedefs. I am using tommath myself in scenarios without libc and I am in exactly the same situation, using my own tiny stdint.h header. Is there an issue which would prevent you from doing that?

minad:
@nijtmans You mentioned before that tcl avoids stdint.h. Is this still the case?

@karel-m @nijtmans Would something like the follwoing help you? Then you have to provide these sized types yourself.

~~~ c
#ifndef MP_NO_STDINT
typedef uint8_t mp_u8;
typedef uint16_t mp_u16;
typedef uint32_t mp_u32;
typedef uint64_t mp_u64;
typedef int8_t mp_i8;
[...]

Issue #104, opened 2018-03-02
libtomath build broken on win32, due to the use of uint8_t


nijtmans:
When importing the "develop" branch of libtom/libtommath into Tcl, it broke the win32 build. The reason is that no header file is used which defines uint8_t, still this symbol is used in tommath_private.h and bn_mp_radix_smap.c. (the use in tommath.h is harmless, since Tcl doesn't use MP_8BIT)

nijtmans:
Suggested fix: Use "unsigned char" in stead of "uint8_t" in two places. See:   http://core.tcl.tk/tcl/info/8531135e454c2231

sjaeckel:
Can you please test if this is solved when using the [msvc-friendly](https://github.com/libtom/libtommath/tree/msvc-friendly) branch by @karel-m ?
@karel-m can you please open a PR for this branch after the OK of @nijtmans ?

sjaeckel:
@nijtmans did you have a chance to test this?

czurnieden:
If I remember it correctly the proper (calling it "proper" when doing it the MS way *does* bear some irony, doesn't it? ;-) ) way would be

    #ifdef (_MSC_VER < 1300)
         typedef unsigned char uint8_t;
    #else
        typedef unsigned __int8 uint8_t;
    #endif

karel-m:
The proper way is IMO https://github.com/libtom/libtommath/blob/msvc-friendly/tommath.h#L20

But it did not help

czurnieden:
Nothing more than the normal fixed-width integers are used in libtommath, the inclusion of the full `stdint.h` is not really necessary.

So, @nijtmans does it help when the lines 22 and 23 in the MSVC-friendly version of `tommath.h` are commented out to use the manually defined  types only?

nijtmans:
No that doesn't help, sorry, at least not for Tcl!   Since Tcl doesn't want to use <stdint.h> at all, we made some modifications to typedef the tommath types (mp_digit|mp_word) to (unsigned char|unsigned short) directly without the intermediate (uint8_t|uint16_t).  So I would prefer libtommath to use either it's own types (mp_digit|mp_word) in it's implementation, either standard C types. The 2 uses of uint8_t in tommath_private.h and bn_mp_radix_smap.c could as well be replaced with "unsigned char" without any loss in functionality. 

So
    extern const uint8_t mp_s_rmap_reverse[];
should become
    extern const unsigned char mp_s_rmap_reverse[];
[...]

Issue #109, opened 2018-04-18
Need help with mp_exptmod and BigInts


VerHext:
I would like to calculate the following in C ++ with libtommath:
`x ^ (2 ^ level) % n`

So I try it ...
```c
        char buff[4096];
        mp_int x, level, n, X, e;

        mp_init(&x);
        mp_init(&level);
        mp_init(&n);
        mp_init(&X);
        mp_init(&e);

        // Calculate RSA  x ^ (2 ^ level) % n

        mp_set(&x, 32878006774570359216307190512414453734815024711845858985814302013359906676224571864057517441898232179839316906470039235887748992902358250854400787652810336070530280967470770997285440429053186803879037);
        mp_set(&level, 10000);
        mp_set(&n, 77898130960070341501772069500669364440519531421534783575397763758775619778096560479521554583192022357575799725548012588149166448319424189949242058358050730052508358466295626425829884371399991831978634);
        mp_set(&e, 2);

        mp_sqr(&e, &level);
        mp_toint(&level, buff);
        printf("LVL = %s\n", buff );  // 4
        mp_exptmod(&x, &level, &n, &X);

        mp_toint(&X, buff);
        printf("X :: %s\n", buff); // 346751013179386221

//Warning: integer constant is too large for its type mp_set(&n, ...);
//Warning: integer constant is too large for its type mp_set(&x, ...);

```
It just doesn't calculate the big numbers correct ... or I'm doing something wrong ...

czurnieden:
Yes, you are doing something wrong, almost all, actually. `mp_set` accepts small integers only, not big integers, you need to read them in as strings with `mp_read_radix()`. To compute `2^level` you need to use either `mp_exp_d()` or, much better here, just a shift with `mp_mul_2d()`. The function `mp_sqr(a,b)` does `b = a^2` which is not what you want to do. The function `mp_toint()` is not able to do what you want, you need to use e.g. `mp_toradix ()`.

LibTomMath has a lot of good documentation, please read it.

tomstdenis:
I constantly in awe and impressed that this long after I (sadly) gave up working on the projects people still are working on them and improving them.  :-)  Thanks to all who support free open source software.

(quick note, In case people are wondering I actually do open source at AMD on the graphics side.  So I'm still fighting the good fight, just have a family to feed now :-)).

PR #311, opened 2019-06-05
Removed bn_conversion.c in favor of individual files

czurnieden:
In reaction to #104 and #309.
It also makes development in #301 easier.


bn_mp_get_i32.c
```diff
@@ -0,0 +1,15 @@
+#include "tommath_private.h"
+#ifdef BN_MP_GET_I32_C
+/* LibTomMath, multiple-precision integer library -- Tom St Denis */
+/* SPDX-License-Identifier: Unlicense */
+
+
+/* Return the least significant 32 bits */
+int32_t mp_get_i32(const mp_int *a)
+{
+   uint64_t res = mp_get_mag32(a);
```
nijtmans: Shouldn't this be uint32_t ?    I think it should.

bn_mp_get_i32.c
```diff
@@ -0,0 +1,15 @@
...
+   return (a->sign == MP_NEG) ? (int32_t)-res : (int32_t)res;
```
nijtmans: Better do the (int32_t) typecast first, then the negation:  Negation of unsigned integers doesn't make much sense (I have seen compiler warnings related to this in another project, don't remember which compiler/flags)

bn_mp_get_i64.c
```diff
@@ -0,0 +1,15 @@
...
+#ifdef BN_MP_GET_I64_C
+/* LibTomMath, multiple-precision integer library -- Tom St Denis */
+/* SPDX-License-Identifier: Unlicense */
+
+
+/* Return the least significant 64 bits */
+int64_t mp_get_i64(const mp_int *a)
+{
+   uint64_t res = mp_get_mag64(a);
+   return (a->sign == MP_NEG) ? (int64_t)-res : (int64_t)res;
```
nijtmans: Same remark as previous one

bn_mp_set_i32.c
```diff
@@ -0,0 +1,15 @@
+#include "tommath_private.h"
+#ifdef BN_MP_SET_I32_C
+/* LibTomMath, multiple-precision integer library -- Tom St Denis */
+/* SPDX-License-Identifier: Unlicense */
+
+
+/* Set the least signficant 32 bit of "a" to b, signed */
+void mp_set_i32(mp_int *a, int32_t b)
+{
+   mp_set_u32(a, (b < 0) ? -(uint32_t)b : (uint32_t)b);
```
nijtmans: Since negation of an unsigned doesn't make much sense (see previous comment), better do negation first, then the (uint32_t) type-cast

bn_mp_set_i64.c
```diff
@@ -0,0 +1,17 @@
+#include "tommath_private.h"
+#ifdef BN_MP_SET_I64_C
+/* LibTomMath, multiple-precision integer library -- Tom St Denis */
+/* SPDX-License-Identifier: Unlicense */
+
+
+/* Set the least signficant 64 bit of "a" to b, signed */
+void mp_set_i64(mp_int *a, int64_t b)
+{
+   mp_set_u64(a, (b < 0) ? -(uint64_t)b : (uint64_t)b);
```
nijtmans: Same as previous remark

bn_mp_get_i32.c
czurnieden: Ouch!
Thanks!
(But that is in the part of the preprocessor expansion? Or really a C&P error? Need to check)

czurnieden: > But that is in the part of the preprocessor expansion?

Yes, it is in the result of `gcc -E`.
So if this PR is not wanted: @minad , please check your macros.

czurnieden: I think both, the `uint64_t` and the `(int32_t)-res` are intentional to avoid UB in case of overflow.
I don't think it works but I was wrong once, so I'll keep my voice low ;-)

bn_mp_get_i32.c
nijtmans: Remember now what warnings I saw: It was Visual Studio 2017:

bn_conversion.c
bn_conversion.c(64): warning C4146: unary minus operator applied to unsigned type, result still unsigned
bn_conversion.c(68): warning C4146: unary minus operator applied to unsigned type, result still unsigned
bn_conversion.c(72): warning C4146: unary minus operator applied to unsigned type, result still unsigned
bn_conversion.c(76): warning C4146: unary minus operator applied to unsigned type, result still unsigned

czurnieden: (VS2015, too)
It also says

```text
demo/test.c(677): warning C4102: 'LBL_ERR': unreferenced label
demo/test.c(627): warning C4101: 'i': unreferenced local variable
```

Which makes no sense at all for me.

I rolled the changes back and will stash when Travis gives its OK.

czurnieden:
> I'm OK now (the type-cast / negation order is not crucial, I even think that other compilers might disagree with MSVC ....)

They way it is done makes it a defined behaviour but does not solve the problem with overflow *per se*. Put a note in the documentation or is this behaviour obvious?

minad:
> They way it is done makes it a defined behaviour but does not solve the problem with overflow per se. 

No. The way negation is done now solves the undefined behavior, since overflow of unsigned numbers is well defined.

bn_mp_get_i32.c
minad: No, the cast should happen in this order. Otherwise we get undefined behavior. Furthermore negation of unsigned numbers is defined.

minad:
Hmm, the macros were quite ugly and generic programming in c simply sucks. However now we also get a bit of code duplication and the functions can get out of sync, which I am not very fond of. #109 and #309 are non-issues and #301 can also be solved with bn_conversion.c as is.

bn_mp_set_i32.c
minad: To elaborate, this would introduce undefined behavior for -0x80000000, resulting in signed integer overflow.

minad:
@czurnieden Before we had these MP_SET_XLONG macros, bn_conversion does the same. Keeping the macros might not make sense, all the macros are used only twice. If you want to fix #301, you could simply split bn_conversion in multiple files, keeping the macros. However we could think of supporting multiple functions per file as introduced in bn_deprecated.c and bn_conversion.c. This was in fact not my idea, but @sjaeckel suggested to put the deprecated stuff in a single file. But I think it is a good idea actually. For the bn_conversion case, splitting might make sense.

czurnieden:
> No. The way negation is done now solves the undefined behavior, since overflow of unsigned numbers is well defined.

Ah, misunderstanding.
I meant the actual overflow, that is if `mp_get_magx()` returns a value larger than 2^(x-1). The behaviour of the code is defined but the user does not have an easy way to detect it. Normally a flag would be set like e.g.: `errno` but we don't use such flags (It would be a real pain in the behind to implement it). The only way to be sure is to check the size of the input which has to be done by the caller, so I'd like to propose a little reminder in the documentation, maybe something like: "The size of the input must be checked by the caller."

> However we could think of supporting multiple functions per file as introduced in bn_deprecated.c and bn_conversion.c

The stuff in `bn_deprecated.c` is, well, deprecated, so it does not make a lot of sense to invest in them.

`bn_conversion.c` on the other hand has all the new stuff in it. It is helpful for at least two people here if it gets cut in multiple files.

> split bn_conversion in multiple files, keeping the macros

The macros are not needed, they do not rely on many compile-time variables, they are just generating code. This PR  keeps the necessary compile-time variables (constants actually, like e.g.: MP_DIGIT_BIT) and cuts it in individual files.

It is just a different formatting, no change in functionality.

> However now we also get a bit of code duplication

Yes. And?
The macros would have generated the same code with the same duplications.

> and the functions can get out of sync, 

Out of sync with what? `bn_conversion.c`? That file doesn't exist anymore. With their "twins"? Yes, that can happen, so a bit more care is necessary when repairing one of the functions, you have to check their "twin", too.

With `bn_conversion.c` cut into individual files it now follows the LTM principle of "one function per file" and it's naming convention. With the exceptions `bn_prime_tab.c`, `bn_cutoffs.c` containing constants and of course `bn_deprecated.c`  the single exception with several functions in it *and* a different naming scheme.

minad:
> With their "twins"? Yes, that can happen, so a bit more care is necessary when repairing one of the functions, you have to check their "twin", too.

@czurnieden Do you think it makes sense to keep the macros to avoid this out of sync issue between the twins? The macros should be improved to be more readable however and get a type argument instead of a bit integer argument as @nijtmans suggested.

czurnieden:
@minad f'up at #313?

minad:
@czurnieden Ok to close this in favor of #313?

PR #314, opened 2019-06-06
MSVC compile:  higher warning-level -Wall, but warning-free

nijtmans:
This updates the warning level for MSVC compiles to -Wall (from -W3).   At the same time, it removes 5 specific MSVC warnings, which are simply wrong and used legitimitely in libtommath:

C4146: unary minus operator applied to unsigned type, result still unsigned
C4127: conditional expression is constant
C4710: 'function' : function not inlined
C4711: function 'function' selected for inline expansion
C4820: 'bytes' bytes padding added after construct 'member_name'

### DIFF
--- a/demo/test.c
+++ b/demo/test.c
@@ -622,6 +622,7 @@ LBL_ERR:
 
 }
 
+#if defined(__STDC_IEC_559__) || defined(__GCC_IEC_559)
 static int test_mp_set_double(void)
 {
    int i;
@@ -632,7 +633,6 @@ static int test_mp_set_double(void)
    }
 
    /* test mp_get_double/mp_set_double */
-#if defined(__STDC_IEC_559__) || defined(__GCC_IEC_559)
    if (mp_set_double(&a, +1.0/0.0) != MP_VAL) {
       printf("\nmp_set_double should return MP_VAL for +inf");
       goto LBL_ERR;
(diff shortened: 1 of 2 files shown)


minad:
@nijtmans is it possible to enable werror on appveyor? We have this enabled for gcc/clang on Travis

nijtmans:
> @nijtmans is it possible to enable werror on appveyor? We have this enabled for gcc/clang on Travis
Well, In Visual Studio 2015 (AppVeyor) I see:

C:\Program Files (x86)\Windows Kits\10\include\10.0.14393.0\um\winbase.h(8946): warning C4668: 'defined_WINBASE_' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'
C:\Program Files (x86)\Windows Kits\10\include\10.0.14393.0\shared\stralign.h(450): warning C4013: 'wcscpy_s' undefined; assuming extern returning int

So, that only can be done if it is _really_  warning-free ;-)

minad:
Oh ok, we are holier than the pope. It is the same with Wconversion warnings and standard headers on linux.

sjaeckel:
well I think it'd make sense to make the code warning-free, also on windows, so we can enable -Werror there

minad:
> well I think it'd make sense to make the code warning-free, also on windows, so we can enable -Werror there

Maybe we could disable the problematic warnings by using msvc-specific pragmas around the critical code blocks?

nijtmans:
> Maybe we could disable the problematic warnings by using msvc-specific pragmas around the critical code blocks?

Yes, that could work.  I'll try

nijtmans:
See #316 as first attempt

PR #315, opened 2019-06-06
remove generate_def.sh, move code to helper.pl

minad:
don't shell out to generate_def.sh, do it in perl instead.


tommath.def
```diff
@@ -93,15 +93,14 @@ EXPORTS
     mp_prime_strong_lucas_selfridge
     mp_radix_size
     mp_rand
-    mp_rand_digit
```
minad: mp_rand_digit is deprecated --> removed

tommath.def
```diff
@@ -23,10 +23,10 @@ EXPORTS
     mp_copy
     mp_count_bits
     mp_decr
+    mp_div
```
minad: file looks the same except for the sorting, perl sorts differently than sort command it seems. sorry for that.

Issue #317, opened 2019-06-11
clang-tidy warnings


karel-m:
`clang-tidy` run win the same rule set as we use in libtomcrypt (and have no warnings):
```
bn_mp_cmp.c:13:9: error: do not use 'else' after 'return' [readability-else-after-return,-warnings-as-errors]
      } else {
        ^~~~~~
bn_mp_cmp.c:22:6: error: do not use 'else' after 'return' [readability-else-after-return,-warnings-as-errors]
   } else {
     ^~~~~~
bn_mp_cmp_d.c:22:6: error: do not use 'else' after 'return' [readability-else-after-return,-warnings-as-errors]
   } else if (a->dp[0] < b) {
     ^~~~~~~~
bn_mp_decr.c:13:6: error: do not use 'else' after 'return' [readability-else-after-return,-warnings-as-errors]
   } else if (a->sign == MP_NEG) {
     ^~~~~~~~
bn_mp_exptmod.c:81:6: error: do not use 'else' after 'return' [readability-else-after-return,-warnings-as-errors]
   } else {
     ^~~~~~
bn_mp_incr.c:12:6: error: do not use 'else' after 'return' [readability-else-after-return,-warnings-as-errors]
   } else if (a->sign == MP_NEG) {
     ^~~~~~~~
bn_mp_prime_is_prime.c:148:7: error: Value stored to 't' is never read [clang-analyzer-deadcode.DeadStores,-warnings-as-errors]
      t = -t;
      ^
bn_mp_prime_is_prime.c:148:7: note: Value stored to 't' is never read
bn_mp_prime_rabin_miller_trials.c:33:9: error: do not use 'else' after 'return' [readability-else-after-return,-warnings-as-errors]
      } else if (sizes[x].k > size) {
        ^~~~~~~~
bn_mp_prime_rand.c:71:17: error: The left expression of the compound assignment is an uninitialized value. The computed value will also be garbage [clang-analyzer-core.uninitialized.Assign,-warnings-as-errors]
      tmp[0]    &= maskAND;
                ^
bn_mp_prime_rand.c:137:11: note: Calling 's_mp_prime_random_ex'
   return s_mp_prime_random_ex(a, t, size, flags, s_mp_rand_cb, NULL);
          ^
bn_mp_prime_rand.c:29:9: note: Assuming 'size' is > 1
   if ((size <= 1) || (t <= 0)) {
        ^
bn_mp_prime_rand.c:29:8: note: Left side of '||' is false
[...]

PR #318, opened 2019-06-12
Removed obsolete code

czurnieden:
@karel-m  run `clang-tidy` and found several things (vid.: #317 ). Two of them are actually obsolete code and this PR removes them.

That code was in cryptographically sensitive parts of LTM, so a second (or more) opinion about the correctness would be highly appreciated!

### DIFF
--- a/bn_mp_prime_strong_lucas_selfridge.c
+++ b/bn_mp_prime_strong_lucas_selfridge.c
@@ -177,26 +177,12 @@ mp_err mp_prime_strong_lucas_selfridge(const mp_int *a, mp_bool *result)
    mp_set(&U2mz, 1uL);  /* U_1 */
    mp_set(&V2mz, (mp_digit)P);  /* V_1 */
 
-   if (Q < 0) {
-      Q = -Q;
-      mp_set_u32(&Qmz, (uint32_t)Q);
-      if ((err = mp_mul_2(&Qmz, &Q2mz)) != MP_OKAY) {
-         goto LBL_LS_ERR;
-      }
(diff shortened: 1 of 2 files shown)


bn_mp_prime_strong_lucas_selfridge.c
```diff
@@ -177,26 +177,12 @@ mp_err mp_prime_strong_lucas_selfridge(const mp_int *a, mp_bool *result)
...
-   if (Q < 0) {
-      Q = -Q;
-      mp_set_u32(&Qmz, (uint32_t)Q);
-      if ((err = mp_mul_2(&Qmz, &Q2mz)) != MP_OKAY) {
-         goto LBL_LS_ERR;
-      }
-      /* Initializes calculation of Q^d */
-      mp_set_u32(&Qkdz, (uint32_t)Q);
-      Qmz.sign = MP_NEG;
-      Q2mz.sign = MP_NEG;
```
minad: What is with Q2mz here? It is negative afterwards.

czurnieden: (Ah, somebody's looking, thanks!)

Yes, that is correct. See the original implementation (file `trn.c`, lines 3145 f.):

```
mpz_set_si(mpzQm, lQ);
mpz_set_si(mpz2Qm, 2*lQ);
```

And because we had no signed `mp_set_*` before I had to go to some lengths which are not necessary anymore.

minad: Ah I see now, I confused something. Looks good :+1:

PR #321, opened 2019-06-12
Provide explicit symbols for bn_xxx_l and bn_xxx_ll functions.

nijtmans:
Please, provide explicit functions for those 8 symbols, in stead of mapping those to the i/u 32/64 versions: There is no guarantee that long or long long are either 32 or 64 bits, there could be platforms where they are 128 bits. Since we have macro's now, it's trivial.

Also I adapted the .tex-file, some symbols were still not documented. Added the long long variants as well.

### DIFF
--- a/tommath.h
+++ b/tommath.h
@@ -299,26 +299,35 @@ uint32_t mp_get_mag32(const mp_int *a) MP_WUR;
 uint64_t mp_get_mag64(const mp_int *a) MP_WUR;
 
 /* get integer, set integer (long) */
-#define mp_get_l(a)        (sizeof (long) == 8 ? (long)mp_get_i64(a) : (long)mp_get_i32(a))
-#define mp_set_l(a, b)     (sizeof (long) == 8 ? mp_set_i64((a), (b)) : mp_set_i32((a), (int32_t)(b)))
+long mp_get_l(const mp_int *a) MP_WUR;
+void mp_set_l(mp_int *a, long b);
 
 /* get integer, set integer (unsigned long) */
-#define mp_get_ul(a)       (sizeof (long) == 8 ? (unsigned long)mp_get_u64(a) : (unsigned long)mp_get_u32(a))
-#define mp_set_ul(a, b)    (sizeof (long) == 8 ? mp_set_u64((a), (b)) : mp_set_u32((a), (uint32_t)(b)))
-#define mp_get_magl(a)     (sizeof (long) == 8 ? (unsigned long)mp_get_mag64(a) : (unsigned long)mp_get_mag32(a))
+#define mp_get_ul(a) ((unsigned long)mp_get_l(a))
+void mp_set_ul(mp_int *a, unsigned long b);
+unsigned long mp_get_magl(const mp_int *a) MP_WUR;
+
+/* get integer, set integer (long long) */
+long long mp_get_ll(const mp_int *a) MP_WUR;
+void mp_set_ll(mp_int *a, long long b);
+
+/* get integer, set integer (unsigned long long) */
+#define mp_get_ull(a) ((unsigned long long)mp_get_ll(a))
+void mp_set_ull(mp_int *a, unsigned long long b);
+unsigned long long mp_get_magll(const mp_int *a) MP_WUR;
 
 /* set to single unsigned digit, up to MP_DIGIT_MAX */
 void mp_set(mp_int *a, mp_digit b);
 mp_err mp_init_set(mp_int *a, mp_digit b) MP_WUR;
 
 /* get integer, set integer and init with integer (deprecated) */
-MP_DEPRECATED(mp_get_mag32/mp_get_u32) unsigned long mp_get_int(const mp_int *a) MP_WUR;
+MP_DEPRECATED(mp_get_magl/mp_get_ul) unsigned long mp_get_int(const mp_int *a) MP_WUR;
 MP_DEPRECATED(mp_get_magl/mp_get_ul) unsigned long mp_get_long(const mp_int *a) MP_WUR;
-MP_DEPRECATED(mp_get_mag64/mp_get_u64) unsigned long long mp_get_long_long(const mp_int *a) MP_WUR;
-MP_DEPRECATED(mp_set_u32) mp_err mp_set_int(mp_int *a, unsigned long b);
+MP_DEPRECATED(mp_get_magll/mp_get_ull) unsigned long long mp_get_long_long(const mp_int *a) MP_WUR;
+MP_DEPRECATED(mp_set_ul) mp_err mp_set_int(mp_int *a, unsigned long b);
 MP_DEPRECATED(mp_set_ul) mp_err mp_set_long(mp_int *a, unsigned long b);
-MP_DEPRECATED(mp_set_u64) mp_err mp_set_long_long(mp_int *a, unsigned long long b);
-MP_DEPRECATED(mp_init_u32) mp_err mp_init_set_int(mp_int *a, unsigned long b) MP_WUR;
+MP_DEPRECATED(mp_set_ull) mp_err mp_set_long_long(mp_int *a, unsigned long long b);
+MP_DEPRECATED(mp_init_ul) mp_err mp_init_set_int(mp_int *a, unsigned long b) MP_WUR;
 
 /* copy, b = a */
 mp_err mp_copy(const mp_int *a, mp_int *b) MP_WUR;
--- a/tommath.h
+++ b/tommath.h
@@ -297,24 +297,28 @@ mp_err mp_init_u64(mp_int *a, uint64_t b) MP_WUR;
 /* get magnitude */
 uint32_t mp_get_mag32(const mp_int *a) MP_WUR;
 uint64_t mp_get_mag64(const mp_int *a) MP_WUR;
+unsigned long mp_get_magl(const mp_int *a) MP_WUR;
+unsigned long long mp_get_magll(const mp_int *a) MP_WUR;
 
 /* get integer, set integer (long) */
 long mp_get_l(const mp_int *a) MP_WUR;
 void mp_set_l(mp_int *a, long b);
+mp_err mp_init_l(mp_int *a, long b) MP_WUR;
 
 /* get integer, set integer (unsigned long) */
 #define mp_get_ul(a) ((unsigned long)mp_get_l(a))
 void mp_set_ul(mp_int *a, unsigned long b);
-unsigned long mp_get_magl(const mp_int *a) MP_WUR;
+mp_err mp_init_ul(mp_int *a, unsigned long b) MP_WUR;
 
 /* get integer, set integer (long long) */
 long long mp_get_ll(const mp_int *a) MP_WUR;
 void mp_set_ll(mp_int *a, long long b);
+mp_err mp_init_ll(mp_int *a, long long b) MP_WUR;
 
 /* get integer, set integer (unsigned long long) */
 #define mp_get_ull(a) ((unsigned long long)mp_get_ll(a))
 void mp_set_ull(mp_int *a, unsigned long long b);
-unsigned long long mp_get_magll(const mp_int *a) MP_WUR;
+mp_err mp_init_ull(mp_int *a, unsigned long long b) MP_WUR;
 
 /* set to single unsigned digit, up to MP_DIGIT_MAX */
 void mp_set(mp_int *a, mp_digit b);
--- a/tommath_private.h
+++ b/tommath_private.h
@@ -264,7 +264,7 @@ MP_DEPRECATED(s_mp_reverse) void bn_reverse(unsigned char *s, int len);
         return MP_OKAY;                                  \
     }
 
-#define MP_GET_MAG(type, name)                                                         \
+#define MP_GET_MAG(name, type)                                                         \
     type name(const mp_int* a)                                                         \
     {                                                                                  \
         unsigned i = MP_MIN((unsigned)a->used, (unsigned)((MP_SIZEOF_BITS(type) + MP_DIGIT_BIT - 1) / MP_DIGIT_BIT)); \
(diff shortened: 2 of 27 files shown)


minad:
I guess you want these symbols since you want to hide the u32/u64 symbols on msvc6 and instead use the l/ul variants. I disagree with providing different APIs on different compilers. This is not good, things should be consistent across platforms. Note that you can replace the u32/u64 functions with long/long long prototypes on msvc6 due to abi compatibility. Only the typedef is not available.

I don't buy the argument regarding 128 bit platforms. If there such platforms we can instead add i128/u128 variants.

I suggest not to

karel-m:
> This is not good, things should be consistent across platforms.

Quite a strong statement and completely wrong. It is **only your opinion** which you are pushing deeper and deeper into libtommath during last couple of months.

Have you got a chance to look into what big guys from top bignum libraries (GMP, OpenSSL/BIGNUM guys, LibreSSL ...) do?
* https://gmplib.org/manual/Assigning-Integers.html
* https://github.com/libressl/libressl/blob/master/src/doc/crypto/bn.pod
* https://www.openssl.org/docs/man1.1.1/man3/BN_set_word.html
* Providing a least basic signed/unsigned long setter/getter is de facto standard.
* None of them is fiddling with `stdint.h` (of course there is a chance, IMHO quite small, that `<joke>`everybody else except us is wrong, maybe they just did not realize that `stdint.h` exists `</joke>`).

Look, the libtommath is **a library** which means that the goal is not to have libtommath as nice and shiny DEB/RPM/DLL or whatever else packages. Our goal is (or IMO should be) that libtommath is widely used in final SW products by either linking or bundling or whatever.

We are not the most frequently used bignum library; however, our past track record is pretty good and many projects use us (some say it publicly, some not). We also have some "prominent" consumers like tcl, dropbear, rakudo/perl6 that give us feedback or even some coding effort. **We cannot ignore them**, but we do, we make their life more complicated, what is worse in some case it looks like we complicate their lives for no obvious reasons.

By our effort to `be consistent across platforms` we are just narrowing down the set of supported compilers and the set of supported platforms. According to my experience with my perl bindings to libtommath + libtomcrypt there are still a lot of people living out of GNU make + gcc/clang wonderland on a pretty obscure operating systems and exotic HW.

If we will continue with this approach we will end up in our bubble **totally rigorously consistent across platforms** but will be losing our users, our consumers.

minad:
>> This is not good, things should be consistent across platforms.

I still stand with this statement and would like to hear why it is wrong. I haven't seen an argument from your side beside arguing what other libraries do. The libraries you mentioned and also ltm have a lot of legacy code but the decisions taken back then are not necessarily the right ones forever.

If I want to write code which is portable across platforms I need guarantees regarding the size of types. The argument for these imprecise types is usually performance and we still have mp_set which can be used to set a single platform dependent digit.

Please read through the discussion in #285 and #278. Both @sjaeckel and @czurnieden voted for using precise types.

> Providing a least basic signed/unsigned long setter/getter is de facto standard.

Actually we provide mp_get/set_l/ul as macros. Therefore I am not against adding the functions @nijtmans proposed in this PR. However I am not fond of compromising in the light of goals I consider questionable, like avoiding stdint.h.

> Our goal is (or IMO should be) that libtommath is widely used in final SW products by either linking or bundling or whatever.

I agree. But a design discussion should be allowed, which takes the goals into consideration. During such a discussion the maintainers decide which API to provide or not. Furthermore things are not frozen in time and there must be the possibility to improve and remove bad functions.

The design discussion should be driven by technical arguments. You refuse to argue about stdint.h. A missing stdint.h can be fixed trivially by providing your own header. This will be a much better solution than doing search and replace on the code base as in your no-stdint-h branch.

> what is worse in some case it looks like we complicate their lives for no obvious reasons.

Please show me those cases. If they seem questionable I can probably tell you the reason. If I don't manage to do so on technical grounds @sjaeckel can revert the changes.

karel-m:
I am tired. I am leaving.

nijtmans:
@karel-m , Before you leave, at least put your approval here. From your remarks I conclude you approve ;-)
 
@minad wrote:
>If I want to write code which is portable across platforms I need guarantees regarding the size of types.
> The argument for these imprecise types is usually performance

Well, Tcl uses "long long" for all calculations, and everything below LLONG_MIN or above LLONG_MAX is handled over to libtommath.   So, you are telling me that Tcl is not portable across platforms ???   Then you have been in another school than I have been ....

The only function that Tcl uses for setting mp_int's is mp_set_long_long, which is bad because this function expects an unsigned long long: I agree with it's deprecation.   But I don't agree with using mp_set_i64 in stead: There is no guarantee that long long is 64 bits, the standard says that long long is _at least_  64 bits. libtommath still lacks a suitable replacement function in my view. It would mean that I am forced to change Tcl's limit checkout to hardcoded 64-bit values in stead of LLONG_MIN/LLONG_MAX.  I'm not prepared to do that, and I'm sure I'm not the only one.

Please hear me!  I'm not trying to remove or disable or discourage the xxx_i32/64 functions. They simply are not sufficient for everyone.  And the following macro's are not what I learned at school. It's as bad as the ##w## constructs used earlier in the macro's (thank God they are gone now).  Please consider putting the REAL signatures of functions in tommath.h, and implement them as documented.  And provide a long long variant aside the long variant, so Tcl doesn't have to invent its own weel. That's what this PR is about IMHO.

/* get integer, set integer (long) */
#define mp_get_l(a)        (sizeof (long) == 8 ? (long)mp_get_i64(a) : (long)mp_get_i32(a))
#define mp_set_l(a, b)     (sizeof (long) == 8 ? mp_set_i64((a), (b)) : mp_set_i32((a), (int32_t)(b)))

/* get integer, set integer (unsigned long) */
#define mp_get_ul(a)       (sizeof (long) == 8 ? (unsigned long)mp_get_u64(a) : (unsigned long)mp_get_u32(a))
#define mp_set_ul(a, b)    (sizeof (long) == 8 ? mp_set_u64((a), (b)) : mp_set_u32((a), (uint32_t)(b)))
#define mp_get_magl(a)     (sizeof (long) == 8 ? (unsigned long)mp_get_mag64(a) : (unsigned long)mp_get_mag32(a))

Enough techical arguments for now.    ;-)

minad:
Ok, as I said - I am not 100% opposed to adding those functions. Therefore I added the macros for long since long varies between 32 and 64 bit on common platforms. long long is 64 bit on all platforms I am aware of. Therefore the ll variants are not necessary. Given that I am not happy with having both ll and i64 variants of the functions. They are only needed if you take the standpoint that you want to avoid stdint.h under all circumstances.

These days it is good common practice to use stdint. And I don't think we should compromise in order to support no-stdint. We already did in #311 which is ok since macro readability improved, but duplicating apis is not good.

Regarding stdint:

* See for example misra 6.3
* See rationale of http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/inttypes.h.html#tag_13_19_06
* https://stackoverflow.com/questions/9834747/reasons-to-use-or-not-stdint

I think we should at this point agree on how to proceed. It could well be that we add more apis which rely on stdint.h what shall we do then? See for example #294.

Therefore please let's not work against each other and pull in different directions regarding the api. In the long run it is not feasible to have an api which is compatible with stdint and no-stdint.

@sjaeckel what do you think?

nijtmans:
Well, in case you ask:  I'm not interested in "short"/"int" variants of those functions, "i32", "i64" "long" and "long long" are more than enough for any practical use I can think of. I would approve (u)int128_t and (u)intmax_t variants, but that's stuff for another PR ;-)

minad:
@nijtmans my point here is mostly - how should we proceed in the light of the two incompatible opinions regarding stdint. Therefore I want to have this resolved now. Otherwise the discussion will come up again and again. It seems you want to somehow shape the api such that both an stdint and long variants are available of all functions. I don't find that satisfying. This leads to unnecessary duplication. How should we proceed in such cases in the future?

And I still don't understand your opposition in that point. I am all for reducing libc usage to keep this library minimal. But stdint is the most benign header in libc I know. It only contains very well defined things and it is small, can be easily provided by the one porting the library. Unfortunately you refuse to discuss this.

nijtmans:
>It seems you want to somehow shape the api such that both an stdint and long variants are available of all functions. I don't find that satisfying

Hm, I think you are refering to the PR about mp_expt_d and mp_n_root. Well, talking about exponents/roots: Tcl has a hardcoded limit of 28 bits here, because exponents higher than that simply cannot practially be calculated. The reason: it fits in a mp_digit.  So, in this case, there is really no need to go beyond 32-bits. I would go for (unsigned) 32-bit there, and - indeed - I would prefer an "unsigned int" parameter, you most likely prefer "uint32_t".  But that's a matter of taste: we should choose one or the other, we shouldn't duplicate. Using "long" would be overkill.  Does that answer your assumption?

minad:
@nijtmans I am not only referring to the expt PR, but to cases like this in general. `mp_set_u32` vs `mp_set_ul` as given in this PR is the same discussion. But let's now consider the expt case. Assuming we would add `mp_expt_u32` (or with an alternative name, I proposed plain `mp_expt` in #304), this function would then be unusable if you hide it in msvc6. It would work obviously in no-stdint-h branch if you do the `uint32_t->unsigned` substitution. So @karel-m's branch should always work. Since you agree that you don't want to duplicate APIs only since we cannot agree, how do you plan to proceed if for example #304 would be merged (but it is still questionable if that will happen)?

It seems we both would like to avoid

~~~ c
#ifndef no_stdint
mp_err mp_expt_u32(const mp_int*, uint32_t);
#endif 
mp_err mp_expt_u(const mp_int*, unsigned);
~~~

I would also like to avoid things like 

~~~ c
#ifdef no_stdint
mp_err mp_expt(const mp_int*, unsigned);
#else 
mp_err mp_expt(const mp_int*, uint32_t);
#endif
~~~

This particular `mp_set_ul` PR is only a mild case of this duplication. As I said multiple times now, I am not totally against it since long/ulong are nice to have given the 32/64 bit transition. But generally I would prefer if we just go with precise types where possible.

tommath.h
```diff
@@ -299,26 +299,35 @@ uint32_t mp_get_mag32(const mp_int *a) MP_WUR;
...
+long mp_get_l(const mp_int *a) MP_WUR;
+void mp_set_l(mp_int *a, long b);
 
 /* get integer, set integer (unsigned long) */
-#define mp_get_ul(a)       (sizeof (long) == 8 ? (unsigned long)mp_get_u64(a) : (unsigned long)mp_get_u32(a))
-#define mp_set_ul(a, b)    (sizeof (long) == 8 ? mp_set_u64((a), (b)) : mp_set_u32((a), (uint32_t)(b)))
-#define mp_get_magl(a)     (sizeof (long) == 8 ? (unsigned long)mp_get_mag64(a) : (unsigned long)mp_get_mag32(a))
+#define mp_get_ul(a) ((unsigned long)mp_get_l(a))
+void mp_set_ul(mp_int *a, unsigned long b);
+unsigned long mp_get_magl(const mp_int *a) MP_WUR;
```
minad: maybe put mp_get_magl under /* get magnitude */

tommath_private.h
```diff
@@ -277,10 +277,10 @@ MP_DEPRECATED(s_mp_reverse) void bn_reverse(unsigned char *s, int len);
         return res;                                                                    \
     }
 
-#define MP_GET_SIGNED(type, name, mag)                        \
+#define MP_GET_SIGNED(name, mag, type, utype)                 \
     type name(const mp_int* a)                                \
     {                                                         \
-        uint64_t res = mag(a);                                \
+        utype res = mag(a);                                   \
```
minad: :+1:

minad:
@nijtmans One more thing I was thinking about - would it make sense to define mp_set_l or mp_set_i64 as defined in the macro in tommath.h as it is now?

~~~ c
void mp_set_l(mp_int* a, long b) {
    if (sizeof (long) == sizeof (int64_t))
        mp_set_i64(a, b);
    else
        ...
}
~~~

This way the symbol mp_set_l would effectively result in an alias/single jmp to the other function but would still be available as separate symbol. Alternatively we could play with `__attribute__((alias))/__declspec` but this would be very non portable. Hmm...

But maybe what you are proposing is more reasonable since we have these MP_SETX macros.

tommath.h
nijtmans: Done now. Thanks!

tommath_private.h
nijtmans: Thanks!

nijtmans:
ping @sjaeckel .   I think this one is ready to be merged. @karel-m  do you agree?

tommath.h
```diff
@@ -297,28 +297,41 @@ mp_err mp_init_u64(mp_int *a, uint64_t b) MP_WUR;
 /* get magnitude */
 uint32_t mp_get_mag32(const mp_int *a) MP_WUR;
 uint64_t mp_get_mag64(const mp_int *a) MP_WUR;
+unsigned long mp_get_magl(const mp_int *a) MP_WUR;
```
sjaeckel: It's clear that the magnitude is always positive, but I still don't like the inconsistency of having the function suffixed by **l** but returning an `unsigned long` ... but `mp_get_magul` doesn't look good, so probably they could be renamed like that?

```
mp_get_mag_u32()
mp_get_mag_u64()
mp_get_mag_ul()
mp_get_mag_ull()
```

nijtmans: Understood:   I agree with renaming those functions. The mp_get_mag_u32/64 functions are already on "develop" so - actually - are out of scope for this PR.  Shouldn't we do this rename in a separate follow-up PR then?

nijtmans: I mean ....  mp_get_mag32/64 are already on develop, their renaming doesn't belong to this PR.

sjaeckel: Okay, I wouldn't have a problem to put them in this PR as well, but fine by me to open another PR for that change.

PR #323, opened 2019-06-13
Doc fixes: Using %lu/%ld formatter for int32_t is not portable across platforms

nijtmans:
Title says it all.


minad:
Ugly as hell but ok.

PR #324, opened 2019-06-13
Simplify test-cases, by using mp_set_l

nijtmans:
Also fix use of types in 2 other files


minad:
great with this #285 pays off a bit :) I didn't touch much of the test suite then.

PR #326, opened 2019-07-15
Import stddef in tommath.h

nomeata:
I am not a great expert on these matters, but at least here (compiling
for WebAssembly (`wasm32-unknown-unknown-wasm`), I need this, otherwise
`size_t` is not available.

### DIFF
--- a/tommath.h
+++ b/tommath.h
@@ -5,6 +5,7 @@
 #define BN_H_
 
 #include <stdint.h>
+#include <stddef.h>
 #include <limits.h>
 
 #ifdef LTM_NO_FILE


nijtmans:
Well, I don't think that's right:  size_t is used in tommath.h, so - at least - whatever header file provides it should be included at the top of tommath.h somewhere.

Second, the header providing size_t is <stddef.h>, so including that in tommath.h should be the 'correct' thing to do. IMHO.

nomeata:
Ah, thanks for educating me about `stddef.h`. That works indeed; updated this PR

PR #327, opened 2019-07-24
tommath.h: do not expose limits.h

minad:
If we make `MP_WARRAY` private we do not have to expose limits.h in the tommath.h header. I believe `MP_WARRAY` is never used by library consumers. We have deprecated it for now, after this change it would be only available in `tommath_private.h`.

### DIFF
--- a/etc/tune.c
+++ b/etc/tune.c
@@ -4,10 +4,8 @@
  */
 #include "../tommath.h"
 #include "../tommath_private.h"
-#include <stdint.h>
 #include <time.h>
 #include <inttypes.h>
-#include <limits.h>
 #include <errno.h>
 
 /*
--- a/tommath.h
+++ b/tommath.h
@@ -6,7 +6,6 @@
 
 #include <stdint.h>
 #include <stddef.h>
-#include <limits.h>
 
 #ifdef LTM_NO_FILE
 #  warning LTM_NO_FILE has been deprecated, use MP_NO_FILE.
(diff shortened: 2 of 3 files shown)


minad:
Grepping through the debian code base shows that MP_WARRAY appears only in vendored tommath code, but never in library user code.

https://codesearch.debian.net/search?q=MP_WARRAY&perpkg=1

nijtmans:
I fully agree! However, this can be improved a little bit:

See:
#define MP_WARRAY (int)(1uLL << (((CHAR_BIT * sizeof(private_mp_word)) - (2 * MP_DIGIT_BIT)) + 1))
It's not really usefull to shift a long long, and finally cast it to an (int).
Better:
#define MP_WARRAY (1 << (((CHAR_BIT * (int)sizeof(private_mp_word)) - (2 * MP_DIGIT_BIT)) + 1))

So: cast the result of the sizeof() operator to (int).

(I know, it already was like that, but if I propose to fix this in another PR then this one has to be rebased anyway)

minad:
I think we also have mp_sizeof_bits. Unfortunately I am on mobile right now. If you have time, @nijtmans feel free to improve and push to this PR.

minad:
@nijtmans I guess it is better now?

minad:
@sjaeckel Ok! Any plans on when you want to do the new release (1.2)? After that we could remove the deprecations and continue with 2.0? Would that be ok?

Issue #328, opened 2019-08-04
Faster mp_toradix?


MasterDuke17:
For background, I starting looking at this when a profile of a Perl 6 program that calculated Ackerman numbers showed that 90% of the time was spent stringifying a big Integer to print the result. I was using Rakudo with the MoarVM backend which uses libtommath under the hood (the stringifying was essentially just a call to `mp_toradix`, https://github.com/MoarVM/MoarVM/blob/master/src/math/bigintops.c#L1023). I looked at the implementation of `mp_toradix` and wondered if there was a faster algorithm. I started searching and found this thread http://code.activestate.com/lists/tcl-core/13692/. This is my rough proof-of-concept transliteration into C using libtommath functions https://gist.github.com/MasterDuke17/956999b54093258f4a7e616a00d95ac2 (it wasn't until after I finished that I noticed some Barrett reduction related functions in libtommath). This is significantly faster. On my machine `mp_toradix` of 2**500000 takes ~13s, but `Barrett_todigits` takes ~0.3s. I haven't extensively benchmarked small and large values to see if using Barrett only makes sense after a certain size, but it definitely seems worthwhile at large values. Would there be any interest in using this for libtommath? Either as the implementation of `mp_toradix`, or maybe an alternative function?

czurnieden:
Yes, the `mp_toradix` function is *really* slow, that's for sure! ;-)

We are doing a lot of refactoring now to bring LibTomMath up-to-date (and to get rid of a lot of legacy stuff that is definitely gone obsolete by now). Speeding things up is next (some things are already done, e.g.: faster "fast multiplication", a usable nth-root and others). I don't know how the rest thinks about it but a simple divide&conquer algorithm for `to_radix` would do (Schoenhage was published it first, I think?) but that needs fast division. All that is already written (I have it in my own version for a couple of years now, i.e.: tried and tested) it just needs a "port" to the current LTM version.
Advantage of that D&C algorithm: very simple and starts to be faster at low values (I have the cut-off at 600 bits).

Another problem: `mp_radix_size` is O(n^2) instead of O(1). Look at the current code and you know why ;-)
Replacement has been written ([table based, i.e.: O(1)](https://github.com/czurnieden/libtommath/blob/spaceforspeed/bn_mp_radix_size.c) ) and fully tested, it just needs to be "ported".
[...]

PR #330, opened 2019-08-23
Add a new mp_todec_fast...

MasterDuke17:
that uses Barrett reduction to speed up stringifying large integers.

Some benchmarking, done with a lightly modified `demo/timing.c`:
```
CLK_PER_SEC == 3600313344
mp_torad of 2**1 =>  22224156/sec,       162 cycles, 0.0000 seconds
mp_todec of 2**1 =>   1492667/sec,      2412 cycles, 0.0000 seconds
mp_torad of 13**1 =>  15385954/sec,       234 cycles, 0.0000 seconds
mp_todec of 13**1 =>   1503890/sec,      2394 cycles, 0.0000 seconds

mp_torad of 2**2 =>  22224156/sec,       162 cycles, 0.0000 seconds
mp_todec of 2**2 =>   1503890/sec,      2394 cycles, 0.0000 seconds
mp_torad of 13**2 =>  11765729/sec,       306 cycles, 0.0000 seconds
mp_todec of 13**2 =>   1459981/sec,      2466 cycles, 0.0000 seconds

mp_torad of 2**4 =>  14286957/sec,       252 cycles, 0.0000 seconds
mp_todec of 2**4 =>   1459981/sec,      2466 cycles, 0.0000 seconds
mp_torad of 13**4 =>   7143478/sec,       504 cycles, 0.0000 seconds
mp_todec of 13**4 =>   1273996/sec,      2826 cycles, 0.0000 seconds

mp_torad of 2**8 =>  11112078/sec,       324 cycles, 0.0000 seconds
mp_todec of 2**8 =>   1438974/sec,      2502 cycles, 0.0000 seconds
mp_torad of 13**8 =>   4348204/sec,       828 cycles, 0.0000 seconds
mp_todec of 13**8 =>    639033/sec,      5634 cycles, 0.0000 seconds

mp_torad of 2**16 =>   7143478/sec,       504 cycles, 0.0000 seconds
mp_todec of 2**16 =>   1242344/sec,      2898 cycles, 0.0000 seconds
mp_torad of 13**16 =>   2325783/sec,      1548 cycles, 0.0000 seconds
mp_todec of 13**16 =>    381712/sec,      9432 cycles, 0.0000 seconds

mp_torad of 2**32 =>   3704026/sec,       972 cycles, 0.0000 seconds
mp_todec of 2**32 =>    617337/sec,      5832 cycles, 0.0000 seconds
mp_torad of 13**32 =>   1000087/sec,      3600 cycles, 0.0000 seconds
mp_todec of 13**32 =>    216469/sec,     16632 cycles, 0.0000 seconds

mp_torad of 2**64 =>   2040993/sec,      1764 cycles, 0.0000 seconds
mp_todec of 2**64 =>    347856/sec,     10350 cycles, 0.0000 seconds
mp_torad of 13**64 =>    376680/sec,      9558 cycles, 0.0000 seconds
mp_todec of 13**64 =>    115952/sec,     31050 cycles, 0.0000 seconds

mp_torad of 2**128 =>    869640/sec,      4140 cycles, 0.0000 seconds
mp_todec of 2**128 =>    197255/sec,     18252 cycles, 0.0000 seconds
mp_torad of 13**128 =>    122484/sec,     29394 cycles, 0.0000 seconds
mp_todec of 13**128 =>     59778/sec,     60228 cycles, 0.0000 seconds

mp_torad of 2**256 =>    336729/sec,     10692 cycles, 0.0000 seconds
mp_todec of 2**256 =>    103851/sec,     34668 cycles, 0.0000 seconds
mp_torad of 13**256 =>     35017/sec,    102816 cycles, 0.0000 seconds
mp_todec of 13**256 =>     29042/sec,    123966 cycles, 0.0000 seconds

mp_torad of 2**512 =>    107305/sec,     33552 cycles, 0.0000 seconds
mp_todec of 2**512 =>     53494/sec,     67302 cycles, 0.0000 seconds
mp_torad of 13**512 =>      9245/sec,    389430 cycles, 0.0001 seconds
mp_todec of 13**512 =>     14733/sec,    244368 cycles, 0.0001 seconds

mp_torad of 2**1024 =>     30425/sec,    118332 cycles, 0.0000 seconds
mp_todec of 2**1024 =>     52900/sec,     68058 cycles, 0.0000 seconds
mp_torad of 13**1024 =>      4827/sec,    745865 cycles, 0.0002 seconds
mp_todec of 13**1024 =>     13844/sec,    260046 cycles, 0.0001 seconds

mp_torad of 2**2048 =>     15603/sec,    230742 cycles, 0.0001 seconds
mp_todec of 2**2048 =>     26176/sec,    137538 cycles, 0.0000 seconds
mp_torad of 13**2048 =>      1269/sec,   2835144 cycles, 0.0008 seconds
mp_todec of 13**2048 =>      6411/sec,    561582 cycles, 0.0002 seconds

mp_torad of 2**4096 =>      4148/sec,    867797 cycles, 0.0002 seconds
mp_todec of 2**4096 =>     12749/sec,    282384 cycles, 0.0001 seconds
mp_torad of 13**4096 =>       323/sec,  11136150 cycles, 0.0031 seconds
mp_todec of 13**4096 =>      2852/sec,   1262358 cycles, 0.0004 seconds

mp_torad of 2**8192 =>      1081/sec,   3328344 cycles, 0.0009 seconds
mp_todec of 2**8192 =>      5991/sec,    600948 cycles, 0.0002 seconds
mp_torad of 13**8192 =>        81/sec,  44357363 cycles, 0.0123 seconds
mp_todec of 13**8192 =>      1198/sec,   3004920 cycles, 0.0008 seconds

mp_torad of 2**16384 =>       274/sec,  13116996 cycles, 0.0036 seconds
mp_todec of 2**16384 =>      2677/sec,   1344762 cycles, 0.0004 seconds
mp_torad of 13**16384 =>        20/sec, 176360310 cycles, 0.0490 seconds
mp_todec of 13**16384 =>       474/sec,   7579854 cycles, 0.0021 seconds

mp_torad of 2**32768 =>        69/sec,  51832746 cycles, 0.0144 seconds
mp_todec of 2**32768 =>      1146/sec,   3139524 cycles, 0.0009 seconds
mp_torad of 13**32768 =>         5/sec, 702303642 cycles, 0.1951 seconds
mp_todec of 13**32768 =>       176/sec,  20395908 cycles, 0.0057 seconds

mp_torad of 2**65536 =>        17/sec, 205633044 cycles, 0.0571 seconds
mp_todec of 2**65536 =>       456/sec,   7885224 cycles, 0.0022 seconds
mp_torad of 13**65536 =>         1/sec, 2803178970 cycles, 0.7786 seconds
mp_todec of 13**65536 =>        57/sec,  62507790 cycles, 0.0174 seconds

mp_torad of 2**131072 =>         4/sec, 820253700 cycles, 0.2278 seconds
mp_todec of 2**131072 =>       165/sec,  21787452 cycles, 0.0061 seconds
mp_torad of 13**131072 =>         0/sec, 11200695156 cycles, 3.1110 seconds
mp_todec of 13**131072 =>        17/sec, 200245194 cycles, 0.0556 seconds

mp_torad of 2**262144 =>         1/sec, 3273670044 cycles, 0.9093 seconds
mp_todec of 2**262144 =>        53/sec,  67087368 cycles, 0.0186 seconds
mp_torad of 13**262144 =>         0/sec, 44720842806 cycles, 12.4214 seconds
mp_todec of 13**262144 =>         5/sec, 661838760 cycles, 0.1838 seconds

mp_torad of 2**524288 =>         0/sec, 13101782489 cycles, 3.6391 seconds
mp_todec of 2**524288 =>        16/sec, 217065492 cycles, 0.0603 seconds
mp_torad of 13**524288 =>         0/sec, 177629371344 cycles, 49.3372 seconds
mp_todec of 13**524288 =>         1/sec, 2334986874 cycles, 0.6486 seconds
```

### DIFF
--- a/demo/test.c
+++ b/demo/test.c
@@ -50,6 +50,49 @@ static uint64_t uabs64(int64_t x)
    return x > 0 ? (uint64_t)x : -(uint64_t)x;
 }
 
+/* This function prototype is needed
+ * to test dead code elimination
+ * which is used for feature detection.
+ *
+ * If the feature detection does not
+ * work as desired we will get a linker error.
+ */
+void does_not_exist(void);
+
+static int test_feature_detection(void)
+{
+#define BN_TEST_FEATURE1_C
+   if (!MP_HAS(TEST_FEATURE1)) {
+      does_not_exist();
+      return EXIT_FAILURE;
+   }
+
+#define BN_TEST_FEATURE2_C 1
+   if (MP_HAS(TEST_FEATURE2)) {
+      does_not_exist();
+      return EXIT_FAILURE;
+   }
+
+#define BN_TEST_FEATURE3_C 0
+   if (MP_HAS(TEST_FEATURE3)) {
+      does_not_exist();
+      return EXIT_FAILURE;
+   }
+
+#define BN_TEST_FEATURE4_C something
+   if (MP_HAS(TEST_FEATURE4)) {
+      does_not_exist();
+      return EXIT_FAILURE;
+   }
+
+   if (MP_HAS(TEST_FEATURE5)) {
+      does_not_exist();
+      return EXIT_FAILURE;
+   }
+
+   return EXIT_SUCCESS;
+}
+
 static int test_trivial_stuff(void)
 {
    mp_int a, b, c, d;
@@ -2046,6 +2089,7 @@ int unit_tests(int argc, char **argv)
       int (*fn)(void);
    } test[] = {
 #define T(n) { #n, test_##n }
+      T(feature_detection),
       T(trivial_stuff),
       T(mp_get_set_i32),
       T(mp_get_set_i64),
--- a/tommath.h
+++ b/tommath.h
@@ -702,6 +702,7 @@ mp_err mp_read_radix(mp_int *a, const char *str, int radix) MP_WUR;
 mp_err mp_toradix(const mp_int *a, char *str, int radix) MP_WUR;
 mp_err mp_toradix_n(const mp_int *a, char *str, int radix, int maxlen) MP_WUR;
 mp_err mp_radix_size(const mp_int *a, int radix, int *size) MP_WUR;
+mp_err mp_todecimal(const mp_int *a, char *str) MP_WUR;
 
 #ifndef MP_NO_FILE
 mp_err mp_fread(mp_int *a, int radix, FILE *stream) MP_WUR;
@@ -717,7 +718,6 @@ mp_err mp_fwrite(const mp_int *a, int radix, FILE *stream) MP_WUR;
 
 #define mp_tobinary(M, S)  mp_toradix((M), (S), 2)
 #define mp_tooctal(M, S)   mp_toradix((M), (S), 8)
-#define mp_todecimal(M, S) mp_toradix((M), (S), 10)
 #define mp_tohex(M, S)     mp_toradix((M), (S), 16)
 
 #ifdef __cplusplus
--- a/tommath.h
+++ b/tommath.h
@@ -206,7 +206,7 @@ TOOM_SQR_CUTOFF;
 #  endif
 #endif
 
-#if defined(__GNUC__) && (__GNUC__ * 100 + __GNUC_MINOR__ >= 301)
+#if defined(__GNUC__) && (__GNUC__ * 100 + __GNUC_MINOR__ >= 405)
 #  define MP_DEPRECATED(x) __attribute__((deprecated("replaced by " #x)))
 #  define PRIVATE_MP_DEPRECATED_PRAGMA(s) _Pragma(#s)
 #  define MP_DEPRECATED_PRAGMA(s) PRIVATE_MP_DEPRECATED_PRAGMA(GCC warning s)
--- a/bn_mp_div_d.c
+++ b/bn_mp_div_d.c
@@ -43,12 +43,10 @@ mp_err mp_div_d(const mp_int *a, mp_digit b, mp_int *c, mp_digit *d)
       return MP_OKAY;
    }
 
-#ifdef BN_MP_DIV_3_C
    /* three? */
-   if (b == 3u) {
+   if (MP_HAS(MP_DIV_3) && b == 3u) {
       return mp_div_3(a, c, d);
    }
-#endif
 
    /* no easy answer [c'est la vie].  Just division */
    if ((err = mp_init_size(&q, a->used)) != MP_OKAY) {
--- a/bn_mp_exptmod.c
+++ b/bn_mp_exptmod.c
@@ -19,9 +19,12 @@ mp_err mp_exptmod(const mp_int *G, const mp_int *X, const mp_int *P, mp_int *Y)
 
    /* if exponent X is negative we have to recurse */
    if (X->sign == MP_NEG) {
-#ifdef BN_MP_INVMOD_C
       mp_int tmpG, tmpX;
       mp_err err;
+
+      if (!MP_HAS(MP_INVMOD)) {
+         return MP_VAL;
+      }
 
       /* first compute 1/G mod P */
       if ((err = mp_init(&tmpG)) != MP_OKAY) {
@@ -46,50 +49,32 @@ mp_err mp_exptmod(const mp_int *G, const mp_int *X, const mp_int *P, mp_int *Y)
       err = mp_exptmod(&tmpG, &tmpX, P, Y);
       mp_clear_multi(&tmpG, &tmpX, NULL);
       return err;
-#else
-      /* no invmod */
-      return MP_VAL;
-#endif
    }
 
    /* modified diminished radix reduction */
-#if defined(BN_MP_REDUCE_IS_2K_L_C) && defined(BN_MP_REDUCE_2K_L_C) && defined(BN_S_MP_EXPTMOD_C)
-   if (mp_reduce_is_2k_l(P) == MP_YES) {
+   if (MP_HAS(MP_REDUCE_IS_2K_L) && MP_HAS(MP_REDUCE_2K_L) && MP_HAS(S_MP_EXPTMOD) &&
+       mp_reduce_is_2k_l(P) == MP_YES) {
       return s_mp_exptmod(G, X, P, Y, 1);
    }
-#endif
 
-#ifdef BN_MP_DR_IS_MODULUS_C
-   /* is it a DR modulus? */
-   dr = (mp_dr_is_modulus(P) == MP_YES) ? 1 : 0;
-#else
-   /* default to no */
-   dr = 0;
-#endif
+   /* is it a DR modulus? default to no */
+   dr = MP_HAS(MP_DR_IS_MODULUS) && mp_dr_is_modulus(P) == MP_YES ? 1 : 0;
 
-#ifdef BN_MP_REDUCE_IS_2K_C
    /* if not, is it a unrestricted DR modulus? */
-   if (dr == 0) {
+   if (MP_HAS(MP_REDUCE_IS_2K) && dr == 0) {
       dr = (mp_reduce_is_2k(P) == MP_YES) ? 2 : 0;
    }
-#endif
 
    /* if the modulus is odd or dr != 0 use the montgomery method */
-#ifdef BN_S_MP_EXPTMOD_FAST_C
-   if (MP_IS_ODD(P) || (dr !=  0)) {
+   if (MP_HAS(S_MP_EXPTMOD_FAST) && (MP_IS_ODD(P) || (dr != 0))) {
       return s_mp_exptmod_fast(G, X, P, Y, dr);
-   } else {
-#endif
-#ifdef BN_S_MP_EXPTMOD_C
+   } else if (MP_HAS(S_MP_EXPTMOD)) {
       /* otherwise use the generic Barrett reduction technique */
       return s_mp_exptmod(G, X, P, Y, 0);
-#else
+   } else {
       /* no exptmod for evens */
       return MP_VAL;
-#endif
-#ifdef BN_S_MP_EXPTMOD_FAST_C
    }
-#endif
 }
 
 #endif
--- a/bn_mp_invmod.c
+++ b/bn_mp_invmod.c
@@ -11,17 +11,13 @@ mp_err mp_invmod(const mp_int *a, const mp_int *b, mp_int *c)
       return MP_VAL;
    }
 
-#ifdef BN_S_MP_INVMOD_FAST_C
    /* if the modulus is odd we can use a faster routine instead */
-   if (MP_IS_ODD(b)) {
+   if (MP_HAS(S_MP_INVMOD_FAST) && MP_IS_ODD(b)) {
       return s_mp_invmod_fast(a, b, c);
    }
-#endif
 
-#ifdef BN_S_MP_INVMOD_SLOW_C
-   return s_mp_invmod_slow(a, b, c);
-#else
-   return MP_VAL;
-#endif
+   return MP_HAS(S_MP_INVMOD_SLOW)
+          ? s_mp_invmod_slow(a, b, c)
+          : MP_VAL;
 }
 #endif
--- a/bn_mp_mul.c
+++ b/bn_mp_mul.c
@@ -6,80 +6,46 @@
 /* high level multiplication (handles sign) */
 mp_err mp_mul(const mp_int *a, const mp_int *b, mp_int *c)
 {
-   mp_err  err;
-   mp_sign neg;
-#ifdef BN_S_MP_BALANCE_MUL_C
-   int len_b, len_a;
-#endif
-   neg = (a->sign == b->sign) ? MP_ZPOS : MP_NEG;
-#ifdef BN_S_MP_BALANCE_MUL_C
-   len_a = a->used;
-   len_b = b->used;
+   mp_err err;
+   int min_len = MP_MIN(a->used, b->used),
+       max_len = MP_MAX(a->used, b->used),
+       digs = a->used + b->used + 1;
+   mp_sign neg = (a->sign == b->sign) ? MP_ZPOS : MP_NEG;
 
-   if (len_a == len_b) {
-      goto GO_ON;
-   }
-   /*
-    * Check sizes. The smaller one needs to be larger than the Karatsuba cut-off.
-    * The bigger one needs to be at least about one KARATSUBA_MUL_CUTOFF bigger
-    * to make some sense, but it depends on architecture, OS, position of the
-    * stars... so YMMV.
-    * Using it to cut the input into slices small enough for fast_s_mp_mul_digs
-    * was actually slower on the author's machine, but YMMV.
-    */
-   if ((MP_MIN(len_a, len_b) < MP_KARATSUBA_MUL_CUTOFF)
-       || ((MP_MAX(len_a, len_b) / 2) < MP_KARATSUBA_MUL_CUTOFF)) {
-      goto GO_ON;
-   }
-   /*
-    * Not much effect was observed below a ratio of 1:2, but again: YMMV.
-    */
-   if ((MP_MAX(len_a, len_b) /  MP_MIN(len_a, len_b)) < 2) {
-      goto GO_ON;
-   }
-
-   err = s_mp_balance_mul(a,b,c);
-   goto END;
-
-GO_ON:
-#endif
-
-   /* use Toom-Cook? */
-#ifdef BN_S_MP_TOOM_MUL_C
-   if (MP_MIN(a->used, b->used) >= MP_TOOM_MUL_CUTOFF) {
+   if (MP_HAS(S_MP_BALANCE_MUL) &&
+       /* Check sizes. The smaller one needs to be larger than the Karatsuba cut-off.
+        * The bigger one needs to be at least about one MP_KARATSUBA_MUL_CUTOFF bigger
+        * to make some sense, but it depends on architecture, OS, position of the
+        * stars... so YMMV.
+        * Using it to cut the input into slices small enough for fast_s_mp_mul_digs
+        * was actually slower on the author's machine, but YMMV.
+        */
+       (min_len >= MP_KARATSUBA_MUL_CUTOFF) &&
+       (max_len / 2 >= MP_KARATSUBA_MUL_CUTOFF) &&
+       /* Not much effect was observed below a ratio of 1:2, but again: YMMV. */
+       (max_len >= (2 * min_len))) {
+      err = s_mp_balance_mul(a,b,c);
+   } else if (MP_HAS(S_MP_TOOM_MUL) &&
+              (min_len >= MP_TOOM_MUL_CUTOFF)) {
       err = s_mp_toom_mul(a, b, c);
-   } else
-#endif
-#ifdef BN_S_MP_KARATSUBA_MUL_C
-      /* use Karatsuba? */
-      if (MP_MIN(a->used, b->used) >= MP_KARATSUBA_MUL_CUTOFF) {
-         err = s_mp_karatsuba_mul(a, b, c);
-      } else
-#endif
-      {
-         /* can we use the fast multiplier?
-          *
-          * The fast multiplier can be used if the output will
-          * have less than MP_WARRAY digits and the number of
-          * digits won't affect carry propagation
-          */
-         int     digs = a->used + b->used + 1;
-
-#ifdef BN_S_MP_MUL_DIGS_FAST_C
-         if ((digs < MP_WARRAY) &&
-             (MP_MIN(a->used, b->used) <= MP_MAXFAST)) {
-            err = s_mp_mul_digs_fast(a, b, c, digs);
-         } else
-#endif
-         {
-#ifdef BN_S_MP_MUL_DIGS_C
-            err = s_mp_mul_digs(a, b, c, a->used + b->used + 1);
-#else
-            err = MP_VAL;
-#endif
-         }
-      }
-END:
+   } else if (MP_HAS(S_MP_KARATSUBA_MUL) &&
+              (min_len >= MP_KARATSUBA_MUL_CUTOFF)) {
+      err = s_mp_karatsuba_mul(a, b, c);
+   } else if (MP_HAS(S_MP_MUL_DIGS_FAST) &&
+              /* can we use the fast multiplier?
+               *
+               * The fast multiplier can be used if the output will
+               * have less than MP_WARRAY digits and the number of
+               * digits won't affect carry propagation
+               */
+              (digs < MP_WARRAY) &&
+              (min_len <= MP_MAXFAST)) {
+      err = s_mp_mul_digs_fast(a, b, c, digs);
+   } else if (MP_HAS(S_MP_MUL_DIGS)) {
+      err = s_mp_mul_digs(a, b, c, digs);
+   } else {
+      err = MP_VAL;
+   }
    c->sign = (c->used > 0) ? neg : MP_ZPOS;
    return err;
 }
--- a/bn_mp_reduce.c
+++ b/bn_mp_reduce.c
@@ -26,21 +26,17 @@ mp_err mp_reduce(mp_int *x, const mp_int *m, const mp_int *mu)
       if ((err = mp_mul(&q, mu, &q)) != MP_OKAY) {
          goto CLEANUP;
       }
-   } else {
-#ifdef BN_S_MP_MUL_HIGH_DIGS_C
+   } else if (MP_HAS(S_MP_MUL_HIGH_DIGS)) {
       if ((err = s_mp_mul_high_digs(&q, mu, &q, um)) != MP_OKAY) {
          goto CLEANUP;
       }
-#elif defined(BN_S_MP_MUL_HIGH_DIGS_FAST_C)
+   } else if (MP_HAS(S_MP_MUL_HIGH_DIGS_FAST)) {
       if ((err = s_mp_mul_high_digs_fast(&q, mu, &q, um)) != MP_OKAY) {
          goto CLEANUP;
       }
-#else
-      {
-         err = MP_VAL;
-         goto CLEANUP;
-      }
-#endif
+   } else {
+      err = MP_VAL;
+      goto CLEANUP;
    }
 
    /* q3 = q2 / b**(k+1) */
--- a/bn_mp_sqr.c
+++ b/bn_mp_sqr.c
@@ -7,35 +7,21 @@
 mp_err mp_sqr(const mp_int *a, mp_int *b)
 {
    mp_err err;
-
-#ifdef BN_S_MP_TOOM_SQR_C
-   /* use Toom-Cook? */
-   if (a->used >= MP_TOOM_SQR_CUTOFF) {
+   if (MP_HAS(S_MP_TOOM_SQR) && /* use Toom-Cook? */
+       a->used >= MP_TOOM_SQR_CUTOFF) {
       err = s_mp_toom_sqr(a, b);
-      /* Karatsuba? */
-   } else
-#endif
-#ifdef BN_S_MP_KARATSUBA_SQR_C
-      if (a->used >= MP_KARATSUBA_SQR_CUTOFF) {
-         err = s_mp_karatsuba_sqr(a, b);
-      } else
-#endif
-      {
-#ifdef BN_S_MP_SQR_FAST_C
-         /* can we use the fast comba multiplier? */
-         if ((((a->used * 2) + 1) < MP_WARRAY) &&
-             (a->used < (MP_MAXFAST / 2))) {
-            err = s_mp_sqr_fast(a, b);
-         } else
-#endif
-         {
-#ifdef BN_S_MP_SQR_C
-            err = s_mp_sqr(a, b);
-#else
-            err = MP_VAL;
-#endif
-         }
-      }
+   } else if (MP_HAS(S_MP_KARATSUBA_SQR) &&  /* Karatsuba? */
+              a->used >= MP_KARATSUBA_SQR_CUTOFF) {
+      err = s_mp_karatsuba_sqr(a, b);
+   } else if (MP_HAS(S_MP_SQR_FAST) && /* can we use the fast comba multiplier? */
+              (((a->used * 2) + 1) < MP_WARRAY) &&
+              (a->used < (MP_MAXFAST / 2))) {
+      err = s_mp_sqr_fast(a, b);
+   } else if (MP_HAS(S_MP_SQR)) {
+      err = s_mp_sqr(a, b);
+   } else {
+      err = MP_VAL;
+   }
    b->sign = MP_ZPOS;
    return err;
 }
--- a/bn_mp_todecimal.c
+++ b/bn_mp_todecimal.c
@@ -1,0 +1,23 @@
+#include "tommath_private.h"
+#ifdef BN_MP_TODECIMAL_C
+/* LibTomMath, multiple-precision integer library -- Tom St Denis */
+/* SPDX-License-Identifier: Unlicense */
+
+/* stores a bignum as a decimal ASCII string, using Barrett
+ * reduction if available.
+ */
+
+mp_err mp_todecimal(const mp_int *a, char *str)
+{
+   mp_err err;
+
+   if (MP_HAS(S_MP_TODECIMAL_FAST)) {
+      err = s_mp_todecimal_fast(a, str);
+   } else {
+      err = mp_toradix(a, str, 10);
+   }
+
+   return err;
+}
+
+#endif
--- a/bn_mp_todecimal_fast.c
+++ b/bn_mp_todecimal_fast.c
@@ -1,6 +1,5 @@
 #include "tommath_private.h"
 #include <string.h>
-#include <stdio.h>
 #ifdef BN_MP_TODECIMAL_FAST_C
 /* LibTomMath, multiple-precision integer library -- Tom St Denis */
 /* SPDX-License-Identifier: Unlicense */
@@ -13,12 +12,9 @@ mp_err mp_todecimal_fast_rec(mp_int *number, mp_int *nL, mp_int *shiftL, mp_int 
    mp_err err;
 
    if (precalc_array_index < 0) {
-      char *next_piece = calloc(4, sizeof(char));
-      int s_s = left ? snprintf(next_piece, 4, "%u", mp_get_i32(number)) : snprintf(next_piece, 4, "%03u",
-                mp_get_i32(number));
-      int r_s = (int)strlen(*result);
-      (*result) = realloc(*result, (size_t)(r_s + s_s + 2));
-      strcat(*result, next_piece);
+      int new_pos = left ? snprintf(*result, 4, "%u", mp_get_i32(number)) : snprintf(*result, 4, "%03u",
+                    mp_get_i32(number));
+      *result += new_pos;
       return MP_OKAY;
    }
 
@@ -70,11 +66,12 @@ LBL_ERR:
    return err;
 }
 
-mp_err mp_todecimal_fast(mp_int *number, char **result)
+mp_err mp_todecimal_fast(mp_int *number, char *result)
 {
    mp_int n, shift, M, M2, M22, M4, M44;
    mp_int nL[20], shiftL[20], mL[20];
    mp_err err;
+   char **result_addr = &result;
    int precalc_array_index = 1;
 
    if ((err = mp_init_multi(&M2, &M22, &M4, &M44, NULL)) != MP_OKAY) {
@@ -85,7 +82,8 @@ mp_err mp_todecimal_fast(mp_int *number, char **result)
       if ((err = mp_neg(number, number)) != MP_OKAY) {
          goto LBL_ERR;
       }
-      *result[0] = '-';
+      result[0] = '-';
+      result_addr += 1;
    }
    if ((err = mp_init_set(&n, (mp_digit)1000)) != MP_OKAY) {
       goto LBL_ERR;
@@ -189,7 +187,7 @@ mp_err mp_todecimal_fast(mp_int *number, char **result)
       precalc_array_index++;
    }
 
-   if ((err = mp_todecimal_fast_rec(number, nL, shiftL, mL, precalc_array_index - 1, 1, result)) != MP_OKAY) {
+   if ((err = mp_todecimal_fast_rec(number, nL, shiftL, mL, precalc_array_index - 1, 1, result_addr)) != MP_OKAY) {
       goto LBL_ERR;
    }
 
--- a/bn_mp_todecimal_fast.c
+++ b/bn_mp_todecimal_fast.c
@@ -1,0 +1,199 @@
+#include "tommath_private.h"
+#include <string.h>
+#ifdef BN_MP_TODECIMAL_FAST_C
+/* LibTomMath, multiple-precision integer library -- Tom St Denis */
+/* SPDX-License-Identifier: Unlicense */
+
+/* store a bignum as a decimal ASCII string */
+mp_err mp_todecimal_fast_rec(mp_int *number, mp_int *nL, mp_int *shiftL, mp_int *mL, int index, int left, char **result) {
+    mp_int q, nLq, r;
+    mp_err err;
+
+    if (index < 0) {
+        char *next_piece = calloc(4, sizeof(char));
+        int s_s = snprintf(next_piece, 4, left ? "%u" : "%03u", mp_get_u32(number));
+        int r_s = strlen(*result);
+        (*result) = realloc(*result, r_s + s_s + 2);
+        strcat(*result, next_piece);
+        return MP_OKAY;
+    }
+
+    if ((err = mp_init_multi(&q, &nLq, &r, NULL)) != MP_OKAY) {
+        goto LBL_ERR;
+    }
+    if ((err = mp_mul(number, &mL[index], &q)) != MP_OKAY) {
+        goto LBL_ERR;
+    }
+    if ((err = mp_div_2d(&q, mp_get_u32(&shiftL[index]), &q, NULL)) != MP_OKAY) {
+        goto LBL_ERR;
+    }
+
+    if ((err = mp_mul(&nL[index], &q, &nLq)) != MP_OKAY) {
+        goto LBL_ERR;
+    }
+
+    if ((err = mp_sub(number, &nLq, &r)) != MP_OKAY) {
+        goto LBL_ERR;
+    }
+
+    if (mp_isneg(&r)) {
+        if ((err = mp_sub_d(&q, 1, &q)) != MP_OKAY) {
+            goto LBL_ERR;
+        }
+        if ((err = mp_add(&r, &nL[index], &r)) != MP_OKAY) {
+            goto LBL_ERR;
+        }
+    }
+
+    --index;
+    if (left && mp_iszero(&q)) {
+        if ((err = mp_todecimal_fast_rec(&r, nL, shiftL, mL, index, 1, result)) != MP_OKAY) {
+            goto LBL_ERR;
+        }
+    } else {
+        if ((err = mp_todecimal_fast_rec(&q, nL, shiftL, mL, index, left, result)) != MP_OKAY) {
+            goto LBL_ERR;
+        }
+        if ((err = mp_todecimal_fast_rec(&r, nL, shiftL, mL, index, 0, result)) != MP_OKAY) {
+            goto LBL_ERR;
+        }
+    }
+
+    err = MP_OKAY;
+
+LBL_ERR:mp_clear_multi (&q, &nLq, &r, NULL);
+    return err;
+}
+
+mp_err mp_todecimal_fast(mp_int *number, char **result) {
+    mp_int n, shift, M, M2, M22, M4, M44;
+    mp_int *nL, *shiftL, *mL;
+    mp_err err;
+    int index = 1;
+
+    if ((err = mp_init_multi(&M2, &M22, &M4, &M44, NULL)) != MP_OKAY) {
+        goto LBL_ERR;
+    }
+
+    if (mp_isneg(number)) {
+        if ((err = mp_neg(number, number)) != MP_OKAY) {
+            goto LBL_ERR;
+        }
+        *result[0] = '-';
+    }
+    if ((err = mp_init_set(&n, 1000)) != MP_OKAY) {
+        goto LBL_ERR;
+    }
+
+    nL = malloc(20 * sizeof(mp_int));
+    if ((err = mp_init_copy(&nL[0], &n)) != MP_OKAY) {
+        goto LBL_ERR;
+    }
+
+    if ((err = mp_init_set(&shift, 20)) != MP_OKAY) {
+        goto LBL_ERR;
+    }
+
+    shiftL = malloc(20 * sizeof(mp_int));
+    if ((err = mp_init_copy(&shiftL[0], &shift)) != MP_OKAY) {
+        goto LBL_ERR;
+    }
+
+    /* (8 * 2**$shift) / $n rounded up */
+    if ((err = mp_init_set(&M, 8389)) != MP_OKAY) {
+        goto LBL_ERR;
+    }
+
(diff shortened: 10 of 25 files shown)


MasterDuke17:
See #328

minad:
:+1:

Regarding the API I think it would be best if we replace the mp_todecimal macros by specialised functions. These could either branch to a fast variant if available or branch to the generic toradix. The mp_todecimal_fast function from this PR should then be internal and named s_mp_todecimal_fast

MasterDuke17:
Now that I’m thinking about it, a lot of functions have the name of the algorithm in them. Maybe make it internal and rename to s_mp_todecimal_barrett?

MasterDuke17:
> These could either branch to a fast variant if available or branch to the generic toradix.

What do you mean, "if available"? Should I change this PR to also change the `mp_todecimal` macro into a function that chooses `mp_toradix` under a certain size and chooses my new function above?

One difficulty is that `mp_toradix` assume it's given a buffer of the necessary size, but my function grows the buffer because `mp_radix_size` is so slow. However, @czurnieden said they have a new `mp_radix_size` that's O(1) in the works, do we need to wait until that's merged?

minad:
If available - via the macro configuration, see #262

minad:
I would prefer if we use a radix size function and don't grow the buffer.

czurnieden:
Why is it called `todecimal`?

Is there a `from_decimal` available?

But after a quick first glimpse: looks good.
It needs some formatting and adaption to the style here and a test and documentation but otherwise&hellip;

> However, @czurnieden said they have a new mp_radix_size that's O(1) in the works, do we need to wait until that's merged?

[It](https://github.com/czurnieden/libtommath/blob/spaceforspeed/bn_mp_radix_size.c) is in a larger PR but independent of the rest of that PR. Needs to get updated to the new state of LibTomMath (more or less just the small_int_2_big_int parts, nothing significant) but no other hurdles as far as I can see.
As whined about in #328 I'm stuck up to the upper edge of my lower lip in work untill the end of August, so if I get a <del>kick in the a</del>thumbs-up from the regulars I'll squeeze it in this WE otherwise use `mp_ilogb` in the meantime.

MasterDuke17:
@minad 
> I would prefer if we use a radix size function and don't grow the buffer.

I can switch to assuming the passed in buffer is the correct size, but it'll take me a little bit of time to re-work.


@czurnieden 
> It needs some formatting and adaption to the style here

I tried to follow the style and it passes `make astyle`, what is wrong?

> a test and documentation

Yep I can work on that, but I wanted confirmation I was on the right path first, so thanks.

MasterDuke17:
A style/correctness question. Should I move the 3 arrays of `mp_int`s that are passed around outside the functions so they don't have to be passed in every call? So 
```
mp_int nL[20], shiftL[20], mL[20];

mp_err mp_todecimal_fast_rec(mp_int *number, int precalc_array_index, int left, char **result) {
...
}
```
instead of the current
```
mp_err mp_todecimal_fast_rec(mp_int *number, mp_int *nL, mp_int *shiftL, mp_int *mL, int precalc_array_index, int left, char **result) {
...
}

**Comment by MasterDuke17** [comment]
Timings with the new version that expects to be passed a pre-allocated buffer (i.e., similar API to `mp_toradix`. It got a bit faster, now the crossover point with `mp_toradix` is closer to 2^1k than 2^2k.
```
CLK_PER_SEC == 3600292536
mp_torad of 2**1 =>  22224028/sec,       162 cycles, 0.0000 seconds
mp_todec of 2**1 =>   1818329/sec,      1980 cycles, 0.0000 seconds
mp_torad of 13**1 =>  14286875/sec,       252 cycles, 0.0000 seconds
mp_todec of 13**1 =>   1785859/sec,      2016 cycles, 0.0000 seconds

mp_torad of 2**2 =>  16668021/sec,       216 cycles, 0.0000 seconds
mp_todec of 2**2 =>   1852002/sec,      1944 cycles, 0.0000 seconds
mp_torad of 13**2 =>  11112014/sec,       324 cycles, 0.0000 seconds
mp_todec of 13**2 =>   1801948/sec,      1998 cycles, 0.0000 seconds

mp_torad of 2**4 =>  14286875/sec,       252 cycles, 0.0000 seconds
mp_todec of 2**4 =>   1801948/sec,      1998 cycles, 0.0000 seconds
mp_torad of 13**4 =>   7143437/sec,       504 cycles, 0.0000 seconds
mp_todec of 13**4 =>   1613034/sec,      2232 cycles, 0.0000 seconds

mp_torad of 2**8 =>  11112014/sec,       324 cycles, 0.0000 seconds
mp_todec of 2**8 =>   1801948/sec,      1998 cycles, 0.0000 seconds
mp_torad of 13**8 =>   4444805/sec,       810 cycles, 0.0000 seconds
mp_todec of 13**8 =>    766345/sec,      4698 cycles, 0.0000 seconds

mp_torad of 2**16 =>   6452137/sec,       558 cycles, 0.0000 seconds
mp_todec of 2**16 =>   1600130/sec,      2250 cycles, 0.0000 seconds
mp_torad of 13**16 =>   2299037/sec,      1566 cycles, 0.0000 seconds
mp_todec of 13**16 =>    451503/sec,      7974 cycles, 0.0000 seconds

mp_torad of 2**32 =>   4000325/sec,       900 cycles, 0.0000 seconds
mp_todec of 2**32 =>    749124/sec,      4806 cycles, 0.0000 seconds
mp_torad of 13**32 =>   1015310/sec,      3546 cycles, 0.0000 seconds
mp_todec of 13**32 =>    256102/sec,     14058 cycles, 0.0000 seconds

mp_torad of 2**64 =>   2062023/sec,      1746 cycles, 0.0000 seconds
mp_todec of 2**64 =>    406537/sec,      8856 cycles, 0.0000 seconds
mp_torad of 13**64 =>    363665/sec,      9900 cycles, 0.0000 seconds
mp_todec of 13**64 =>    141154/sec,     25506 cycles, 0.0000 seconds

mp_torad of 2**128 =>    921733/sec,      3906 cycles, 0.0000 seconds
mp_todec of 2**128 =>    230965/sec,     15588 cycles, 0.0000 seconds
mp_torad of 13**128 =>    120419/sec,     29898 cycles, 0.0000 seconds
mp_todec of 13**128 =>     74744/sec,     48168 cycles, 0.0000 seconds

mp_torad of 2**256 =>    332805/sec,     10818 cycles, 0.0000 seconds
mp_todec of 2**256 =>    126512/sec,     28458 cycles, 0.0000 seconds
mp_torad of 13**256 =>     34955/sec,    102996 cycles, 0.0000 seconds
mp_todec of 13**256 =>     38875/sec,     92610 cycles, 0.0000 seconds

mp_torad of 2**512 =>    105940/sec,     33984 cycles, 0.0000 seconds
mp_todec of 2**512 =>     67459/sec,     53370 cycles, 0.0000 seconds
mp_torad of 13**512 =>      9089/sec,    396090 cycles, 0.0001 seconds
mp_todec of 13**512 =>     19411/sec,    185472 cycles, 0.0001 seconds

mp_torad of 2**1024 =>     30141/sec,    119448 cycles, 0.0000 seconds
mp_todec of 2**1024 =>     46150/sec,     78012 cycles, 0.0000 seconds
mp_torad of 13**1024 =>      4917/sec,    732096 cycles, 0.0002 seconds
mp_todec of 13**1024 =>     19351/sec,    186048 cycles, 0.0001 seconds

mp_torad of 2**2048 =>     15745/sec,    228654 cycles, 0.0001 seconds
mp_todec of 2**2048 =>     36241/sec,     99342 cycles, 0.0000 seconds
mp_torad of 13**2048 =>      1274/sec,   2824362 cycles, 0.0008 seconds
mp_todec of 13**2048 =>      8878/sec,    405504 cycles, 0.0001 seconds

mp_torad of 2**4096 =>      4170/sec,    863334 cycles, 0.0002 seconds
mp_todec of 2**4096 =>     17672/sec,    203724 cycles, 0.0001 seconds
mp_torad of 13**4096 =>       324/sec,  11106486 cycles, 0.0031 seconds
mp_todec of 13**4096 =>      3837/sec,    938070 cycles, 0.0003 seconds

mp_torad of 2**8192 =>      1085/sec,   3315510 cycles, 0.0009 seconds
mp_todec of 2**8192 =>      8256/sec,    436068 cycles, 0.0001 seconds
mp_torad of 13**8192 =>        82/sec,  43743563 cycles, 0.0122 seconds
mp_todec of 13**8192 =>      1577/sec,   2282616 cycles, 0.0006 seconds

mp_torad of 2**16384 =>       276/sec,  13006458 cycles, 0.0036 seconds
mp_todec of 2**16384 =>      3589/sec,   1003068 cycles, 0.0003 seconds
mp_torad of 13**16384 =>        20/sec, 174191328 cycles, 0.0484 seconds
mp_todec of 13**16384 =>       643/sec,   5597280 cycles, 0.0016 seconds

mp_torad of 2**32768 =>        70/sec,  51329502 cycles, 0.0143 seconds
mp_todec of 2**32768 =>      1555/sec,   2314188 cycles, 0.0006 seconds
mp_torad of 13**32768 =>         5/sec, 695193174 cycles, 0.1931 seconds
mp_todec of 13**32768 =>       254/sec,  14172930 cycles, 0.0039 seconds

mp_torad of 2**65536 =>        17/sec, 203966838 cycles, 0.0567 seconds
mp_todec of 2**65536 =>       637/sec,   5649137 cycles, 0.0016 seconds
mp_torad of 13**65536 =>         1/sec, 2778189192 cycles, 0.7717 seconds
mp_todec of 13**65536 =>        97/sec,  36809424 cycles, 0.0102 seconds

mp_torad of 2**131072 =>         4/sec, 813089286 cycles, 0.2258 seconds
mp_todec of 2**131072 =>       250/sec,  14385618 cycles, 0.0040 seconds
mp_torad of 13**131072 =>         0/sec, 11102668398 cycles, 3.0838 seconds
mp_todec of 13**131072 =>        35/sec, 100849896 cycles, 0.0280 seconds

mp_torad of 2**262144 =>         1/sec, 3254062661 cycles, 0.9038 seconds
mp_todec of 2**262144 =>        96/sec,  37234620 cycles, 0.0103 seconds
mp_torad of 13**262144 =>         0/sec, 44394883092 cycles, 12.3309 seconds
mp_todec of 13**262144 =>        13/sec, 272151000 cycles, 0.0756 seconds

mp_torad of 2**524288 =>         0/sec, 12977733078 cycles, 3.6046 seconds
mp_todec of 2**524288 =>        35/sec, 101212992 cycles, 0.0281 seconds
mp_torad of 13**524288 =>         0/sec, 177622493220 cycles, 49.3356 seconds
mp_todec of 13**524288 =>         4/sec, 737604720 cycles, 0.2049 seconds
```

**Comment by MasterDuke17** [comment]
@minad #262 hasn't been merged. Are you saying wait until it has? Or rebase this branch onto `feature-detection2` and use its `MP_HAS`?

**Comment by sjaeckel** [comment]
Thanks for this PR!

Sorry for blocking you guys, I will have more time from next week on

Yes, please rebase, squash together into logical commits and use MP_HAS

**Comment by sjaeckel** [comment]
... squash already happened ... :+1:

**Comment by minad** [comment]
Close in favor of #331

**Comment by sjaeckel** [comment]
Re-opened as #331 has its base to a non-existant branch now

**Comment by sjaeckel** [comment]
c.f. https://github.com/libtom/libtommath/pull/331#issuecomment-526922246

> Snprintf should not be used.

what else?

btw. I can't reproduce the error (yet)

c.f. https://github.com/libtom/libtommath/pull/331#issuecomment-526863425

> how does this look? 

looks fine IMO, I'll have a closer look soon

two things already now:
1. this function should get a `size_t maxlen` argument  (c.f. #332)
1. how about leaving `mp_todecimal` as a macro and calling `s_mp_todecimal_fast()` from `mp_to_radix()` if the `radix == 10`?


> I'm in the process of adding tests, but I'm not sure about documentation. How much is needed and where does it go?

c.f. docs: as it's no public API endpoint documentation is not really required, but if I could choose I'd really love to see an explanation on how the algorithm works :-) 

c.f. tests: as already proposed by @minad 'You can test against mp_toradix for random numbers.' which turned into a `...could test against..., but that one is deprecated now.` - probably we could test against `mp_read_radix()`?

**Comment by minad** [comment]
> Snprintf should not be used.

Instead you should convert the number to a string "by hand". It does not make sense to call out to a very expensive (heavyweight and slow) printf function here. I want to use tommath in scenarios without libc and in particular without printf/sprintf/snprintf. They are not available on tiny libcs on microcontrollers or if compiling to bare metal.

> how about leaving mp_todecimal as a macro and calling s_mp_todecimal_fast() from mp_to_radix() if the radix == 10?

Not possible since MP_HAS is only available privately and we should keep it like this.

>      c.f. docs: as it's no public API endpoint documentation is not really required, but if I could choose I'd really love to see an explanation on how the algorithm works :-)

:+1:

> c.f. tests: as already proposed by @minad 'You can test against mp_toradix for random numbers.' which turned into a '...could test against..., but that one is deprecated now.' - probably we could test against mp_read_radix()?

We still have mp_toradix and I guess we will keep it? Or we will have mp_to_radix as in #332. You can write the test against mp_toradix as I have proposed.

**Comment by sjaeckel** [comment]
> > how about leaving mp_todecimal as a macro and calling s_mp_todecimal_fast() from mp_to_radix() if the radix == 10?
> 
> Not possible since MP_HAS is only available privately and we should keep it like this.

I'm thinking about a code-path in the implementation of `mp_to_radix()` which would lead to not being able to test this function against `mp_to_radix()` as it would call itself... something like

```
$ git diff -U1
diff --git a/bn_mp_to_radix.c b/bn_mp_to_radix.c
index c75ee5b..2317dc4 100644
--- a/bn_mp_to_radix.c
+++ b/bn_mp_to_radix.c
@@ -29,2 +29,6 @@ mp_err mp_to_radix(const mp_int *a, char *str, size_t maxlen, int radix)
 
+   if (MP_HAS(S_MP_TODECIMAL_FAST) && radix == 10) {
+      return s_mp_todecimal_fast(a, str, maxlen);
+   }
+
    if ((err = mp_init_copy(&t, a)) != MP_OKAY) {
```

**Comment by sjaeckel** [comment]
> > Snprintf should not be used.
> 
> Instead you should convert the number to a string "by hand". It does not make sense to call out to a very expensive (heavyweight and slow) printf function here. I want to use tommath in scenarios without libc and in particular without printf/sprintf/snprintf. They are not available on microcontrollers or if compiling to bare metal.

fine by me

**Comment by MasterDuke17** [comment]
> > > Snprintf should not be used.
> > 
> > 
> > Instead you should convert the number to a string "by hand". It does not make sense to call out to a very expensive (heavyweight and slow) printf function here. I want to use tommath in scenarios without libc and in particular without printf/sprintf/snprintf. They are not available on microcontrollers or if compiling to bare metal.
> 
> fine by me

I have this implemented, but I'm currently trying to get this branch rebased onto the HEAD of develop. 

> c.f. docs: as it's no public API endpoint documentation is not really required, but if I could choose I'd really love to see an explanation on how the algorithm works :-)

Ha, I'm sorry, but I can't help there (other than just pointing to https://en.wikipedia.org/wiki/Barrett_reduction and hoping something there makes more sense to you than to me). I don't understand it myself, I just adopted/translated some code.

>     1. this function should get a `size_t maxlen` argument  (c.f. #332)

I'll work on that once I get the rebase completed.

>     2. how about leaving `mp_todecimal` as a macro and calling `s_mp_todecimal_fast()` from `mp_to_radix()` if the `radix == 10`?

Same here.

**Comment by czurnieden** [comment]
> Snprintf should not be used.

You can use `mp_to_radix`  which also allows for bigger chunks at the tree-leaves to speed it up.

> I'd really love to see an explanation on how the algorithm works

At first I thought it was [Bouvier et al.](https://hal.inria.fr/hal-00864293v1/document) but it is really "just" D&C  (Schoenhage, Knuth, Bernstein, and probably many more but I think Arnold Schoenhage was first) and uses Barret reduction as the divsion algorithm.

Oh, and my question is still relevant: why is it called `todecimal`? The algorithm itself allows all bases `>= 2`, why the restriction to decimal? And why no `fromdecimal`, the algorithm works both ways?

**Comment by MasterDuke17** [comment]
> > Snprintf should not be used.
> 
> You can use `mp_to_radix` which also allows for bigger chunks at the tree-leaves to speed it up.

At the point in the code where I was using `snprintf` I just have `int`s, not `mp_int`s, so I would assume the specialized code I just pushed would be faster, but could be persuaded otherwise.

> Oh, and my question is still relevant: why is it called `todecimal`? The algorithm itself allows all bases `>= 2`, why the restriction to decimal? And why no `fromdecimal`, the algorithm works both ways?

Heh, it's just because I didn't realize it would work for other bases (and I personally was most interested in decimal). I'll try adopting all bases (and renaming to something like `s_mp_to_radix_fast`). Figuring out how to do it in reverse might take me longer, I'm not sure I can promise that one.

**Comment by MasterDuke17** [comment]
I see I'll need to implement a max length parameter to pass on to the new `mp_to_radix`. Or is the suggestion in https://github.com/libtom/libtommath/pull/330#issuecomment-527201525 what I should do instead of turning a macro into a function?

**Comment by MasterDuke17** [comment]
> At first I thought it was [Bouvier et al.](https://hal.inria.fr/hal-00864293v1/document) but it is really "just" D&C (Schoenhage, Knuth, Bernstein, and probably many more but I think Arnold Schoenhage was first) and uses Barret reduction as the divsion algorithm.

@czurnieden I wasn't kidding when I said I'm not a mathematician. If the algorithm you linked is faster and someone implements that instead I'll be happy. I just want faster stringification in libtommath, if it comes from code I didn't write that's one less thing for me to worry about/follow in the future.

**Comment by czurnieden** [comment]
> the specialized code I just pushed would be faster, but could be persuaded otherwise.

Not a matter of persuasion more a matter of benchmarking. Using bigger chunks (200-300 bits or so, maybe more) with `mp_to_radix` might or might not be faster than using small chunks with some hand-optimized conversion. You need to test it out.

>  Figuring out how to do it in reverse might take me longer

Take your time, nobody is rushing you.

>  If the algorithm you linked is faster

I don't know, probably.

But I didn't link it for that reason, the link was meant as a possible explanation; it is what I thought you did the first time you mentioned it (the "without division" part to be exact) before I looked at your code and the original from [Lars Hellström](https://sourceforge.net/p/tcl/mailman/message/31972670/).
Both are connected nevertheless, yours is more or less  a simplified and also specialized version (base 2 to base 10 only).

> if it comes from code I didn't write that's one less thing for me to worry about/follow in the future.

You started it, you have the right to finish it.

**Comment by MasterDuke17** [comment]
> > how about leaving mp_todecimal as a macro and calling s_mp_todecimal_fast() from mp_to_radix() if the radix == 10?
> 
> Not possible since MP_HAS is only available privately and we should keep it like this.

@minad so you don't like the proposed diff in https://github.com/libtom/libtommath/pull/330#issuecomment-527201525? Do you want me to do to the new `mp_to_decimal` macro what I did with the old `mp_todecimal` macro (i.e., turn it into a simple wrapper function)?

**Comment by minad** [comment]
Would it be possible to get the generalized version for arbitrary bases replacing the current impl?

**Comment by MasterDuke17** [comment]
> Would it be possible to get the generalized version for arbitrary bases replacing the current impl?

Once I get the current version renamed and accepting (and using) a maxlen parameter I'll attempt to convert to an arbitrary base implementation. However,


> Both are connected nevertheless, yours is more or less a simplified and also specialized version (base 2 to base 10 only).

I'll admit I don't understand the algorithm enough the know how easy it would be to extend to the rest of the bases. Should I just extend it to 2..10?

**Comment by sjaeckel** [comment]
> Would it be possible to get the generalized version for arbitrary bases replacing the current impl?

let's first get this one finished by @MasterDuke17 and then we'll have a look on how to get the generalized version, okay?

**Comment by MasterDuke17** [comment]
How do people feel about me creating a new PR (or aggressively cleaning up history for this one) once I have the maxlen support implemented? I think at that point it could be merged as is (with some tests and documentation) and then another new one created with a more generalized implementation.

**Comment by sjaeckel** [comment]
> How do people feel about me creating a new PR (or aggressively cleaning up history for this one) once I have the maxlen support implemented? I think at that point it could be merged as is (with some tests and documentation)

please simply squash everything together, rebase on develop and force-push

> and then another new one created with a more generalized implementation.

:+1:

**Comment by sjaeckel** [comment]
and if you think it's ready, please request the review so we can clean some more things up/improve some stuff before merging

**Comment by minad** [comment]
Do we want the specialised version if we get the generalized one? Will the specialised one be faster (smaller constant factors)?

**Comment by sjaeckel** [comment]
> Do we want the specialised version if we get the generalized one? Will the specialised one be faster (smaller constant factors)?

good questions! we can check that as soon as we have the specialised version ;-)

File: bn_s_mp_to_decimal_fast.c [hunk 1]
```diff
@@ -0,0 +1,234 @@
...
+      goto LBL_ERR;
+   }
+   if (mp_isneg(&number)) {
+      if ((err = mp_neg(&number, &number)) != MP_OKAY) {
+         goto LBL_ERR;
+      }
+      result[0] = '-';
+      result_addr += 1;
+   }
+   if ((err = mp_init_set(&n, (mp_digit)1000)) != MP_OKAY) {
```
czurnieden: As long as we still support MP_8BIT `mp_digit` can be as small as 7 (seven) bits. Please use e.g.: `mp_set_u32` instead for any (unsigned) value larger than 127.

MasterDuke17: Done in 9d1bae2.

czurnieden:
@MasterDuke17 
I'm too lazy to fiddle all of the hardware optimizations out of my C-version yet, but I have it in my [Javascript-bignum library, too](https://github.com/czurnieden/Little/blob/master/libs/bignum.js), search for  `Bigint.prototype.toString` (the naive part is in `Bigint.prototype.toStringSmall` above it). You might need to switch off fast-division if you want to run that JS-lib, it doesn't seem to work in newer browsers anymore, had no time to look for the reason(s) yet (most likely PEBCAK ;-) ).

It might get you an insight at how your algorithm works and how it can be made to work for other bases too and also how to read from a string (`String.prototype.toBigint`, the naive version in `String.prototype.toBigintSmall` above it.).

Hope it'll help at least a little bit.

Hu?
Oh, yes, it's GPL, I see, sorry.
\*ahem\*
I hereby and herewith give you permission to use that code to implement fast string-handling in LibTomMath.
Better? ;-)

MasterDuke17:
Btw, @czurnieden, @minad, @sjaeckel are you guys on IRC anywhere? I've had some questions that are kind of not directly on-topic to this issue.

sjaeckel:
> are you guys on IRC anywhere?

I'm on Freenode in #libtom

bn_s_mp_to_decimal_fast.c
```diff
@@ -0,0 +1,234 @@
...
+   if ((err = mp_init_copy(&number, a)) != MP_OKAY) {
+      goto LBL_ERR;
+   }
+   if (mp_isneg(&number)) {
+      if ((err = mp_neg(&number, &number)) != MP_OKAY) {
+         goto LBL_ERR;
+      }
+      result[0] = '-';
+      *result_addr += 1;
+      maxlen -= 1;
```
sjaeckel: we need to check `maxlen` before writing to the buffer/decrementing it

minimum length must be 2 IIUC

bn_s_mp_to_decimal_fast.c
```diff
@@ -0,0 +1,234 @@
...
+         t++;
+         if (n2 < 10) {
+            t++;
+         }
+         if (n2 == 0) {
+            t++;
+         }
+      }
+
+      if (*maxlen < (size_t)t || (*maxlen -= (size_t)t) < 1) {
```
sjaeckel: can you please split this up in
1. the check of `maxlen < t+1` here
2. the decrement operation besides the increment of the `result` pointer

tommath.h
```diff
@@ -703,6 +703,7 @@ MP_DEPRECATED(mp_to_radix) mp_err mp_toradix(const mp_int *a, char *str, int rad
 MP_DEPRECATED(mp_to_radix) mp_err mp_toradix_n(const mp_int *a, char *str, int radix, int maxlen) MP_WUR;
 mp_err mp_to_radix(const mp_int *a, char *str, size_t maxlen, int radix) MP_WUR;
 mp_err mp_radix_size(const mp_int *a, int radix, int *size) MP_WUR;
+mp_err mp_to_decimal(const mp_int *a, char *str, size_t maxlen) MP_WUR;
```
sjaeckel: I've already asked this before: is there a reason why this has a separate public API function instead of calling it in `mp_to_radix()` when `radix == 10`?

demo/test.c
```diff
@@ -2213,6 +2213,54 @@ static int test_s_mp_toom_sqr(void)
...
+   mp_int a, b;
+   int size, err, strlength;
+   char *str;
+
+   if ((err = mp_init_multi(&a, &b, NULL)) != MP_OKAY) {
+      goto LTM_ERR;
+   }
+   for (size = 1; size < 1000; size += 10) {
+      int times;
+      printf("Testing mp_to_decimal: %5d bits    \r", size);
```
sjaeckel: bits would be `size * MP_DIGIT_BIT` so we should use the word 'digits' here  instead

bn_s_mp_to_decimal_fast.c
```diff
@@ -0,0 +1,234 @@
...
+         if ((err = mp_div_2d(&M4, mp_get_i32(&shift) + 6, &M4, NULL)) != MP_OKAY) {
+            goto LBL_ERR;
+         }
+         if ((err = mp_mul_2(&M2, &M2)) != MP_OKAY) {
+            goto LBL_ERR;
+         }
+         if ((err = mp_sub(&M4, &M2, &M4)) != MP_OKAY) {
+            goto LBL_ERR;
+         }
+         if ((err = mp_add_d(&M4, 1, &M4)) != MP_OKAY) {
```
sjaeckel: probably `mp_incr()`?

bn_s_mp_to_decimal_fast.c
```diff
@@ -0,0 +1,234 @@
...
+            goto LBL_ERR;
+         }
+         if ((err = mp_div_2d(&M4, 3, &M4, NULL)) != MP_OKAY) {
+            goto LBL_ERR;
+         }
+         if ((err = mp_sub_d(&M4, 1, &M4)) != MP_OKAY) {
```
sjaeckel: `mp_decr()`?

bn_s_mp_to_decimal_fast.c
```diff
@@ -0,0 +1,234 @@
...
+   if ((err = mp_mul(&nL[precalc_array_index], &q, &nLq)) != MP_OKAY) {
+      goto LBL_ERR;
+   }
+
+   if ((err = mp_sub(number, &nLq, &r)) != MP_OKAY) {
+      goto LBL_ERR;
+   }
+
+   if (mp_isneg(&r)) {
+      if ((err = mp_sub_d(&q, 1, &q)) != MP_OKAY) {
```
sjaeckel: `mp_decr`?

bn_s_mp_to_decimal_fast.c
```diff
@@ -0,0 +1,234 @@
...
+   if ((err = s_mp_to_decimal_fast_rec(&number, nL, shiftL, mL, precalc_array_index - 1, 1, result_addr,
+                                       &maxlen)) != MP_OKAY) {
+      goto LBL_ERR;
+   }
+   *result_addr[0] = '\0';
+
+   err = MP_OKAY;
+
+LBL_ERR:
+   mp_clear_multi(&number, &n, &shift, &M, &M2, &M22, &M4, &M44, NULL);
```
sjaeckel: the error handling has to be a bit more careful
here and in the statement below you're probably trying to clear some mp_int's which weren't initialized before

bn_s_mp_to_decimal_fast.c
```diff
@@ -0,0 +1,234 @@
...
+                                int left,
+                                char **result,
+                                size_t *maxlen)
+{
+   mp_int q, nLq, r;
+   mp_err err;
+
+   if (precalc_array_index < 0) {
+      int n = mp_get_i32(number), n2 = n, t = 0, c;
+      char *i = *result;
```
czurnieden: I don't care much about these things normally, but `i` is a really bad variable name.

bn_s_mp_to_decimal_fast.c
```diff
@@ -0,0 +1,234 @@
...
+            goto LBL_ERR;
+         }
+         if ((err = mp_neg(&M4, &M4)) != MP_OKAY) {
+            goto LBL_ERR;
+         }
+      }
+      if ((err = mp_init_copy(&mL[precalc_array_index], &M4)) != MP_OKAY) {
+         goto LBL_ERR;
+      }
+      precalc_array_index++;
```
czurnieden: I don't think that `precalc_array_index` can get larger than 20  but I would prefer a little check here, just in case.

bn_s_mp_to_decimal_fast.c
czurnieden: With the expected large cutoff values it might even be economically more reasonable to just init all `mp_int`s instead of keeping pace with all of the individual inits..

MasterDuke17:
I'm working on addressing the comments now. However, should I also switch to the new error-handling format (i.e., following https://github.com/libtom/libtommath/wiki/The-Sourcecode:-Formatting-and-Style#exceptions)?

tommath.h
MasterDuke17: I personally think it would be a little nicer to just call it in `mp_to_radix` (though I'll admit that's partly just because then it wouldn't require any changes in MoarVM to take advantage of it (the very reason I made this PR in the first place)). However, I think it was @minad who had an objection/reason.

czurnieden:
@MasterDuke17 
> However, should I also switch to the new error-handling format

Not yet, it is still in limbo in #339, that wiki page is just a stub for now.

MasterDuke17:
All comments so far addressed in 818de1a.

bn_s_mp_to_decimal_fast.c
sjaeckel: why not simply do a `for-loop` instead of the `while(1)` and check the `precalc_array_index` directly after the loop?

as it is now it could still be possible that you're overwriting `nl` and `shiftL`

bn_s_mp_to_decimal_fast.c
```diff
@@ -0,0 +1,242 @@
...
+   if ((err = s_mp_to_decimal_fast_rec(&number, nL, shiftL, mL, precalc_array_index - 1, 1, result_addr,
+                                       &maxlen)) != MP_OKAY) {
+      goto LBL_ERR;
+   }
+   *result_addr[0] = '\0';
+
+   err = MP_OKAY;
+
+LBL_ERR:
+   mp_clear_multi(&number, &n, &shift, &M, &M2, &M22, &M4, &M44, &nL[0], &shiftL[0], &mL[0], NULL);
```
sjaeckel: It's better now, but the potential clear of uninitialized variables still persists!

I think the easiest way to solve it now would be by initializing `nL, shiftL, mL` with `0`.

tommath.h
sjaeckel: Yep, @minad said that in https://github.com/libtom/libtommath/pull/330#issuecomment-524537046

@minad why do you think we should have separate functions for the different radices instead of all going through `mp_to_radix()` and potentially going to a fast version from there?

bn_s_mp_to_decimal_fast.c
MasterDuke17: I'm not sure I follow. Aren't they initialized at https://github.com/libtom/libtommath/pull/330/files#diff-e2627302799cc9765e140c386c4906acR113?

sjaeckel: grml, this initial comment is on the wrong line m( 
...it should've been two lines below, sorry!

> I'm not sure I follow. Aren't they initialized at https://github.com/libtom/libtommath/pull/330/files#diff-e2627302799cc9765e140c386c4906acR113?

The first element of the array is initialized there, yes. But all the other 19 elements are initialized on the fly via `mp_init_copy()` calls in lines 194, 197 & 220.

demo/test.c
```diff
@@ -2264,8 +2312,10 @@ int unit_tests(int argc, char **argv)
       T1(s_mp_karatsuba_sqr, S_MP_KARATSUBA_SQR),
       T1(s_mp_toom_mul, S_MP_TOOM_MUL),
       T1(s_mp_toom_sqr, S_MP_TOOM_SQR),
+      T1(mp_to_decimal, S_MP_TO_DECIMAL_FAST)
 #undef T2
 #undef T1
+#undef T
```
sjaeckel: thanks, I missed that, but it should be `T0`

bn_s_mp_to_decimal_fast.c
```diff
@@ -0,0 +1,245 @@
+#include "tommath_private.h"
+#include <string.h>
+#ifdef BN_S_MP_TO_DECIMAL_FAST_C
+/* LibTomMath, multiple-precision integer library -- Tom St Denis */
+/* SPDX-License-Identifier: Unlicense */
+
+/* store a bignum as a decimal ASCII string */
+mp_err s_mp_to_decimal_fast_rec(const mp_int *number, mp_int *nL, mp_int *shiftL, mp_int *mL, int precalc_array_index,
```
sjaeckel: oh I just saw that this should be static

bn_s_mp_to_decimal_fast.c
```diff
@@ -0,0 +1,245 @@
...
+      char *result_str = *result;
+      char sprintf_str[4] = "000";
+
+      while (n) {
+         sprintf_str[2 - digits_to_copy] = mp_s_rmap[n % 10];
+         digits_to_copy++;
+         n /= 10;
+      }
+
+      if (!left && n_orig < 100) {
```
sjaeckel: @fperrad can you please expand [1] with the rules on how such statements should be embraced as of your usual linting?

[1] https://github.com/libtom/libtommath/wiki/The-Sourcecode:-Formatting-and-Style

bn_s_mp_to_decimal_fast.c
sjaeckel: as the array members are initialized via `mp_init_copy()` I thought it'd be more expensive to init them first and then grow&set them.
Therefore we zero the arrays beforehand so `mp_clear_multi()` can decide internally whether there's something to clear or not

sjaeckel:
> > However, should I also switch to the new error-handling format
> 
> Not yet, it is still in limbo in #339, that wiki page is just a stub for now.

right, you can also do that now

MasterDuke17:
@minad ping re using `MP_HAS` in `mp_to_radix` where `radix == 10` vs the approach as is.

minad:
@MasterDuke17 My proposal would be to work on a generalized solution supporting all possible radices. Then we can just replace the existing mp_to_radix function. However there is still some discussion going on, how the function should look like, with regards to allocation etc. See #348.

sjaeckel:
> My proposal would be to work on a generalized solution supporting all possible radices. Then we can just replace the existing mp_to_radix function

IIUC that's not an option as this implementation is only faster for big numbers (@MasterDuke17 mentioned something like 500 digits on an older machine and 1500 digits on a newer one, I don't know whether it's decimal- or ltm-digits)

MasterDuke17:
> > My proposal would be to work on a generalized solution supporting all possible radices. Then we can just replace the existing mp_to_radix function
> 
> IIUC that's not an option as this implementation is only faster for big numbers (@MasterDuke17 mentioned something like 500 digits on an older machine and 1500 digits on a newer one, I don't know whether it's decimal- or ltm-digits)

I wasn't very clear, those are numbers I was raising two to the power of (e.g., 2^512 as you can see in some of my benchmarks in earlier commits). On my faster machine it was somewhere around `mp_int.used > 10` (https://github.com/libtom/libtommath/pull/330/commits/3ee4c5ebe32e5dc6da734b7d10924bb27dec2521#diff-32d269f5bde171a5a4d1b7709b3b0d4eR14).

@czurnieden has some ways to speed up this implementation (and re-writing to not be recursive may help, also trading memory for speed and keeping around the arrays of calculated values), but the constant factor may always put it slower than the current `mp_to_radix` implementation for small values. I don't have enough intuition to know if the constant factor can be made low enough that the loss in speed is compensated for by the decrease in complexity to switch completely to this.

minad:
@sjaeckel You are right. We should do something similar like karatsuba. However I don't think it makes sense to merge a specialized version for base 10 if we can get a generalized version which works as well.

sjaeckel:
> @czurnieden has some ways to speed up this implementation

Cool then let's do that!


> re-writing to not be recursive may help

That's also a good idea


> also trading memory for speed and keeping around the arrays of calculated values

You mean something like a "registry by radix" with the pre-calculated values of `nl[], shiftL[], mL[]`? sounds very nice, especially since it's most likely only radices 10 and 16, probably sometimes 2 and for mtest 64!


> We should do something similar like karatsuba.

:+1: 

> However I don't think it makes sense to merge a specialized version for base 10 if we can get a generalized version which works as well.

Fine by me!

I thought we could merge this version soon, so we have at least a faster `mp_to_radix(..., 10)` for big inputs in the next release... but if you think we should wait it's also okay.

bn_s_mp_to_decimal_fast.c
MasterDuke17: Both `s_mp_to_decimal_fast` and `s_mp_to_decimal_fast_rec` I assume?

MasterDuke17: Oh, looks like not.

MasterDuke17:
> > @czurnieden has some ways to speed up this implementation
> 
> Cool then let's do that!

I believe @czurnieden's suggestions are straightforward, but will require me to understand more of the actual math than I do now, so it may take a while.

> > re-writing to not be recursive may help
> 
> That's also a good idea

I'll work on this, but same re may take a while.

> > also trading memory for speed and keeping around the arrays of calculated values
> 
> You mean something like a "registry by radix" with the pre-calculated values of `nl[], shiftL[], mL[]`? sounds very nice, especially since it's most likely only radices 10 and 16, probably sometimes 2 and for mtest 64!

Is there anything similar in the codebase I could take a look at for inspiration?

> > However I don't think it makes sense to merge a specialized version for base 10 if we can get a generalized version which works as well.
> 
> Fine by me!
> 
> I thought we could merge this version soon, so we have at least a faster `mp_to_radix(..., 10)` for big inputs in the next release... but if you think we should wait it's also okay.

If a fully generalized version would cause a delay until the release after next I'd advocate for including the base-10 version. But that's only so it be more likely to get into MoarVM faster (since my completely unmeasured opinion is that 99% of the uses of mp_to_radix are for base-10).

bn_s_mp_to_decimal_fast.c
MasterDuke17: Done in d2678b0.

minad:
@MasterDuke17 I don't think we will add this to the next release. It needs more time to mature. In particular since there are already many changes and lots of deprecations which need to be released. But I am interested in getting this in soon, I would use this faster to_radix myself. My current impression of this library is that it is not moving particularly fast, but rather taking it's time to figure out how things should be.

With regards to moarvm, it is not yet up to date with the latest release. And it would also probably help if moar gets ported to the tommath develop branch. This would help both projects I guess (advancing moar and feedback for tommath).

czurnieden:
@MasterDuke17

Couldn't sleep, so did the main cleanup in `bn_mp_set_str.c` (string2bigint) and made it stand-alone (with option, see comments there). It is twice as large as `bn_mp_read_radix.o` but O(log(n) * M(n)) (In the current LTM M(n) = n^1.46 with Toom-Cook 3-way) instead of O(n^2).

But a fast string2bigint isn't very urgent, a fast bigint2string on the other side is. Any progress made with it? Can I help?

MasterDuke17:
Nice. Looking at your code has helped me to understand what’s going on more (not yet completely though). I’ve made a *tiny* bit of progress, but I would have absolutely no objection if you carried on ahead (and if you do, feel free to take all or nothing from what I have committed so far). If you don’t, I’ll keep slowly plugging away.

czurnieden:
@MasterDuke17

Did most of the chores in `bn_mp_get_str.c`. It should be ready now to fill the cache with the approximate inverses and use multiplication and Barrett's correction instead of `mp_div`.
A hint to get the `k` in `(2^k)/n`: in `a/(base^n)` `a` is more than 2 times larger in magnitude than `base^n`. If you take that large multiplier (that is `k = floor(log_2(a))`) the error is unity at most, only one correction is necessary.

For further reading try [Brent et al.: "Modern Computer Arithmetic"](https://arxiv.org/abs/1004.4710) section 2.4.1. and Kalles, pardon, Karl Hasselström's master thesis ["Fast Division of Large Integers"](https://treskal.com/s/masters-thesis.pdf) (shows correction if the error is bigger than unity allowing for smaller `k`), and the [wiki-page](https://en.wikipedia.org/wiki/Barrett_reduction),  of course, for the error computation.

Another hint: Barrett division is not as complicated as it seems, so if you think you need to do something complicated you are probably on the wrong path (optimizing `k` is a bit hard, but you don't need it now, only later and only if it is **way** too slow without).

Example for `a/n` (all divisions here are truncating) resulting in `q` the quotient and `r` the remainder (assuming I didn't do anything wrong here ;-) )

Input `a, n` with `log(a) >= 2*log(n)` (roughly, the exact ranges and proportions are described in the papers linked above)

```text
k = floor(log_2(a)) # which is too large, but good enough for now
m = (2^k)/n         # the approximate reciprocal
q = (a * m) >> k    # compute approximate quotient
r = a - q * n       # compute approximate remainder
if (r >= n)         # check if off-by-one
   q = q + 1        # correct quotient (one too small with the value of k chosen) 
   r = r - n        # correct remainder (one n too big with the value of k chosen)
endif
return q,r
```

MasterDuke17:
@czurnieden I'm making progress. I now have a Frankenstein combination of my bn_s_mp_to_decimal_fast.c and your bn_mp_get_str.c that sort of works. However, I currently have two problems. The first is that my new combined code is adding some extra leading `0`s. The second is that valgrind does not like your code (nor the combined version). It's complaining about `Invalid write of size 1` here https://github.com/czurnieden/libtommath/blob/faster_number_conversion/bn_mp_get_str.c#L55 and `Invalid read of size 1` here https://github.com/czurnieden/libtommath/blob/faster_number_conversion/bn_mp_get_str.c#L116 and `Conditional jump or move depends on uninitialised value(s)` here https://github.com/czurnieden/libtommath/blob/faster_number_conversion/bn_mp_get_str.c#L115. If you're interested, this is what I have so far: https://gist.github.com/MasterDuke17/cae7ba73306ad315372a570b041c9c52

czurnieden:
> The first is that my new combined code is adding some extra leading 0s

Yeah, that part is a bit "how you doin'" and may need some extra massage.

>  The second is that valgrind does not like your code

It is more or less a "working sketch" to give you one way to tackle your problem as an example. It makes me wonder that valgrind didn't crash&burn completely ;-)

I uploaded a slightly more "ready to go" version (none of the functions have their `if( (err = mp_foo()) != MP_OKAY) goto ERR;` to make it more readable). It even survived valgrind ;-)

MasterDuke17:
Hm, at this point I'm not sure what's missing from your new version (other than the error handling and such)? Is there actually something that should be ported in from my code?

czurnieden:
> Is there actually something that should be ported in from my code?

You still have the base-10 optimized part the computing of the reciprocal) which should find it's way into it because the whole thing would work much better with fast division in general but as the most used bases are probably 16, 10, and 64, and maybe base 2, too. The power-of-two bases are quickly computed, only base 10 is more work.

There is even a  discussion here to reduce the functionality in a later LTM version to just those power-of-two bases and base 10.

So I would say that you should concentrate on what you started with and implement a (preprocessor/ `MP_HAS()`) branch for base 10 (the power-of-two bases will get their optimization in  `mp_to/from_radix`) such that people who love their odd bases can use the general version and the rest can get the handoptimized ones.
As noted above: the latter might be the default in version 2.0.

Soooo: no cheap excuses here, go on! :-)

czurnieden:
Please rebase from time to time.
It gets too hectic otherwise as I can tell from experience ;-)

czurnieden:
@MasterDuke17 took the liberty to take the algorithm apart and [wrote down](https://github.com/czurnieden/libtommath/blob/faster_number_conversion/doc/barrett_todigits.tex) what I found.
(I don't think I used more Latex packages than exist in a recent TeTex bundle but if it doesn't build drop me a note)
Translating from code to Latex might have introduced some errors, please check.

It seems as if all of the odd radices get dropped in version 2.0.0 (who actully needs more than the base 2, 10, 16 and for saving space 64?) so I think you can drop the rest and concentrate on refining and optimizing what you started with: base 10. I think Lars Hellström made some suggestions in his [post archived at Sourceforge](https://sourceforge.net/p/tcl/mailman/message/31972670/) like dropping recursion and adding some additional caching.

Thank you for your patience!

czurnieden:
@sjaeckel 
>  if I could choose I'd really love to see an explanation on how the algorithm works

It took me a bit, but I finally got it.
I think ;-)
Put it a LaTeX file [here](https://github.com/czurnieden/libtommath/blob/faster_number_conversion/doc/barrett_todigits.tex).

czurnieden:
@MasterDuke17 

I extended my [description](https://github.com/czurnieden/libtommath/blob/faster_number_conversion/doc/barrett_todigits.tex) and hope it is complete now (sans the iterative version(s)). If not feel free to ask me, I know it is still a bit of a mess.

MasterDuke17:
> @MasterDuke17
> 
> I extended my [description](https://github.com/czurnieden/libtommath/blob/faster_number_conversion/doc/barrett_todigits.tex) and hope it is complete now (sans the iterative version(s)). If not feel free to ask me, I know it is still a bit of a mess.

Cool. If I understand it correctly we would need `n==24` to be able to handle the largest possible number? I've been a bit distracted, but I hope to have some time in the next couple of days to do some more work.

czurnieden:
> If I understand it correctly we would need n==24 to be able to handle the largest possible number?

The names used in that text should be `n` for the base (`n = 10^3` for base 10), `s` for the shift (`s = 20` for base 10) and `k` for the exponent (the number of squarings or the index into the lists if you like). If that is mixed up somewhere it is a bug in my text and I would appreciate it if you would drop me a note if that is the case.
So I interpret the number in your question as `k == 24`. Is that correct?

The exact type of the sizes is still in limbo (`size_t` or not, that seems to be the question here.) but I think I calculated `k = 22` for the current type? Yes, I should clean it up a bit, it's quite a mess, admitted.

I wouldn't go with static arrays in the first place but use the heap instead. You can calculate the number of reciprocals in advance and check the result (and repair if necessary), so I would just `malloc` the right amount and keep the lib a bit smaller that way.

Oh, and although I added the necessary information to extend it to all of the other bases you can stay with base 10 for now.

Did I add the "bigger chunks" stuff? No, I didn't, I was lying when I said it is complete ;-)
OK, will add but not today.

czurnieden:
@MasterDuke17 
Took the liberty to update my [description](https://github.com/czurnieden/libtommath/blob/faster_number_conversion/doc/barrett_todigits.tex). Added the necessary numerical corrections for the other bases and generalized the algorithm to use all these bases. You do not have to use the optimization for the large leafs for but keep in  mind that even the small leafs need a different size of your small buffer `sprintf_str` now (between 4 and 9 bytes). Will be a bit fuddly, admitted, but should not be a giant problem.

MasterDuke17:
@czurnieden Great, thanks! I've been sidetracked by getting MoarVM upgraded to libtommath 1.2.0. We just recently turned on -Wall and -Wextra when building MoarVM and that threw up a ton of warnings in our code that uses libtommath (e.g., we were almost never handling return values and that now warns). But I should have that done soon and next thing on the list is working on this.

minad:
@MasterDuke17 Good to hear that you are doing the update! The last time I looked, the MoarVM codebase had some math functions which are now available in upstream ltm, e.g., floating point conversion and bitwise operations. If something else is missing here let us know!

sjaeckel:
Looks like @MasterDuke17 won't finish this, as MoarVM has decided to go for gmp instead of ltm.

PR #331, opened 2019-08-31
Barrett todecimal

MasterDuke17:
https://github.com/libtom/libtommath/pull/330, but rebased onto feature-detection2 so `mp_todecimal` can be converted into a function that uses `MP_HAS` to selectively use the new `s_mp_todecimal_fast`.

### DIFF
--- a/bn_mp_todecimal.c
+++ b/bn_mp_todecimal.c
@@ -1,0 +1,23 @@
+#include "tommath_private.h"
+#ifdef BN_MP_TODECIMAL_C
+/* LibTomMath, multiple-precision integer library -- Tom St Denis */
+/* SPDX-License-Identifier: Unlicense */
+
+/* stores a bignum as a decimal ASCII string, using Barrett
+ * reduction if available.
+ */
+
+mp_err mp_todecimal(const mp_int *a, char *str)
+{
+   mp_err err;
+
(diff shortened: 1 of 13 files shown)


MasterDuke17:
@minad @czurnieden @sjaeckel how does this look? I'm in the process of adding tests, but I'm not sure about documentation. How much is needed and where does it go?

minad:
Looks good! The CI should be green and we need tests. You can test against mp_toradix for random numbers.

MasterDuke17:
@minad The two problems in travis are the same, but I'm not exactly sure how to fix them.
```
bn_s_mp_todecimal_fast.c:16:28: warning: implicitly declaring library function 'snprintf' with type 'int (char *, unsigned long, const char *, ...)' [-Wimplicit-function-declaration]

      int new_pos = left ? snprintf(*result, 4, "%u", mp_get_i32(number)) : snprintf(*result, 4, "%03u",

                           ^

bn_s_mp_todecimal_fast.c:16:28: note: include the header <stdio.h> or explicitly provide a declaration for 'snprintf'
```
However, stdio.h is already included. Some googling suggests other people have seem similar problems and it's some weird thing about different versions of header files. Do you have any suggestions?

minad:
Snprintf should not be used.

sjaeckel:
closed again in favor of #330

PR #332, opened 2019-09-02
Add `mp_to_radix()`

sjaeckel:
... as a replacement function of `mp_toradix` resp. `mp_toradix_n`


minad:
Are we going to replace all the mp_to* functions by mp_to_* functions just for the sake of more consistent names?

bn_mp_to_radix.c
```diff
@@ -7,7 +7,7 @@
  *
  * Stores upto maxlen-1 chars and always a NULL byte
  */
-mp_err mp_toradix_n(const mp_int *a, char *str, int radix, int maxlen)
+mp_err mp_to_radix(const mp_int *a, char *str, size_t maxlen, int radix)
```
minad: size_t :+1:

sjaeckel:
> Are we going to replace all the mp_to* functions by mp_to_* functions just for the sake of more consistent names?

they should be gone already, aren't they?

minad:
>  they should be gone already, aren't they?

What do you mean? You are deprecating them here, replacing them by `_to_` functions with size_t argument. But this is the only change?

sjaeckel:
I'm replacing them for 2 reasons

1. to add the `maxlength` parameter
2. consistency

or do you have other plans with those `mp_toXX` functions/macros?

minad:
Ah, I missed the maxlen param you added. I agree 100% with this change. While the user could always ensure to have a buffer which is large enough by computing radix_size, this is easy to misuse in practice and leading to unsafe code.

Deprecate all the unsafe stuff :+1: :+1: :+1:

Maybe move mp_toradix also to bn_deprecated.c?

minad:
In the case of overflow, mp_to_radix should return some kind of error I guess?

sjaeckel:
> Maybe move mp_toradix also to bn_deprecated.c?

I also thought about it, but as there are only trampolines for now I preferred to keep it like that

> In the case of overflow, mp_to_radix should return some kind of error I guess?

right! I wanted to add that, but forgot it... `MP_VAL` will be it... which reminds me of... should we start to add (some) more meaningful error codes at one point?

minad:
Meaningful errors would be nice, but as I am using ltm - I just care about yay or nay...

You can trampoline to mp_to_radix with maxlen=SIZETMAX?

sjaeckel:
> You can trampoline to mp_to_radix with maxlen=SIZETMAX?

ew, that's ugly and shows how awful no length checking is... I like it

minad:
The compiler might even constant propagate and optimize away the check then 🙈

minad:
Found a few issues. There are a few mp_toradix left in the codebase and the docs if you grep. The ifdef check in mp_to_radix.c is wrong.

sjaeckel:
> Found a few issues. There are a few mp_toradix left in the codebase and the docs if you grep. The ifdef check in mp_to_radix.c is wrong.

looking into it

Issue #333, opened 2019-09-02
Unexpected behavior of `mp_prime_next_prime()`


sjaeckel:
I'll just leave this here... :-D

the diff is based on #332 

```
$ git diff -U1
diff --git a/demo/test.c b/demo/test.c
index 38a75a9..5832472 100644
--- a/demo/test.c
+++ b/demo/test.c
@@ -1135,3 +1135,3 @@ static int test_mp_read_radix(void)
 
-   while (0) {
+   while (01) {
       char *s = fgets(buf, sizeof(buf), stdin);
@@ -1139,2 +1139,4 @@ static int test_mp_read_radix(void)
       mp_read_radix(&a, buf, 10);
+      mp_to_radix(&a, buf, sizeof(buf), 10);
+      printf("%s, %lu\n", buf, a.dp[0] & 3);
       mp_prime_next_prime(&a, 5, 1);
$ make test_standalone -j9 >/dev/null 
$ ./test read_radix
Digit size 64 Bit 
Size of mp_digit: 8
Size of mp_word: 16
MP_DIGIT_BIT: 60
MP_PREC: 32
TEST mp_read_radix

a == 56
a == 456
a == 123456
8
8, 0
7, 3

```

czurnieden:
And I thought I could relax for a day or two and enjoy the last days of summer?
\*sigh\*
Nah, you're right, that would be boring! ;-)

> `+   while (01) {`

Did it come with greetings from Rear Admiral G. Hopper? ;-)

Could it be a problem with the specific diet you fed `char *s = fgets(buf, sizeof(buf), stdin);` with?

But `mp_prime_next_prime` should return an error in this case, will look into it.

sjaeckel:
> And I thought I could relax for a day or two and enjoy the last days of summer?

please do so!

> Did it come with greetings from Rear Admiral G. Hopper? ;-)

of course it did! sir! ;-)

> Could it be a problem with the specific diet you fed `char *s = fgets(buf, sizeof(buf), stdin);` with?
> 
> But `mp_prime_next_prime` should return an error in this case, will look into it.

uhm, why should it return an error?

I gave an 8 to `mp_prime_next_prime` and I wouldn't expect to get 7 as return value as that's obviously the previous prime or did I understand 'next prime' wrong?

sjaeckel:
Is it possible that this is the fix?

```
$ git diff -U1 --ignore-space-change 
[...]

PR #334, opened 2019-09-02
repair of #333

czurnieden:
As the title suggests: in re #333

### DIFF
--- a/bn_mp_prime_next_prime.c
+++ b/bn_mp_prime_next_prime.c
@@ -39,6 +39,9 @@ mp_err mp_prime_next_prime(mp_int *a, int t, int bbs_style)
                         return MP_OKAY;
                      }
                   }
+               } else {
+                  mp_set(a, s_mp_prime_tab[x + 1]);
+                  return MP_OKAY;
                }
             } else {
                mp_set(a, s_mp_prime_tab[x + 1]);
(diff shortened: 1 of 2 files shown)


sjaeckel:
wouldn't "always doing the scan upwards" have the same result but with less code?

czurnieden:
> wouldn't "always doing the scan upwards" have the same result but with less code?

Probably?

But I try to change as little as possible in bugfixes.
Refactoring is a terrain better fitted for e,.g.: @minad ;-)

minad:
Add regression test?

czurnieden:
> Add regression test?

Oh, that's very nice of you!

minad:
😁

czurnieden:
Found another problem with input zero while writing the test.
Or the test is buggy ;-)

czurnieden:
Argh, forgot to mention that the bug seems to be fixed now (#333 and the problem with input smaller than 2).
Squash and merge?

@sjaeckel : 
> wouldn't "always doing the scan upwards" have the same result but with less code?

Makes more sense now, faced with the second problem (input smaller than 2) so: refactor or use this bugfix (for now)?

sjaeckel:
> Pushed it onto this branch for you to look at.

cool!

> But doesn't look that much cleaner, tbh.

you really think so? I think it's much cleaner that way and also it's easier to understand, would you mind doing a rebase and opening a PR?

PR #337, opened 2019-09-03
Fixup 332

sjaeckel:
fixup #332


minad:
Concerning the docs - what happens in case of overflow? The number is still 0 terminated?  And the high digits are missing?

PR #341, opened 2019-09-05
Improve tests

sjaeckel:
a.k.a I found an error that I couldn't localize yet ... ~~and now I hope @czurnieden is quicker to find out why `mp_prime_is_prime()` doesn't work with the stripped-down version...~~ found and fixed it!

### DIFF
--- a/demo/test.c
+++ b/demo/test.c
@@ -977,7 +977,7 @@ static int test_mp_prime_is_prime(void)
       printf("A certified prime is a prime but mp_prime_is_prime says it is not.\n");
    }
    if ((err != MP_OKAY) || (cnt == MP_NO)) {
-      printf("prime tested was: ");
+      printf("prime tested was: 0x");
       mp_fwrite(&a,16,stdout);
       putchar('\n');
       goto LBL_ERR;
@@ -1011,10 +1011,10 @@ static int test_mp_prime_is_prime(void)
          printf("\nfrobenius-underwood says sub is not prime!\n");
       }
       if ((err != MP_OKAY) || (cnt == MP_NO)) {
-         printf("prime tested was: ");
+         printf("prime tested was: 0x");
          mp_fwrite(&a,16,stdout);
          putchar('\n');
-         printf("sub tested was: ");
+         printf("sub tested was: 0x");
          mp_fwrite(&b,16,stdout);
          putchar('\n');
          goto LBL_ERR;
@@ -1036,7 +1036,7 @@ static int test_mp_prime_is_prime(void)
       printf("\n\nissue #143 - mp_prime_strong_lucas_selfridge FAILED!\n");
    }
    if ((err != MP_OKAY) || (cnt == MP_NO)) {
-      printf("prime tested was: ");
+      printf("prime tested was: 0x");
       mp_fwrite(&a,16,stdout);
       putchar('\n');
       goto LBL_ERR;
--- a/demo/test.c
+++ b/demo/test.c
@@ -2220,53 +2220,58 @@ int unit_tests(int argc, char **argv)
       const char *name;
       int (*fn)(void);
    } test[] = {
-#define T(n) { #n, test_##n }
-      T(feature_detection),
-      T(trivial_stuff),
-      T(mp_get_set_i32),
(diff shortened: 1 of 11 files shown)


sjaeckel:
okay, I've found out that it's caused by the way the perl script creates `tommath_class.h` and misses the function pointers to the different reduce functions (the ones assigned to `redux`) in bn_s_mp_exptmod.c resp. bn_s_mp_exptmod_fast.c

The following hack + rerun `helper.pl -u` fixes this error...
```
diff --git a/bn_s_mp_exptmod_fast.c b/bn_s_mp_exptmod_fast.c
index 56e2b2b..fcee498 100644
--- a/bn_s_mp_exptmod_fast.c
+++ b/bn_s_mp_exptmod_fast.c
@@ -94,2 +94,6 @@ mp_err s_mp_exptmod_fast(const mp_int *G, const mp_int *X, const mp_int *P, mp_i
          goto LBL_M;
+         s_mp_montgomery_reduce_fast(NULL, NULL, 0);
+         mp_montgomery_reduce(NULL, NULL, 0);
+         mp_dr_reduce(NULL, NULL, 0);
+         mp_reduce_2k(NULL, NULL, 0);
       }
```

demo/test.c
```diff
@@ -1543,262 +1705,277 @@ static int test_mp_ilogb(void)
    int size;
    const uint32_t max_base = MP_MIN(UINT32_MAX, MP_DIGIT_MAX);
```
czurnieden: The conversion check complaint here (https://travis-ci.org/libtom/libtommath/jobs/581756157)
(No, I don't know how the extra forward slash came into existence at that place)
My error?

minad:
Hmm, I am not sure if I like those changes. Instead of adding MP_HAS checks to every test (a lot of boilerplate), just check for MP_HAS in the test registry and only add the functions to the table which are enabled. The other test__x functions won't be linked in.

minad:
~~~
registry [] = { MP_HAS(X) ? test_x : 0, ... }
~~~

demo/test.c
```diff
@@ -2219,54 +2224,61 @@ int unit_tests(int argc, char **argv)
...
-      T(mp_prime_rand),
-      T(mp_rand),
-      T(mp_read_radix),
-      T(mp_reduce_2k),
-      T(mp_reduce_2k_l),
+#define T_OFF           {"", NULL}
+#define T(n, o)         { #n, MP_HAS(o) ? test_##n : NULL }
+#define T2(n, o1, o2)   { #n, MP_HAS(o1) && MP_HAS(o2) ? test_##n : NULL }
+      T(feature_detection, TEST_FEATURE_DETECTION),
+      T(trivial_stuff, TEST_TRIVIAL_STUFF),
```
minad: Instead define T0, T1 and T2? Besides that it looks good.

sjaeckel: done!
now I've only to find out why that single build fails...

demo/test.c
```diff
@@ -59,6 +60,8 @@ static uint64_t uabs64(int64_t x)
  */
 void does_not_exist(void);
 
+#define BN_TEST_FEATURE_DETECTION_C
```
minad: If you define T0 you can avoid these unnecessary defines

demo/test.c
```diff
@@ -93,6 +96,8 @@ static int test_feature_detection(void)
    return EXIT_SUCCESS;
 }
 
+#define BN_TEST_TRIVIAL_STUFF_C
+
```
minad: Same here

etc/makefile
```diff
@@ -1,4 +1,5 @@
-CFLAGS += -Wall -W -Wextra -Wshadow -O3 -I../
+LTM_CFLAGS += -Wall -W -Wextra -Wshadow -O3 -I../
```
minad: Why complicating things here? the user could already add flags to CFLAGS, couldn't they?

sjaeckel: yeah but only through `CFLAGS="-foo" make` now it's also possible via `make CFLAGS="-foo"`

minad: Ok

PR #342, opened 2019-09-06
Refactor tune

sjaeckel:
As indicated by @MasterDuke17 via IRC it may make sense to introduce a tuning parameter for the new `mp_to_decimal()` implementation with a cutoff point similar to the ones for karatsuba etc.

I decided it'll make sense to clean-up the current tune implementation so we can probably plug this new cutoff-point calculation in there.

### DIFF
--- a/etc/tune.c
+++ b/etc/tune.c
@@ -249,6 +249,26 @@ struct cutoffs {
    int tcmul, tcsqr;
 };
 
+const struct cutoffs max_cutoffs =
+{ INT_MAX, INT_MAX, INT_MAX, INT_MAX };
+
+static void set_cutoffs(const struct cutoffs *c)
+{
+   KARATSUBA_MUL_CUTOFF = c->kmul;
+   KARATSUBA_SQR_CUTOFF = c->ksqr;
+   TOOM_MUL_CUTOFF = c->tcmul;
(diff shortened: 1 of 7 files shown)


minad:
Looks much better. You could also use MP_HAS. One could also take it further and maybe add a similar function registry like in test.c for the different cutoffs but this would probably be overengineering for this tiny utility.

etc/tune.c
```diff
@@ -473,160 +431,40 @@ int main(int argc, char **argv)
...
-            count--;
-         }
-      }
-      KARATSUBA_SQR_CUTOFF = x - s_stabilization_extra * increment_print;
+      s_run("Karatsuba squaring", s_time_sqr, &KARATSUBA_SQR_CUTOFF);
+      updated.ksqr = KARATSUBA_SQR_CUTOFF;
+      KARATSUBA_SQR_CUTOFF = INT_MAX;
 #endif
 #ifdef BN_S_MP_TOOM_MUL_C
-      if ((verbose == 1) || (testmode == 1)) {
```
minad: use MP_HAS instead of all the ifdefs?

sjaeckel: I did that already and forgot to push, sorry

PR #343, opened 2019-09-06
Addition of mp_radix_size_overestimate (behaviour like the function in GMP)

czurnieden:
Also changed mp_radix_size to use mp_ilogb while deprecating it at the same time and rewrote mp_fwrite to use mp_radix_sizeinbase instead.

See comment in `bn_mp_fwrite.c` for some of the consequences.

### DIFF
--- a/bn_mp_fwrite.c
+++ b/bn_mp_fwrite.c
@@ -33,6 +33,7 @@ mp_err mp_fwrite(const mp_int *a, int radix, FILE *stream)
       err = MP_ERR;
       goto LBL_ERR;
    }
+
    err = MP_OKAY;
 
 
--- a/bn_mp_fwrite.c
+++ b/bn_mp_fwrite.c
@@ -8,12 +8,13 @@ mp_err mp_fwrite(const mp_int *a, int radix, FILE *stream)
 {
    char *buf;
    mp_err err;
-   int len;
+   int len = 0;
    size_t written;
 
-   /* TODO: this function is not in this PR */
    if (MP_HAS(MP_RADIX_SIZE_OVERESTIMATE)) {
-      /* if ((err = mp_radix_size_overestimate(&t, base, &len)) != MP_OKAY)      goto LBL_ERR; */
+      if ((err = mp_radix_size_overestimate(a, radix, &len)) != MP_OKAY) {
+         return err;
+      }
    } else {
       if ((err = mp_radix_size(a, radix, &len)) != MP_OKAY) {
          return err;
@@ -35,7 +36,6 @@ mp_err mp_fwrite(const mp_int *a, int radix, FILE *stream)
    }
 
    err = MP_OKAY;
-
 
 LBL_ERR:
    MP_FREE_BUFFER(buf, (size_t)len);
--- a/bn_mp_radix_size.c
+++ b/bn_mp_radix_size.c
@@ -6,14 +6,14 @@
 /* returns size of ASCII representation */
 mp_err mp_radix_size(const mp_int *a, int radix, int *size)
 {
-
    mp_err err;
-   int digs;
-   mp_int   t;
-   mp_digit d;
    mp_int a_, b;
 
-   *size = 0;
+   if (size == NULL) {
+      return MP_VAL;
+   } else {
+      *size = 0;
+   }
 
    /* make sure the radix is in range */
    if ((radix < 2) || (radix > 64)) {
@@ -36,15 +36,9 @@ mp_err mp_radix_size(const mp_int *a, int radix, int *size)
       *size = bit_count/y;
       rem = bit_count - ((*size) * y);
       /* Add 1 for the remainder if any and 1 for "\0". */
-      *size += (rem == 0) ? 1 : 2;
-      /* And one extra character for the minus sign */
-      if (a->sign == MP_NEG) {
-         (*size)++;
-      }
+      *size += ((rem == 0) ? 1 : 2) + (a->sign == MP_NEG);
       return MP_OKAY;
    }
-
-
 
    if ((err = mp_init(&b)) != MP_OKAY) {
       goto LBL_ERR;
@@ -58,16 +52,8 @@ mp_err mp_radix_size(const mp_int *a, int radix, int *size)
 
    *size = (int)mp_get_l(&b);
 
-   /* mp_ilogb truncates to zero, hence we need one extra put on top. */
-   (*size)++;
-
-   /* The sign needs to have a place to live, too*/
-   if (a->sign == MP_NEG) {
-      (*size)++;
-   }
-
-   /* return digs + 1, the 1 is for the NUL byte that would be required. */
-   (*size)++;
+   /* mp_ilogb truncates to zero, hence we need one extra put on top and one for `\0`. */
+   *size += 2 + (a->sign == MP_NEG);
 
 LBL_ERR:
    mp_clear(&b);
--- a/bn_mp_radix_size.c
+++ b/bn_mp_radix_size.c
@@ -6,10 +6,12 @@
 /* returns size of ASCII representation */
 mp_err mp_radix_size(const mp_int *a, int radix, int *size)
 {
-   mp_err  err;
+
+   mp_err err;
    int digs;
    mp_int   t;
    mp_digit d;
+   mp_int a_, b;
 
    *size = 0;
 
@@ -23,42 +25,52 @@ mp_err mp_radix_size(const mp_int *a, int radix, int *size)
       return MP_OKAY;
    }
 
-   /* special case for binary */
-   if (radix == 2) {
-      *size = (mp_count_bits(a) + ((a->sign == MP_NEG) ? 1 : 0) + 1);
+   /* A small shortcut for powers of two. */
+   if (!(radix&(radix-1))) {
+      unsigned int x = (unsigned int)radix;
+      int y, bit_count, rem;
+      for (y=0; (y < 7) && !(x & 1u); y++) {
+         x >>= 1;
+      }
+      bit_count = mp_count_bits(a);
+      *size = bit_count/y;
+      rem = bit_count - ((*size) * y);
+      /* Add 1 for the remainder if any and 1 for "\0". */
+      *size += (rem == 0) ? 1 : 2;
+      /* And one extra character for the minus sign */
+      if (a->sign == MP_NEG) {
+         (*size)++;
+      }
       return MP_OKAY;
    }
 
-   /* digs is the digit count */
-   digs = 0;
 
-   /* if it's negative add one for the sign */
+
+   if ((err = mp_init(&b)) != MP_OKAY) {
+      goto LBL_ERR;
+   }
+
+   a_ = *a;
+   a_.sign = MP_ZPOS;
+   if ((err = mp_ilogb(&a_, (uint32_t)radix, &b)) != MP_OKAY) {
+      goto LBL_ERR;
+   }
+
+   *size = (int)mp_get_l(&b);
+
+   /* mp_ilogb truncates to zero, hence we need one extra put on top. */
+   (*size)++;
+
+   /* The sign needs to have a place to live, too*/
    if (a->sign == MP_NEG) {
-      ++digs;
+      (*size)++;
    }
 
-   /* init a copy of the input */
-   if ((err = mp_init_copy(&t, a)) != MP_OKAY) {
-      return err;
-   }
-
-   /* force temp to positive */
-   t.sign = MP_ZPOS;
-
-   /* fetch out all of the digits */
-   while (!MP_IS_ZERO(&t)) {
-      if ((err = mp_div_d(&t, (mp_digit)radix, &t, &d)) != MP_OKAY) {
-         goto LBL_ERR;
-      }
-      ++digs;
-   }
-
-   /* return digs + 1, the 1 is for the NULL byte that would be required. */
-   *size = digs + 1;
-   err = MP_OKAY;
+   /* return digs + 1, the 1 is for the NUL byte that would be required. */
+   (*size)++;
 
 LBL_ERR:
-   mp_clear(&t);
+   mp_clear(&b);
    return err;
 }
 
--- a/bn_mp_radix_size_overestimate.c
+++ b/bn_mp_radix_size_overestimate.c
@@ -1,0 +1,258 @@
+#include "tommath_private.h"
+#ifdef BN_MP_RADIX_SIZE_OVERESTIMATE_C
+/* LibTomMath, multiple-precision integer library -- Tom St Denis */
+/* SPDX-License-Identifier: Unlicense */
+
+
+
+/* LibTomMath, multiple-precision integer library -- Tom St Denis */
+/* SPDX-License-Identifier: Unlicense */
+
+/* returns size of ASCII representation */
+
+/*
+
+   Behaves like the function mp(n|z)_sizeinbase of GMP for all digit-sizes except MP_8BIT
+   and MP_16BIT (if the "int" for MP_16BIT has more than 16 bits) where it can overshoot by two.
+   The results for bases that are powers-of-two are exact.
+*/
+
+/*
+   Tables of {0, log_2([1..64])} times 2^p where p is the scale
+   factor defined in LTM_RADIX_SIZE_SCALE.
+
+   The tables have been computed with GP/PARI 2.9.4  with the
+   x86-64/GMP-5.1.3 kernel and (unnecessary) 100 decimal digits
+   precision. The fractional parts of the results after the
+   multiplication by 2^p have been discarded (floor(x)).
+
+   The number that measures the number of the digits of the bignum is
+   of type "int" which is at least 16 bits large and is 32 bit on the
+   architectures common at the time of writing. If you have hard- or
+   software that has a different size, please feel free to contact the
+   author, adding another table can be done quite quickly!
+ */
+
+/* *INDENT-OFF* */
+#ifdef MP_8BIT
+#define LTM_RADIX_SIZE_SCALE 29
+#define LTM_RADIX_SIZE_HALF_TABLE_SCALE 16
+static const mp_word logbases_high[65] = {
+       0u,     0u,  8192u, 12984u, 16384u,
+   19021u, 21176u, 22997u, 24576u, 25968u,
+   27213u, 28339u, 29368u, 30314u, 31189u,
+   32005u, 32768u, 33484u, 34160u, 34799u,
+   35405u, 35981u, 36531u, 37057u, 37560u,
+   38042u, 38506u, 38952u, 39381u, 39796u,
+   40197u, 40584u, 40960u, 41323u, 41676u,
+   42019u, 42352u, 42675u, 42991u, 43298u,
+   43597u, 43889u, 44173u, 44451u, 44723u,
+   44989u, 45249u, 45503u, 45752u, 45995u,
+   46234u, 46468u, 46698u, 46923u, 47144u,
+   47360u, 47573u, 47783u, 47988u, 48190u,
+   48389u, 48584u, 48776u, 48965u, 49152u
+};
+static const mp_word logbases_low[65] = {
+       0u,     0u,     0u,   839u,     0u,
+   15397u,   839u, 55805u,     0u,  1678u,
+   15397u, 43504u,   839u,   142u, 55805u,
+   16237u,     0u, 32479u,  1678u,  1454u,
+   15397u, 56644u, 43504u,  1280u,   839u,
+   30795u,   142u,  2517u, 55805u, 38031u,
+   16237u, 50867u,     0u, 44343u, 32479u,
+    5667u,  1678u, 55179u,  1454u,   981u,
+   15397u,  4326u, 56644u, 62971u, 43504u,
+   17076u,  1280u, 12574u,   839u, 46074u,
+   30795u, 33318u,   142u,  8150u,  2517u,
+   58902u, 55805u,  2293u, 38031u, 40098u,
+   16237u, 39339u, 50867u, 57483u,     0u
+};
+#elif (defined MP_16BIT)
+#define LTM_RADIX_SIZE_SCALE 61
+#define LTM_RADIX_SIZE_HALF_TABLE_SCALE 32
+static const mp_word logbases_high[65] = {
+            0ul,          0ul,  536870912ul,  850920263ul, 1073741824ul,
+   1246575653ul, 1387791175ul, 1507187197ul, 1610612736ul, 1701840526ul,
+   1783446565ul, 1857268208ul, 1924662087ul, 1986658446ul, 2044058109ul,
+   2097495917ul, 2147483648ul, 2194439903ul, 2238711438ul, 2280588718ul,
+   2320317477ul, 2358107460ul, 2394139120ul, 2428568832ul, 2461532999ul,
+   2493151307ul, 2523529358ul, 2552760789ul, 2580929021ul, 2608108687ul,
+   2634366829ul, 2659763891ul, 2684354560ul, 2708188471ul, 2731310815ul,
+   2753762851ul, 2775582350ul, 2796803979ul, 2817459630ul, 2837578709ul,
+   2857188389ul, 2876313830ul, 2894978372ul, 2913203707ul, 2931010032ul,
+   2948416180ul, 2965439744ul, 2982097182ul, 2998403911ul, 3014374394ul,
+   3030022219ul, 3045360166ul, 3060400270ul, 3075153878ul, 3089631701ul,
+   3103843862ul, 3117799933ul, 3131508981ul, 3144979599ul, 3158219938ul,
+   3171237741ul, 3184040363ul, 3196634803ul, 3209027723ul, 3221225472ul
+};
+static const mp_word logbases_low[65] = {
+            0ul,          0ul,          0ul, 1065013491ul,          0ul,
+   3868050815ul, 1065013491ul, 1343271782ul,          0ul, 2130026983ul,
+   3868050815ul,  427100031ul, 1065013491ul, 1200147405ul, 1343271782ul,
+    638097011ul,          0ul, 1495425743ul, 2130026983ul, 1083835016ul,
+   3868050815ul, 2408285274ul,  427100031ul, 3594167896ul, 1065013491ul,
+   3441134334ul, 1200147405ul, 3195040475ul, 1343271782ul, 1429396043ul,
+    638097011ul, 1651143338ul,          0ul, 1492113523ul, 1495425743ul,
+    916355301ul, 2130026983ul, 1832592635ul, 1083835016ul, 2265160897ul,
+   3868050815ul, 3464916779ul, 2408285274ul, 2621526098ul,  427100031ul,
+   1703110503ul, 3594167896ul, 2513440227ul, 1065013491ul, 2686543564ul,
+   3441134334ul, 2560439235ul, 1200147405ul, 1309103046ul, 3195040475ul,
+       183550ul, 1343271782ul, 2148848508ul, 1429396043ul, 3785381171ul,
+    638097011ul, 1073048670ul, 1651143338ul, 3473298766ul,          0ul
+};
+
+/* TODO: This assumes the type "long long" exists. If it doesn't: use the tables for MP_16BIT */
+#elif ( (defined MP_28BIT) || (defined MP_31BIT) || (defined MP_32BIT)  || (defined MP_64BIT) )
+#define LTM_RADIX_SIZE_SCALE 61
+#ifdef MP_64BIT
+static const mp_digit logbases[65] = {
+#else
+static const mp_word logbases[65] = {
+#endif
+                      0ull,                    0ull,  2305843009213693952ull,
+    3654674702153732339ull,  4611686018427387904ull,  5354001665492895103ull,
+    5960517711367426291ull,  6473319721408181094ull,  6917529027641081856ull,
+    7309349404307464679ull,  7659844674706589055ull,  7976906213687625599ull,
+    8266360720581120243ull,  8532633055092329421ull,  8779162730621875046ull,
+    9008676367646627443ull,  9223372036854775808ull,  9425047617917838031ull,
+    9615192413521158631ull,  9795053960520401544ull,  9965687683920283007ull,
+   10127994423561913434ull, 10282749222901319551ull, 10430623713119086168ull,
+   10572203729794814195ull, 10708003330985790206ull, 10838476064306023373ull,
+   10964024106461197019ull, 11085005739835568998ull, 11201741516507896395ull,
+   11314519376860321395ull, 11423598928577852074ull, 11529215046068469760ull,
+   11631580915841357939ull, 11730890627131531983ull, 11827321386901076197ull,
+   11921035422734852583ull, 12012181624960263419ull, 12100896969734095496ull,
+   12187307757246061761ull, 12271530693133976959ull, 12353673836347420459ull,
+   12433837432775607386ull, 12512114650772492370ull, 12588592232115013503ull,
+   12663351069800359783ull, 12736466722332780120ull, 12808009872697200099ull,
+   12878046739008508147ull, 12946639442816362188ull, 13013846340199484158ull,
+   13079722320071570371ull, 13144319073519717325ull, 13207685337486676934ull,
+   13269867115674890971ull, 13330907879180520702ull, 13390848749049262950ull,
+   13449728662674133884ull, 13507584525721590347ull, 13564451351070528819ull,
+   13620362386074015347ull, 13675349229302017118ull, 13729441937791546026ull,
+   13782669125715645774ull, 13835058055282163712ull
+};
+#else
+#  ifdef _MSC_VER
+#    pragma message("mp_radix_size_overestimate: unknown or undefined MP_xBIT type")
+#  else
+#    warning "mp_radix_size_overestimate: unknown or undefined MP_xBIT type"
+#  endif
+#endif
+/* *INDENT-ON* */
+
+mp_err mp_radix_size_overestimate(const mp_int *a, const int base, int *size)
+{
+   mp_int bi_bit_count, bi_k;
+#if ( (defined MP_8BIT) || (defined MP_16BIT) )
+   mp_int bi_k_bis;
+#endif
+   int rem, bit_count;
+   mp_err e = MP_OKAY;
+
+   if ((base < 2) || (base > 64)) {
+      *size = 0;
+      return MP_VAL;
+   }
+   /* floor( log_2(a) ) + 1 */
+   bit_count = mp_count_bits(a);
+
+   if (bit_count == 0) {
+      *size = 2;
+      return MP_OKAY;
+   }
+   if (base == 2) {
+      *size = bit_count + 2;
+      return MP_OKAY;
+   }
+
+   /* A small shortcut for powers of two. */
+   if (!(base&(base-1))) {
+      unsigned int x = (unsigned int)base;
+      int y;
+      for (y=0; (y < 7) && !(x & 1u); y++) {
+         x >>= 1;
+      }
+      *size = bit_count/y;
+      rem = bit_count - ((*size) * y);
+      /* Add 1 for the remainder if any and 1 for "\0". */
+      *size += (rem == 0) ? 1 : 2;
+      /* And one extra character for the minus sign */
+      if (a->sign == MP_NEG) {
+         (*size)++;
+      }
+      return MP_OKAY;
+   }
+
+   /* d(bitcount,base) = floor( (bitcount * 2^p) / k(base) ) + 1 */
+
+   if ((e = mp_init_multi(&bi_bit_count, &bi_k, NULL)) != MP_OKAY) {
+      *size = 0;
+      return e;
+   }
+#if ( (defined MP_8BIT) || (defined MP_16BIT) )
+   if ((e = mp_init(&bi_k_bis)) != MP_OKAY) {
+      /* No "goto" to avoid cluttering this code with even more preprocessor branches */
+      mp_clear_multi(&bi_bit_count, &bi_k, NULL);
+      *size = 0;
+      return e;
+   }
+#endif
+   /* "long" is at least as large as "int" according to even the oldest C standards */
+   mp_set_l(&bi_bit_count, bit_count);
+#if ( (defined MP_8BIT) || (defined MP_16BIT) )
+   /* There is no mp_set_u16 for MP_8BIT */
+   mp_set_u32(&bi_k, logbases_high[base]);
+   if ((e = mp_mul_2d(&bi_k, LTM_RADIX_SIZE_HALF_TABLE_SCALE, &bi_k)) != MP_OKAY) {
+      *size = 0;
+      goto LTM_E1;
+   }
+   mp_set_u32(&bi_k_bis, logbases_low[base]);
+   if ((e = mp_add(&bi_k_bis, &bi_k, &bi_k)) != MP_OKAY) {
+      *size = 0;
+      goto LTM_E1;
+   }
+#else
+   mp_set_u64(&bi_k, logbases[base]);
+#endif
+   if ((e = mp_mul_2d(&bi_bit_count, LTM_RADIX_SIZE_SCALE, &bi_bit_count)) != MP_OKAY) {
+      *size = 0;
+      goto LTM_E1;
+   }
+   if ((e = mp_div(&bi_bit_count, &bi_k, &bi_bit_count, NULL)) != MP_OKAY) {
+      *size = 0;
+      goto LTM_E1;
+   }
+   /*
+      The first addition of one is done because of the truncating last division,
+      and the second addition of one is done to make room for NIL ('\0').
+
+      Casting from "long" to "int" can be done because "bi_bit_count" fits into an "int"
+      by definition.
+    */
+   *size = (int)mp_get_l(&bi_bit_count) + 1 + 1 ;
+
+#if ( (defined MP_8BIT) && (INT_MAX > 0xFFFF))
+   /* TODO: Add a third table? But how likely is it that "int" is 32-bit in
+            an 8-bit environment? */
+   /* diff. is 3 bits, hence add 2 (two) */
+   *size += 2;
+#endif
+
+   /* And before we forget it: one extra character for the minus sign */
+   if (a->sign == MP_NEG) {
+      (*size)++;
+   }
+
+LTM_E1:
+   mp_clear_multi(&bi_bit_count, &bi_k, NULL);
+#if ( (defined MP_8BIT) || (defined MP_16BIT) )
+   mp_clear(&bi_k_bis);
+#endif
+   return e;
+}
+
+
+
+
+#endif
--- a/bn_mp_radix_size_overestimate.c
+++ b/bn_mp_radix_size_overestimate.c
@@ -141,17 +141,22 @@ static const mp_word logbases[65] = {
 #endif
 /* *INDENT-ON* */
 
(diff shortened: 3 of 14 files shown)


czurnieden:
Oh, using deprectated functions (added test for `mp_radix_size` in `test.c`) is a full error, ok.
But only with GCC not clang?

I don't understand the Windows error but I assume it is for the same reason?

czurnieden:
May I propose (again) to kill `MP_8BIT`?
'Cause it's killing me! ;-)

sjaeckel:
> May I propose (again) to kill `MP_8BIT`?
> 'Cause it's killing me! ;-)

let it rot, disable what doesn't work and add it to the deprecation list, we'll remove it after the next release

czurnieden:
> let it rot, disable what doesn't work and add it to the deprecation list, we'll remove it after the next release

One of the handful of 8-bit MCUs which have enough Flash and RAM to run a stripped down LTM with enough circulation to make it worth the hassle are the bigger versions of ATxmega (and the biggest versions of ATmega to some extent). Will see if I can get my hands on one the next couple of days. No virtual MCUs here, I need the power consumption with the internal clock (datasheet gives only the values with an external clock?), too.
Is there some kind of main forum where most of the fanbois sit?

And how much work would it be?

```text
czurnieden ~/GITHUB/libtommath O (radix_sizeinbase)$ find ./ -type f  -name '*c' -exec grep  MP_8BIT {} \; -print
#ifdef MP_8BIT
       * MP_8BIT (It is unknown if the Lucas-Selfridge test works with 16-bit
#if defined (MP_8BIT) || defined (LTM_USE_FROBENIUS_TEST)
#ifdef MP_8BIT
#ifdef MP_8BIT
./bn_mp_prime_is_prime.c
#ifndef MP_8BIT
#ifdef MP_8BIT
#ifdef MP_8BIT
./demo/test.c
#ifdef MP_8BIT
./demo/main.c
#ifndef MP_8BIT
./bn_mp_read_unsigned_bin.c
#ifndef MP_8BIT
./bn_mp_to_unsigned_bin.c
#ifndef MP_8BIT
   /* CZ TODO: Some of them need the full 32 bit, hence the (temporary) exclusion of MP_8BIT */
./bn_mp_prime_strong_lucas_selfridge.c
#ifndef MP_8BIT
./bn_prime_tab.c
      Since mp_radix_sizeinbase() can overshoot by one (two with MP_8BIT)
./bn_mp_fwrite.c
#if !defined(MP_8BIT)
#if defined(MP_64BIT) || !(defined(MP_8BIT) || defined(MP_16BIT))
./bn_mp_montgomery_setup.c
#ifdef MP_8BIT
#if ( (defined MP_8BIT) || (defined MP_16BIT) )
#if ( (defined MP_8BIT) || (defined MP_16BIT) )
#if ( (defined MP_8BIT) || (defined MP_16BIT) )
   /* There is no mp_set_u16 for MP_8BIT */
#if ( (defined MP_8BIT) && (INT_MAX > 0xFFFF))
#if ( (defined MP_8BIT) || (defined MP_16BIT) )
./bn_mp_radix_sizeinbase.c
#ifdef MP_8BIT
./mtest/mtest.c
#ifdef MP_8BIT
      /* (32764^2 - 4) < 2^31, no bigint for >MP_8BIT needed) */
./bn_mp_prime_frobenius_underwood.
``` 

Oh, that's not much, thought it to be more!
Plus wiki and "issue"-entry, `bn.tex`, ChangeLog, and the little hints you metioned here and there (deprecation).
Nothing I would call exhausting.
OK. let's make it official (in another PR).

Oh, am away tonight, now, actually, (brother-in-law has his 50th and who says no to a free dinner? ;-) ), so till tomorrow.

sjaeckel:
> OK. let's make it official (in another PR).

:+1: 

> Oh, am away tonight, now, actually, (brother-in-law has his 50th and who says no to a free dinner? ;-) ), so till tomorrow.

Enjoy & have fun!

bn_mp_fwrite.c
```diff
@@ -19,10 +19,33 @@ mp_err mp_fwrite(const mp_int *a, int radix, FILE *stream)
       return MP_MEM;
    }
 
+   /*
+      Since mp_radix_sizeinbase() can overshoot by one (two with MP_8BIT)
```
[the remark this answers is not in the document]
czurnieden: > The string is NUL terminated so everything is fine IMO.

It is more or less only a problem with `fwrite` which would write more than expected by the user. Not a problem if it writes to `stdout` or `stderr` but might result in a lot of cussing&cursing if used elsewhere.

> To get the correct length, why not change the length parameter of `mp_to_radix()` to a `size_t*`?

Yes, changing `mp_to_radix` would be the simplest way to do it because we already now the exact number at that time. The API has already been changed and not that long ago, so it wouldn't be a problem to do it again, I think.

OK, do it here or in another PR? (Somebody should send @MasterDuke17 a note which I hereby did. If Gihub's "you got mentioned" notification mechanism still works, that is).

Ah, all green. Squash or add the `mp_to_radix()` change?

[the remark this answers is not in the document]
MasterDuke17: Hm, does this mean anything for my PR? It looks like my new `mp_to_decimal` already has its maxlen as `size_t`.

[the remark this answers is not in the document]
czurnieden: @MasterDuke17 
> Hm, does this mean anything for my PR?

Yes, you would need to put the exact amount of digits plus the zero back in `maxlen`.

But wait until it's actually done.

> It looks like my new `mp_to_decimal` already has its maxlen as `size_t`.

You would need to change the prototype from
```c
mp_err mp_to_decimal(const mp_int *a, char *str, size_t maxlen);
```
to
```c
mp_err mp_to_decimal(const mp_int *a, char *str, size_t *maxlen);
```
(and change the rest according to that.)

So nothing big, I just thought it would be nice if you get a warning to avoid uncomfortable surprises and can prepare for it in advance.

Thank you for your patience with us!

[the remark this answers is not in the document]
MasterDuke17: Ah, I see. Ok, I'll make those changes to my PR after this PR is merged.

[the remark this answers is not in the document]
sjaeckel: > OK, do it here or in another PR?

please in another PR

sjaeckel:
> How about calling this `mp_radix_size_approx()`?

or `mp_radix_size_fast()`?

bn_mp_fwrite.c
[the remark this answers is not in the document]
czurnieden: > please in another PR

In #348 



> >   How about calling this `mp_radix_size_approx()`?
> or `mp_radix_size_fast()`?

The name `sizeinbase` is shamelessly stolen from GMP where it also shows the same behaviour and because programmers always have a hard time to come up with new and exiting names I took that instead.
Yeah, call me lazy, that's what I am ;-)

minad:
mp_radix_size_overestimate?

czurnieden:
> mp_radix_size_overestimate?

\*grmbl\*

This commit is just the addition of  `mp_radix_size_overestimate`, all the rest will be handled in #348 if necessary.

(`helper.pl --update-files` does not seem to format `tommath.def` and `tommath_class.h` correctly. Is that a known problem?)

minad:
@czurnieden the lookup tables look rather large. Could we use smaller tables and a more Cruse estimate? I think that might be preferable for embedded systems. We will pay more for the allocations then but the constant costs of the tables would be smaller.

However allocators are usually rounding themselves so a more crude estimate might not be too bad.

czurnieden:
>  the lookup tables look rather large

They are 520 bytes(!) large (260 for 8-bit as long as we have MP_8BIT).
Half a kilobyte. We will drop 8-bit support in the near future, so the most memory-restricted MCUs will not get supported anymore at that time.

But I know where you are coming from and as I am there myself, too, from time to time, I understand.
Mmh&hellip;
I can offer a stripped down version with the same accuracy that is much smaller but only supports all powers of two (as they are calculated directly) and a hardcoded base 10. This would get rid of the table(s) and all the logic to handle the tables and will safe about 1k (probably less).

> However allocators are usually rounding themselves so a more crude estimate might not be too bad.

You can split the tables in halve (e.g.: using a 16-bit type instead of a 32-bit one) but that would loose way too much of the accuracy. Yes, I played with it while designing this function but had a hard time to guarantee the `-0` in the `-0/+x` tolerance with such insufficient accuracy without an "extra" on top. How large does that "extra" need to be? Hard to tell without a proper error-analysis.

The alternatives, e.g: a fixed-point log, would add a lot of logic which I doubt needs less memory than the tables.

So the only thing I can come up with is the functionally stripped down version described above.

Which lead me to the question: do we really need all the bases? We got a quick base-10 (didn't take a closer look, but looks quite hardcoded, don't know how much work it would be to extend it) and we can rewrite the old loop to get the power-of-two bases directly.

minad:
Ok. 500 bytes are small enough I guess. What's the status?

minad:
We could add a configuration option? MP_RADIX_ALL_BASES? Otherwise only enable 10 and power of 2?

czurnieden:
> What's the status?

We are only waiting for your OK, so if you are OK with it, drop me a note and I will squash&finish.

> We could add a configuration option? MP_RADIX_ALL_BASES? Otherwise only enable 10 and power of 2?

So no "squash&finish" here?
But serious: what kind of configuration, just the macro `MP_RADIX_ALL_BASES` or something a bit more sophisticated (whatever that might be)? I haven't  implemented the `only base 10 and powers of 2` yet, please give me a day (which means "wee hours of the night" in my case ;-) ) to do so.

czurnieden:
@minad just a quick hack as an example. Do you want something like that?

minad:
Hmm I am not sure if I like further complications of this function

czurnieden:
> > > We could add a configuration option? MP_RADIX_ALL_BASES? Otherwise only enable 10 and power of 2?
> > just a quick hack as an example
> Hmm I am not sure if I like further complications of this function

Uhm&hellip;ooookaaaay?

So, can I take that as the answer "No." to my question
>  Do you want something like that?
?

minad:
@sjaeckel @czurnieden What do you think? Is this ready or push this after 1.2 since it is a strict addition? I would love if we could get it out sooner than later since there are already so many changes. After that we can get rid of the deprecated stuff hopefully. 2.0 can then take a while to mature before release, including additions like the following or faster mp_to_radix. Since 1.2 is more like a backward compatibility release I think that would be fine.

czurnieden:
> Is this ready or push this after 1.2 since it is a strict addition?

It was ready for 1.2 but isn't anymore with the new branch `MP_RADIX_ALL_BASES`. We need to either define `MP_RADIX_ALL_BASES` per default, invert the branch with  `MP_RADIX_REDUCED_BASES`, or adapt `mp_to_radix`, too (a bit more work but not *that* much).

But the biggest problem ist the reduction of the functionality itself. Not much has been changed up to commit   https://github.com/libtom/libtommath/commit/724db0ac889f72794a7e8fbbb66b17734b207b51 that couldn't get squeezed into 1.2 but reducing functionality is a bit much. Either invert the logic (`MP_RADIX_REDUCED_BASES`) or roll back to  https://github.com/libtom/libtommath/commit/724db0ac889f72794a7e8fbbb66b17734b207b51 for 1.2

minad:
MP_RADIX_BASES was just me thinking too loud. As I said, i am not sure if we want that due to the complications it brings. But we can still discuss all that.

My point was more to get 1.2 out at some point since I am getting worried with all the changes. 2.0 would give us a bit of a clean slate (think size_t issues, 8bit gone etc).

czurnieden:
Rolled back for now.
Would propose to do the radix reducing in a different PR to get it all together, including the fast base-10 conversion.

czurnieden:
> My point was more to get 1.2 out at some point since I am getting worried with all the changes (API types, deprecations etc etc). 2.0 would give us a bit of a clean slate (think size_t issues, 8bit gone etc).

Don't forget that we also can do versions 1.3, 1.4, 1.5 and so on! You can stop at every point where you think that we reached a nice stable(!) milestone and put out a new minor version.
We could, for example, stop here, get all of the PRs that are stable & wanted, put them in version 1.2, get it out and start working at version 1.3.

Version 1.3 could be the "Call for enhanced velocity" (yes, I'm afraid of EA ;-) ) for example?
(No, I don't want to give versions a name, I hate that!)

The changes now are more or less small and incremental, the average user would have not much work to do to adapt their product to the new LTM version 1.2 if at all.

Doing something massive, like dropping 8-bit support or reducing functionality (e.g. the number of radices in in number conversion), or removing all of the deprecated functions deserves indeed a major version increment, 2.0 in our case.

minad:
Well, the question is when we want to introduce breaking changes. Some changes cannot be introduced easily, like introducing size_t in mp_int for example. This should go into 2.0.
If we are doing 1.3, 1.4 etc these changes won't happen for a long time. Furthermore we have to maintain all the cruft for these versions too. Given that we have PRs tagged as 2.0, I guess we start working on 2.0 in the develop branxh after the release of 1.2.

czurnieden:
> If we are doing 1.3, 1.4 etc these changes won't happen for a long time.

Yepp, naturally occurring procrastination is indeed one of the bigger disadvantages of that concept.

I know it is a week argument but nevertheless: a lot of projects would have released at least half a dozen intermediate versions with that amount of commits we have hauled up in develop now ;-)

minad:
I am closely following develop with my projects - so I don't bother too much about the releases and I would rather have faster releases and breaking changes. But I have to admit that @sjaeckel's approach of not breaking things has also great merits. By discussing things, I feel we end up with pretty good solutions in the end. But always keeping the version backward compatible will not allow us to reach certain optima. At some point we have to do the step and do some breaks, there is no real way around it (except if you perform a pure append only development style where even broken stuff is kept forever).

bn_mp_radix_size_overestimate.c
```diff
@@ -0,0 +1,333 @@
...
+   12878046739008508147ull, 12946639442816362188ull, 13013846340199484158ull,
+   13079722320071570371ull, 13144319073519717325ull, 13207685337486676934ull,
+   13269867115674890971ull, 13330907879180520702ull, 13390848749049262950ull,
+   13449728662674133884ull, 13507584525721590347ull, 13564451351070528819ull,
+   13620362386074015347ull, 13675349229302017118ull, 13729441937791546026ull,
+   13782669125715645774ull, 13835058055282163712ull
+};
+#else
+#  ifdef _MSC_VER
+#    pragma message("mp_estimate_string_size: unknown or undefined MP_xBIT type")
```
minad: mp_estimate_string_size?

bn_mp_radix_size_overestimate.c
```diff
@@ -0,0 +1,333 @@
...
+#  else
+#    warning "mp_estimate_string_size: unknown or undefined MP_xBIT type"
+#  endif
+#endif
+#if 0
```
minad: #if 0? Why add dead code?

czurnieden: Because: 
 - it was quite fuddly to compute, didn't want to dump it so easily.
 - it is also not "dead code" in its strictest meaning because it must be used for 64-bit `int`s.
 - it shows some optimism.
 - it is futureproof until 2105. According to Nkasiogi Koslowsky who will have invented a time-machine in 2091, went to 2106 and nicked a mobile. At least that's what she told me but I doubt it because she said that it was an Apple.

But I can delete it, if you insist?

bn_mp_radix_size_overestimate.c
czurnieden: Oopsie ;-)

bn_mp_radix_size.c
```diff
@@ -23,42 +22,66 @@ mp_err mp_radix_size(const mp_int *a, int radix, int *size)
       return MP_OKAY;
    }
 
-   /* special case for binary */
-   if (radix == 2) {
-      *size = mp_count_bits(a) + ((a->sign == MP_NEG) ? 1 : 0) + 1;
+   /* A small shortcut for powers of two. */
```
minad: This looks kind of ugly. Could you check for power of two instead via bitwise and then shift bit_count in a for loop iterating over the radix zero bits?

bn_mp_radix_size_overestimate.c
```diff
@@ -0,0 +1,273 @@
...
+      ln2 = 5;
+      break;
+   case 64:
+      ln2 = 6;
+      break;
+   default:
+      ln2 = 0;
+      break;
+   }
+   if (ln2 != 0) {
```
minad: Same here.

bn_mp_radix_size.c
czurnieden: > This looks kind of ugly.

Beauty lies in the eyes of the beholder! ;-)

What do you mean by "bitwise"? Something like `(x != 0) && (!(x&(x-1)))`? And then count the zero LSB's?
Can do, if you insist?

minad: Yes, I think I would like that more.

bn_mp_radix_size_overestimate.c
minad: Hmm but why is it then not enabled for 64 bit ints. How do you detect int size?

bn_mp_radix_size.c
czurnieden: Well, now it's me who finds it a bit ugly. ;-)

bn_mp_radix_size_overestimate.c
czurnieden: > How do you detect int size?

I don't think that there will be 64-bit large `int`s in the near future and I don't need the size elswhere in that function, so I just don't check the size of an `int` and wait until someone complains.

bn_mp_radix_size.c
minad: Puuh but you could make it better. Why is rem a toplevel variable. Why not use it directly in the condition? Radix should probably also be unsigned in the arguments. Since we are deprecating the old function we can also change that?

minad: I like it more since it looks more general, no hardcoded lut.

czurnieden: > Why is rem a toplevel variable. Why not use it directly in the condition?

It was a simple C&P because I wasn't sure if that's exactly what you wanted. If you want it that way I can put a small speck of Vienna chalk on it and buff it a little.

minad: Whatever :)

czurnieden:
> Radix should probably also be unsigned in the arguments. Since we are deprecating the old function we can also change that?

I'd like to wait until all string-functions came together to have a single PR which would make such changes a bit simpler.

minad:
mp_radix_size/mp_radix_size_overestimate should also return size_t since they are referring to buffer sizes. In 2.0 we should do these breaking type changes.

bn_mp_radix_size_overestimate.c
```diff
@@ -0,0 +1,258 @@
...
+#if ( (defined MP_8BIT) && (INT_MAX > 0xFFFF))
+   /* TODO: Add a third table? But how likely is it that "int" is 32-bit in
+            an 8-bit environment? */
+   /* diff. is 3 bits, hence add 2 (two) */
+   *size += 2;
+#endif
+
+   /* And before we forget it: one extra character for the minus sign */
+   if (a->sign == MP_NEG) {
+      (*size)++;
```
minad: Why not always add 1 for the sign? We are overestimating?

bn_mp_radix_size_overestimate.c
```diff
@@ -0,0 +1,258 @@
...
+      return MP_VAL;
+   }
+   /* floor( log_2(a) ) + 1 */
+   bit_count = mp_count_bits(a);
+
+   if (bit_count == 0) {
+      *size = 2;
+      return MP_OKAY;
+   }
+   if (base == 2) {
```
minad: Here you always add one for the sign. Why special case for 2? Should be handled by power of two code.

czurnieden: > Here you always add one for the sign

No, it is the plus one in `floor( log_2(a) ) + 1 ` and plus one for `\0`

> Why special case for 2? Should be handled by power of two code.

Yes, that's right, needs to be merged.

bn_mp_radix_size_overestimate.c
czurnieden: > Why not always add 1 for the sign? We are overestimating

Overestimating by at most 1 (one) like the GMP function (with the exception for MP_8BIT where it is 3 at most). That was my goal, it had quite some work involved to reach it and I won't move any posts here.

bn_mp_radix_size_overestimate.c
minad: Then you are missing the sign?

bn_mp_radix_size_overestimate.c
minad: Why not. As I said before I would rather like a simpler function with a less precise estimate.

bn_mp_radix_size_overestimate.c
czurnieden: > Here you always add one for the sign

No I didn't add the one for the sign at all for base 2, that was a bug you found!
Thanks!

bn_mp_radix_size_overestimate.c
```diff
@@ -0,0 +1,254 @@
...
+   mp_err e = MP_OKAY;
+
+   if ((base < 2) || (base > 64)) {
+      *size = 0;
+      return MP_VAL;
+   }
+   /* floor( log_2(a) ) + 1 */
+   bit_count = mp_count_bits(a);
+
+   if (bit_count == 0) {
```
minad: Same here!

czurnieden: Yes, I could check for zero  before counting the bits (although `mp_count_bits` has a shortcut for it, too). Shall I?
Or even before the test if the base given is in range?

bn_mp_radix_size_overestimate.c
czurnieden: Why am I missing push-notes (at least with Firefox)? This website does not update regularily, The "Then your are missing the sign" came by email first, the "same here" came as a a push-note immediately!
Do I have to change to "Edge"? ;-)

bn_mp_radix_size_overestimate.c
czurnieden: > As I said before I would rather like a simpler/more compact function with a less precise estimate.

By saving a single branch, probably less than half a dozen CPU operations?
Really?

bn_mp_radix_size_overestimate.c
minad: Maybe move the zero check before count bits.

bn_mp_radix_size_overestimate.c
```diff
@@ -0,0 +1,254 @@
...
+
+   /* A small shortcut for powers of two. */
+   if (!(base&(base-1))) {
+      unsigned int x = (unsigned int)base;
+      int y;
+      for (y=0; (y < 7) && !(x & 1u); y++) {
+         x >>= 1;
+      }
+      *size = bit_count/y;
+      rem = bit_count - ((*size) * y);
```
minad: rem scope. Could you please keep the two functions in sync?

bn_mp_radix_size_overestimate.c
minad: It is more a question of intend. I would write it differently, but it doesn't really matter. It should reduce to a single add operation.

*size += sign;

Or

*size += sign ? 1 : 0;

bn_mp_radix_size_overestimate.c
```diff
@@ -0,0 +1,254 @@
...
+
+   if ((e = mp_init_multi(&bi_bit_count, &bi_k, NULL)) != MP_OKAY) {
+      *size = 0;
+      return e;
+   }
+#if ( (defined MP_8BIT) || (defined MP_16BIT) )
+   if ((e = mp_init(&bi_k_bis)) != MP_OKAY) {
+      /* No "goto" to avoid cluttering this code with even more preprocessor branches */
+      mp_clear_multi(&bi_bit_count, &bi_k, NULL);
+      *size = 0;
```
sjaeckel: I just asked myself if 0 is really the good value to return on error... whatever we choose, for sure we should set it once on function-entry and not x-times on each error-case

bn_mp_radix_size_overestimate.c
sjaeckel: > Or even before the test if the base given is in range

no, base not in range is an error

whether to do a `MP_IS_ZERO` or call `mp_count_bits()` first I think it doesn't really matter, the code stays nearly the same, or not!?

bn_mp_radix_size_overestimate.c
sjaeckel: Trying to be GMP compatible makes somehow sense, but also I wouldn't care if we'd choose to be sometimes off-by-two or even off-by-X if it would make the implementation a lot easier.

@minad Doesn't this boil down to only a style question here? Or what do you see as an advantage of writing `*size += sign ? 1 : 0;` over the way it is now?

bn_mp_radix_size_overestimate.c
sjaeckel: speaking of "keep in sync" shouldn't `base` then be called `radix` or the other way around? :-)

bn_mp_radix_size_overestimate.c
minad: No advantage. I would write *size += 1 :)

bn_mp_radix_size_overestimate.c
minad: Ok.

bn_mp_radix_size.c
```diff
@@ -23,42 +21,52 @@ mp_err mp_radix_size(const mp_int *a, int radix, int *size)
...
+   if (!(radix&(radix-1))) {
```
sjaeckel: how about making a private function out of this?

I'd say we don't even have put it in a separate file, simply here on the top and the prototype to `tommath_private.h`!?

minad: Yes I also thought about that.

bn_mp_radix_size_overestimate.c
czurnieden: > if it would make the implementation a lot easier.

```text
$ gcc --version
gcc (Ubuntu 7.4.0-1ubuntu1~18.04.1) 7.4.0
$ gcc -g3 -O3 -c bn_mp_radix_size_overestimate.c
$ objdump -d -S bn_mp_radix_size_overestimate.o
...
   if (a->sign == MP_NEG) {
 1e6:	83 7d 08 01          	cmpl   $0x1,0x8(%rbp)
 1ea:	75 91                	jne    17d <mp_radix_size_overestimate+0x17d>
      (*size)++;
 1ec:	83 c0 03             	add    $0x3,%eax
 1ef:	41 89 04 24          	mov    %eax,(%r12)
 1f3:	eb 88                	jmp    17d <mp_radix_size_overestimate+0x17d>
 1f5:	0f 1f 00             	nopl   (%rax)
...
```

If you want to set `*size = 1` at the start instead you save about 5 CPU-cycles (3 if you compile with `-Os) with my AMD A8-6600K.
In the pale face of the imminent danger to repeat myself: really?
But your definition of "a **lot** easier" [emphasis by me] might differ ;-)

> Doesn't this boil down to only a style question here?

Yes, and that means: I don't care. Just tell me how you want it and I'll do it and if I feel especially adventurous at that time I might even add a note to the wiki-page ;-)

bn_mp_radix_size_overestimate.c
czurnieden: > speaking of "keep in sync" shouldn't base then be called radix or the other way around?

\*grmbl\*
;-)

> Could you please keep the two functions in sync?

I still would like to to deprecate `mp_radix_size` because it just calls `mp_ilogb` now and replace it with a note in the documentation (So it would be `MP_DEPRECATED(mp_ilogb)` and **not**  `MP_DEPRECATED(mp_radix_size_overestimate)`).

bn_mp_radix_size_overestimate.c
czurnieden: > I just asked myself if 0 is really the good value to return on error

The returns on error are `MP_VAL`s?
Yes, we could set `size` to `-1` but as we want to change the type of all size-related variables to `size_t` later we shouldn't introduce things that **we know** get changed again later.

And also: if the user skips the error-check and use the resulting `size` for `malloc` directly and if that value is `-1` and it is a 32 bit machine with `sizeof(size_t) == 4` and there is enough RAM and and/or space in the swap-partition/file&hellip;

> set it once on function-entry

Yepp, that's right, will do.

Ahm, now that you mentioned it, would it be a good idea to check if we have something to write on in the first place? I  mean `if (size == NULL) ...`?

bn_mp_radix_size.c
czurnieden: > how about making a private function out of this?

As said above, I would like to deprecate `mp_radix_size` in favor of "`mp_ilogb` plus a note in the documentation".
But I can do it if you insist, no prob'!

bn_mp_radix_size_overestimate.c
minad: No it is not just ilogb. It is also the sign and the 0 byte addition. I would like to keep it. What is the problem with it?

bn_mp_radix_size_overestimate.c
```diff
@@ -0,0 +1,245 @@
...
+      goto LTM_ERR;
+   }
+   /*
+      The first addition of one is done because of the truncating division,
+      and the second addition of one is for NIL ('\0').
+
+      Casting from "long" to "int" can be done because "bi_bit_count" fits into an "int"
+      by definition.
+    */
+   *size = (int)mp_get_l(&bi_bit_count) + 1 + 1 + (a->sign == MP_NEG);
```
minad: Please just write + 3 here

bn_mp_radix_size_overestimate.c
```diff
@@ -0,0 +1,245 @@
...
+      unsigned int x = (unsigned int)radix;
+      int y, rem;
+      for (y=0; (y < 7) && !(x & 1u); y++) {
+         x >>= 1;
+      }
+      *size = bit_count/y;
+      rem = bit_count - ((*size) * y);
+      /* Add 1 for the remainder if any and 1 for "\0". */
+      *size += ((rem == 0) ? 1 : 2) + (a->sign == MP_NEG);
+      return MP_OKAY;
```
minad: Please remove rem here and just write + 3.

bn_mp_radix_size_overestimate.c
minad: This is not about optimization but about simplicity. Why write something more complicated than necessary?

bn_mp_radix_size_overestimate.c
czurnieden: > No it is not just ilogb. It is also the sign and the 0 byte addition. I would like to keep it. 
> What is the problem with it?

The main use case, determining the length of the *string* for memory allocation, has been made redundant by `mp_radix_size_overestimate`.

If you have a case where you need `mp_radix_size` and can not, by any means use `mp_ilogb` instead: tell me.

Otherwise it's around 2.5 Kb that can be spend better elsewhere.

minad: Determining string length is something which I imagine to be useful from time to time. Printing etc. 

Sure the user could also write ilogb+1+sign. But they would also have to check for 0? 

If someone cares about the kbs they will configure the lib to their needs, so nothing significant is being spent.

bn_mp_radix_size_overestimate.c
czurnieden: > Please remove rem here and just write + 3.

That'll lengthen the error-bar to `-0, +3`.
No,  `-0, +1` is good and also expected.

But as I have more time now after the summer rush I will try my teeth at one of your suggestions and reduce the accuracy of the tables (or do something completely different, I have some ideas) to make it smaller and maybe even simpler for the cost of the size of `x` in `-0, +x`. There it will be where I can add your `3`.

minad: I don't see why you want this kind of precision? I don't expect anything from such a function except that it is overestimating and maybe that the error is a constant offset or a constant factors of the real size. This function is for allocation and if the function can be made simpler without being made worse we should do it. If you want to write another version which is more compact, I would appreciate that. But then the question is if we can then replace this function by the new function. In particular if you are specifing exact error bounds.

bn_mp_radix_size_overestimate.c
czurnieden: > Determining string length is something which I imagine to be useful from time to time. Printing etc.

But that's *exactly* what `mp_radix_size` does and is meant for and described to do in the documentation.

> Sure the user could also write ilogb+1+sign.

Yes.

> But they would also have to check for 0?

Yes. And?

Does it happen that often that you need the exact value that it pays for a dedicated function?

minad: I say we keep it. @sjaeckel what do you think?

bn_mp_radix_size_overestimate.c
czurnieden: > I don't see why you want this kind of precision?

I have it and was able to reach it without much ado.
Relaxing accuracy to save a dozen or so CPU cycles?
Reducing code complexity by a very small fraction (There is no "black art" in it)?

> But then the question is if we can then replace this function by the new function. In particular if you are specifing exact error bounds.

I don't know the error-bounds of the new function yet, too many variables. It will be most likely `-0, +6`  (including sign and remainder), although I aim for less

bn_mp_radix_size_overestimate.c
sjaeckel: > And also: if the user skips the error-check 

that was the origin of my thought...

> and use the resulting `size` for `malloc` directly and if that value is `-1` and it is a 32 bit machine with `sizeof(size_t) == 4` and there is enough RAM and and/or space in the swap-partition/file…

:laughing: 

> Ahm, now that you mentioned it, would it be a good idea to check if we have something to write on in the first place? I mean `if (size == NULL) ...`?

puh, that's nearly nowhere checked... I think yes, we should do it, but maybe we should first discuss it?

bn_mp_radix_size.c
sjaeckel: okay, then we can also do the deprecation "now"!?

sjaeckel: let's discuss this above, there's more context already.

bn_mp_radix_size_overestimate.c
sjaeckel: > Determining string length is something which I imagine to be useful from time to time.

I ask myself why someone would want to use `mp_radix_size` after there's `mp_radix_size_overestimate`?

bn_mp_radix_size_overestimate.c
minad: Yes it is cheap but it is useless. The precision is not needed at all. +-1 is as good as +-10 or some other small constant.

sjaeckel: > +-1 is as good as +-10 or some other small constant.

:+1:

bn_mp_radix_size_overestimate.c
czurnieden: > but maybe we should first discuss it?

Hu? Oh, left it in after testing, sorry.

bn_mp_radix_size_overestimate.c
czurnieden: If both of you are OK with such large error I can try something completely different here. Please give me a day or two for it.

minad: Cool! What about splitting your PR? One for the improvements to mp_radix_size by using ilogb and one for overestimate? The first one can be merged as is!

czurnieden:
Tried several variations but decide for the simplest one, the "minimalistic version".
Absolute error is quite high for *very* large input but OK for normal use.

100% tested.

Excerpt from the logs:
```text
First occurence of an error equal to 10

base = radix
bits = floor(log_2(a))
largest error at a = 2^(2^31)

base      bits   floor(log_2(bits))      largest error
 2          n/a      n/a                         3
 3     11250067       23                      1339
 4          n/a      n/a                         3
 5      1315969       20                     11427
 6     29921688       24                       505
 7       531174       19                     28327
 8          n/a      n/a                         3
 9     22499649       24                       671
10      2694087       21                      5584
11      1033948       19                     14543
12     57550267       25                       264
13    361691491       28                        44      
14       976518       19                     15402
15      3533931       21                      4258
16          n/a      n/a                         3
17      1933701       20                      7779
18     38933860       25                       389
19     46634755       25                       325
20      4559686       22                      3300
21      1280545       20                     11747
22      1718219       20                      8753
23     60041844       25                       253
24     94134750       26                       162
25      2631938       21                      5715
26    583596936       29                        28
27     33749716       25                       448
28      1556410       20                      9662
29      2332545       21                      6449
30      5573791       22                      2700
31      1813652       20                      8293
32          n/a      n/a                         3
33      2157447       21                      6973
34      2995666       21                      5022
35     17448836       24                       864
36     59843376       25                       254
37      1849059       20                      8136
38     71174130       26                       214
39    106965320       26                       143
40      6914084       22                      2177
41     24933672       24                       605
42      1929630       20                      7795
43      1757578       20                      8557
44      2575132       21                      5841
45      6637893       22                      2267
46     89524677       26                       170
47      9221534       23                      1633
48    139675137       27                       110
49      2572340       21                      5849
 5      1315969       20                     11427
50      3887934       21                      3870
51      3630477       21                      4145
52    858309240       29                        20
53     15129752       23                       996
54     49436377       25                       307
55      2132680       21                      7052
56      2271965       21                      6621
57     55751906       25                       272
58      3391812       21                      4436 
59      3244460       21                      4638
60      8076397       22                      1864
61      3360593       21                      4476
62      2619894       21                      5742
63      2336967       21                      6438
64          n/a      n/a                         3          
```

minad:
I will take a look. Just one thing - could you please split your PRs.

1. mp_to_radix improvements
2. mp_to_radix_overestimate (old version for reference)
3. mp_to_radix_overestimate (new version with larger error)

minad:
Concerning the error table - the error looks indeed quite large. Do you know the error bounds? It would also be helpful if your table would show base, bits, exact size, overestimated size, relative and absolute errors?

czurnieden:
> Just one thing - could you please split your PRs.
> mp_to_radix improvements

Now that you mentioned it: I think those improvements (the power-of-two shortcuts) would be better placed into `mp_ilogb`?

> mp_to_radix_overestimate (old version for reference)

But you didn't want it?

> mp_to_radix_overestimate (new version with larger error)

Please take a look first, I'm already juggling way too many PRs :-) 

> I'd like to see what @czurnieden means by "very large" :-D

Size is in the eye of the beholder ;-)

I tarred&feathered the logs and put it into my [Google Drive](https://drive.google.com/open?id=185eNgwH3Op7O6QHhT4Y-EKQGs8bSXw2q)

Aaaand forgot to include the actual testing rig, yeah, typical.
But it's small, so here it is:

```c
#include "tommath_private.h"

/*
   Table of {0, log_2([1..64])} times 2^p where p is the scale
   factor defined in LTM_RADIX_SIZE_SCALE.
 */
/* *INDENT-OFF* */
#define LTM_RADIX_SIZE_SCALE 13
static const uint16_t logbases[65] = {
       0u,     0u,  8192u, 12984u, 16384u,
   19021u, 21176u, 22997u, 24576u, 25968u,
   27213u, 28339u, 29368u, 30314u, 31189u,
   32005u, 32768u, 33484u, 34160u, 34799u,
   35405u, 35981u, 36531u, 37057u, 37560u,
   38042u, 38506u, 38952u, 39381u, 39796u,
   40197u, 40584u, 40960u, 41323u, 41676u,
   42019u, 42352u, 42675u, 42991u, 43298u,
   43597u, 43889u, 44173u, 44451u, 44723u,
   44989u, 45249u, 45503u, 45752u, 45995u,
   46234u, 46468u, 46698u, 46923u, 47144u,
   47360u, 47573u, 47783u, 47988u, 48190u,
   48389u, 48584u, 48776u, 48965u, 49152u
};
/* *INDENT-ON* */

static mp_err mp_radix_sizeinbase2(const mp_int *a, const int bits, const int base, int *size)
{
   mp_int bi_bit_count, bi_k;
   int bit_count;
   mp_err err = MP_OKAY;

   *size = 0;

   if ((base < 2) || (base > 64)) {
      return MP_VAL;
   }
   if (a == NULL) {
      bit_count = bits;
   } else {
     bit_count = mp_count_bits(a) + 1;
   }

   if (bit_count == 0) {
      *size = 2;
      return MP_OKAY;
   }

   if ((err = mp_init_multi(&bi_bit_count, &bi_k, NULL)) != MP_OKAY) {
      return err;
   }

   mp_set_l(&bi_bit_count, bit_count);
   mp_set_u32(&bi_k, logbases[base]);

   if ((err = mp_mul_2d(&bi_bit_count, LTM_RADIX_SIZE_SCALE, &bi_bit_count)) != MP_OKAY) {
      goto LTM_E1;
   }
   if ((err = mp_div(&bi_bit_count, &bi_k, &bi_bit_count, NULL)) != MP_OKAY) {
      goto LTM_E1;
   }

   *size = (int)mp_get_l(&bi_bit_count) + 3;
#if ( (defined MP_8BIT) && (INT_MAX > 0xFFFF))
   *size += 2;
#endif

LTM_E1:
   mp_clear_multi(&bi_bit_count, &bi_k, NULL);
   return err;
}


#include <stdio.h>
#include <stdlib.h>
#include <limits.h>
#include <math.h>

#ifndef M_LN2
#define M_LN2 0.69314718055994530942
#endif
int main(int argc, char **argv)
{

   int integer_size, double_size, base;
   double log2_base;
   int i;
   int is;
   int diff_max = 0;
   int max_int = INT_MAX;

   if (argc < 2) {
      printf("usage: %s base\n",argv[0]);
      exit(EXIT_FAILURE);
   }

   base = atoi(argv[1]);
   log2_base = log(base)/M_LN2;
#ifdef PRINT_DIFFS
   printf("# bits    base  max diff   ds    is\n");
#endif
   for (i = 1; i < max_int; i++) {
      double_size = (int)lrint(ceil((double)i / log2_base));
      mp_radix_sizeinbase2(NULL,i,base,&integer_size);
      if (integer_size < double_size) {
         fprintf(stderr,"FAILURE at i = %d, should be at least %d but is %d\n",i,double_size, integer_size);
         exit(EXIT_FAILURE);
      }
      is = integer_size - double_size;
#ifdef PRINT_DIFFS
      if(is > diff_max) {
         diff_max = is;
         printf("%d %d %d %d %d\n",i,base,diff_max,double_size, integer_size);
      }
#endif
#ifdef PRINT_PROGRESS
      if (i % 100000000 == 0) {
         fprintf(stderr,"%d: ds  = %d is =  %d\n",i,double_size, integer_size);
      }
#endif
   }
   exit(EXIT_SUCCESS);
}
```

As you can see: I used floating point math to get a value to compare against. That additional error does not influence the result for the max. input of `2^31`.

czurnieden:
> Just one thing - could you please split your PRs.

- Full version of mp_radix_size_overestimate is in #369 
- Minimal version of bn_mp_radix_size_overestimate is in #368
- Use of mp_ilogb in mp_radix_size #366

I put the the shortcuts for bases that are powers of two in `mp_ilogb` and it is in#367 so it might go together with #366 if it (#366) is still wanted.

minad:
@czurnieden @sjaeckel I guess it is okay to close this PR in favor of the other ones? From my side I would prefer if we could finish 1.2 now instead of delaying it further. The discussion regarding radix_size_overestimate is not yet concluded, so I would prefer if we move that too to 2.0. Then we can also decide if we add a specialzed to_radix and a specialied radix_size_overestimate only for base 10. Would that be ok?

sjaeckel:
> I guess it is okay to close this PR in favor of the other ones? From my side I would prefer if we could finish 1.2 now instead of delaying it further. The discussion regarding radix_size_overestimate is not yet concluded, so I would prefer if we move that too to 2.0. Then we can also decide if we add a specialzed to_radix and a specialied radix_size_overestimate only for base 10. Would that be ok?

OK

PR #348, opened 2019-09-08
make mp_to_radix return the count of characters of the converted number

czurnieden:
As suggested by @sjaeckel in #343 

Mmh&hellip;

### DIFF
--- a/bn_deprecated.c
+++ b/bn_deprecated.c
@@ -113,7 +113,10 @@ mp_err mp_toom_sqr(const mp_int *a, mp_int *b)
 #ifdef S_MP_REVERSE_C
 void bn_reverse(unsigned char *s, int len)
 {
-   s_mp_reverse(s, len);
+   if (len < 0) {
+      return MP_VAL;
+   }
+   s_mp_reverse(s, (size_t)len);
 }
 #endif
 #ifdef BN_MP_TC_AND_C
@@ -304,13 +307,13 @@ mp_err mp_toradix_n(const mp_int *a, char *str, int radix, int maxlen)
    if (maxlen < 0) {
       return MP_VAL;
    }
-   return mp_to_radix(a, str, (size_t)maxlen, radix);
+   return mp_to_radix(a, str, (size_t)maxlen, NULL, radix);
 }
 #endif
 #ifdef BN_MP_TORADIX_C
 mp_err mp_toradix(const mp_int *a, char *str, int radix)
 {
-   return mp_to_radix(a, str, SIZE_MAX, radix);
+   return mp_to_radix(a, str, SIZE_MAX, NULL, radix);
 }
 #endif
 #endif
--- a/bn_mp_fwrite.c
+++ b/bn_mp_fwrite.c
@@ -9,9 +9,15 @@ mp_err mp_fwrite(const mp_int *a, int radix, FILE *stream)
    char *buf;
    mp_err err;
    int len;
+   size_t written;
 
-   if ((err = mp_radix_size(a, radix, &len)) != MP_OKAY) {
-      return err;
+   /* TODO: this function is not in this PR */
+   if (MP_HAS(MP_RADIX_SIZE_OVERESTIMATE)) {
+      /* if ((err = mp_radix_size_overestimate(&t, base, &len)) != MP_OKAY)      goto LBL_ERR; */
+   } else {
+      if ((err = mp_radix_size(a, radix, &len)) != MP_OKAY) {
+         return err;
+      }
    }
 
    buf = (char *) MP_MALLOC((size_t)len);
@@ -19,15 +25,16 @@ mp_err mp_fwrite(const mp_int *a, int radix, FILE *stream)
       return MP_MEM;
    }
 
-   if ((err = mp_to_radix(a, buf, (size_t)len, radix)) != MP_OKAY) {
+   if ((err = mp_to_radix(a, buf, (size_t)len, &written, radix)) != MP_OKAY) {
       goto LBL_ERR;
    }
 
-   if (fwrite(buf, (size_t)len, 1uL, stream) != 1uL) {
+   if (fwrite(buf, written, 1uL, stream) != 1uL) {
       err = MP_ERR;
       goto LBL_ERR;
    }
    err = MP_OKAY;
+
 
 LBL_ERR:
    MP_FREE_BUFFER(buf, (size_t)len);
--- a/bn_mp_radix_size.c
+++ b/bn_mp_radix_size.c
@@ -3,11 +3,11 @@
 /* LibTomMath, multiple-precision integer library -- Tom St Denis */
 /* SPDX-License-Identifier: Unlicense */
 
-/* returns size of ASCII reprensentation */
+/* returns size of ASCII representation */
 mp_err mp_radix_size(const mp_int *a, int radix, int *size)
 {
    mp_err  err;
-   int     digs;
+   int digs;
    mp_int   t;
    mp_digit d;
 
@@ -25,7 +25,7 @@ mp_err mp_radix_size(const mp_int *a, int radix, int *size)
 
    /* special case for binary */
    if (radix == 2) {
-      *size = mp_count_bits(a) + ((a->sign == MP_NEG) ? 1 : 0) + 1;
+      *size = (mp_count_bits(a) + ((a->sign == MP_NEG) ? 1 : 0) + 1);
       return MP_OKAY;
    }
 
--- a/bn_mp_to_radix.c
+++ b/bn_mp_to_radix.c
@@ -5,9 +5,10 @@
 
 /* stores a bignum as a ASCII string in a given radix (2..64)
  *
- * Stores upto maxlen-1 chars and always a NULL byte
+ * Stores upto "size - 1" chars and always a NULL byte, puts the number of characters
+ * written, including the '\0', in "written".
  */
-mp_err mp_to_radix(const mp_int *a, char *str, size_t maxlen, int radix)
+mp_err mp_to_radix(const mp_int *a, char *str, size_t maxlen, size_t *written, int radix)
 {
    size_t  digs;
    mp_err  err;
@@ -15,7 +16,13 @@ mp_err mp_to_radix(const mp_int *a, char *str, size_t maxlen, int radix)
    mp_digit d;
    char   *_s = str;
 
-   /* check range of the maxlen, radix */
+
+   /* If we want to fill a bucket we need a bucket in the first place. */
+   if (str == NULL) {
+      return MP_VAL;
+   }
+
+   /* check range of radix and size*/
    if ((maxlen < 2u) || (radix < 2) || (radix > 64)) {
       return MP_VAL;
    }
@@ -24,6 +31,9 @@ mp_err mp_to_radix(const mp_int *a, char *str, size_t maxlen, int radix)
    if (MP_IS_ZERO(a)) {
       *str++ = '0';
       *str = '\0';
+      if (written != NULL) {
+         *written = 2u;
+      }
       return MP_OKAY;
    }
 
@@ -43,11 +53,12 @@ mp_err mp_to_radix(const mp_int *a, char *str, size_t maxlen, int radix)
       /* subtract a char */
       --maxlen;
    }
-
    digs = 0u;
    while (!MP_IS_ZERO(&t)) {
       if (--maxlen < 1u) {
          /* no more room */
+         /* TODO: It could mimic mp_to_radix_n if that is not an error
+                  or at least not this error (MP_ITER or a new one?). */
          err = MP_VAL;
          break;
       }
@@ -57,7 +68,6 @@ mp_err mp_to_radix(const mp_int *a, char *str, size_t maxlen, int radix)
       *str++ = mp_s_rmap[d];
       ++digs;
    }
-
    /* reverse the digits of the string.  In this case _s points
     * to the first digit [exluding the sign] of the number
     */
@@ -65,6 +75,10 @@ mp_err mp_to_radix(const mp_int *a, char *str, size_t maxlen, int radix)
 
    /* append a NULL so the string is properly terminated */
    *str = '\0';
+   digs++;
+   if (written != NULL) {
+      *written = (a->sign == MP_NEG) ? digs + 1u: digs;
+   }
 
 LBL_ERR:
    mp_clear(&t);
--- a/demo/test.c
+++ b/demo/test.c
@@ -1189,6 +1189,8 @@ static int test_mp_montgomery_reduce(void)
    int ix, i, n;
    char buf[4096];
 
+   /* size_t written; */
+
    mp_int a, b, c, d, e;
    if (mp_init_multi(&a, &b, &c, &d, &e, NULL)!= MP_OKAY) {
       return EXIT_FAILURE;
@@ -1225,6 +1227,7 @@ static int test_mp_montgomery_reduce(void)
                mp_to_decimal(&e, buf, sizeof(buf)); printf("e = %s\n", buf);
                mp_to_decimal(&d, buf, sizeof(buf)); printf("d = %s\n", buf);
                mp_to_decimal(&c, buf, sizeof(buf)); printf("c = %s\n", buf);
+
                printf("compare no compare!\n"); goto LBL_ERR;
 /* *INDENT-ON* */
             }
@@ -1250,31 +1253,67 @@ LBL_ERR:
 static int test_mp_read_radix(void)
 {
    char buf[4096];
+   size_t written;
+   mp_err err;
 
    mp_int a;
-   if (mp_init_multi(&a, NULL)!= MP_OKAY) {
-      return EXIT_FAILURE;
-   }
+   if (mp_init_multi(&a, NULL)!= MP_OKAY)                                       goto LTM_ERR;
 
-   mp_read_radix(&a, "123456", 10);
-   mp_to_radix(&a, buf, 3, 10);
+   if ((err = mp_read_radix(&a, "123456", 10)) != MP_OKAY)                     goto LTM_ERR;
+   /* Must fail */
+   if ((err = mp_to_radix(&a, NULL, SIZE_MAX, NULL, 10)) != MP_VAL)             goto LTM_ERR;
+
+   if ((err = mp_to_radix(&a, buf, SIZE_MAX, &written, 10)) != MP_OKAY)        goto LTM_ERR;
+   printf(" '123456' a == %s, length = %zu\n", buf, written);
+
+   /* See comment in bn_mp_to_radix.c */
+   /*
+      if( (err = mp_to_radix(&a, buf, 3u, &written, 10) ) != MP_OKAY)              goto LTM_ERR;
+      printf(" '56' a == %s, length = %zu\n", buf, written);
+
+      if( (err = mp_to_radix(&a, buf, 4u, &written, 10) ) != MP_OKAY)              goto LTM_ERR;
+      printf(" '456' a == %s, length = %zu\n", buf, written);
+      if( (err = mp_to_radix(&a, buf, 30u, &written, 10) ) != MP_OKAY)             goto LTM_ERR;
+      printf(" '123456' a == %s, length = %zu, error = %s\n",
+             buf, written, mp_error_to_string(err));
+   */
+   if ((err = mp_read_radix(&a, "-123456", 10)) != MP_OKAY)                    goto LTM_ERR;
+   if ((err = mp_to_radix(&a, buf, SIZE_MAX, &written, 10)) != MP_OKAY)        goto LTM_ERR;
+   printf(" '-123456' a == %s, length = %zu\n", buf, written);
+
+   if ((err = mp_read_radix(&a, "0", 10)) != MP_OKAY)                          goto LTM_ERR;
+   if ((err = mp_to_radix(&a, buf, SIZE_MAX, &written, 10)) != MP_OKAY)        goto LTM_ERR;
+   printf(" '0' a == %s, length = %zu\n", buf, written);
+
+
+
+   /* Although deprecated it needs to function as long as it isn't dropped */
+   /*
+   printf("Testing deprecated mp_toradix_n\n");
+   if( (err = mp_read_radix(&a, "-123456", 10) ) != MP_OKAY)                    goto LTM_ERR;
+   if( (err = mp_toradix_n(&a, buf, 10, 3) ) != MP_OKAY)                        goto LTM_ERR;
    printf("a == %s\n", buf);
-   mp_to_radix(&a, buf, 4, 10);
+   if( (err = mp_toradix_n(&a, buf, 10, 4) ) != MP_OKAY)                        goto LTM_ERR;
    printf("a == %s\n", buf);
-   mp_to_radix(&a, buf, 30, 10);
+   if( (err = mp_toradix_n(&a, buf, 10, 30) ) != MP_OKAY)                       goto LTM_ERR;
    printf("a == %s\n", buf);
+   */
+
 
    while (0) {
       char *s = fgets(buf, sizeof(buf), stdin);
       if (s != buf) break;
       mp_read_radix(&a, buf, 10);
       mp_prime_next_prime(&a, 5, 1);
-      mp_to_radix(&a, buf, sizeof(buf), 10);
+      mp_to_radix(&a, buf, sizeof(buf), NULL, 10);
       printf("%s, %lu\n", buf, (unsigned long)a.dp[0] & 3uL);
    }
 
-   mp_clear_multi(&a, NULL);
+   mp_clear(&a);
    return EXIT_SUCCESS;
+LTM_ERR:
+   mp_clear(&a);
+   return EXIT_FAILURE;
 }
 
 static int test_mp_cnt_lsb(void)
@@ -1444,6 +1483,7 @@ static int test_mp_reduce_2k_l(void)
    mp_int a, b, c, d;
    int cnt;
    char buf[4096];
+   size_t length[1];
    if (mp_init_multi(&a, &b, NULL)!= MP_OKAY) {
       return EXIT_FAILURE;
    }
@@ -1463,9 +1503,9 @@ static int test_mp_reduce_2k_l(void)
 #      else
 #         error oops
 #      endif
-
-   mp_to_decimal(&a, buf, sizeof(buf));
-   printf("\n\np==%s\n", buf);
+   *length = sizeof(buf);
+   mp_to_radix(&a, buf, length, 10);
+   printf("\n\np==%s, length = %zu\n", buf, *length);
    /* now mp_reduce_is_2k_l() should return */
    if (mp_reduce_is_2k_l(&a) != 1) {
       printf("mp_reduce_is_2k_l() return 0, should be 1\n");
@@ -1824,6 +1864,7 @@ LTM_ERR:
    All numbers as strings to simplifiy things, especially for the
    low-mp branch.
 */
+
 static int test_mp_root_u32(void)
 {
    mp_int a, c, r;
--- a/tommath.h
+++ b/tommath.h
@@ -718,7 +718,7 @@ mp_err mp_to_sbin(const mp_int *a, unsigned char *buf, size_t maxlen, size_t *wr
 mp_err mp_read_radix(mp_int *a, const char *str, int radix) MP_WUR;
 MP_DEPRECATED(mp_to_radix) mp_err mp_toradix(const mp_int *a, char *str, int radix) MP_WUR;
 MP_DEPRECATED(mp_to_radix) mp_err mp_toradix_n(const mp_int *a, char *str, int radix, int maxlen) MP_WUR;
-mp_err mp_to_radix(const mp_int *a, char *str, size_t maxlen, int radix) MP_WUR;
+mp_err mp_to_radix(const mp_int *a, char *str, size_t maxlen, size_t *written, int radix) MP_WUR;
 mp_err mp_radix_size(const mp_int *a, int radix, int *size) MP_WUR;
 
 #ifndef MP_NO_FILE
@@ -738,10 +738,10 @@ mp_err mp_fwrite(const mp_int *a, int radix, FILE *stream) MP_WUR;
 #define mp_todecimal(M, S) (MP_DEPRECATED_PRAGMA("replaced by mp_to_decimal") mp_toradix((M), (S), 10))
 #define mp_tohex(M, S)     (MP_DEPRECATED_PRAGMA("replaced by mp_to_hex")     mp_toradix((M), (S), 16))
 
-#define mp_to_binary(M, S, N)  mp_to_radix((M), (S), (N), 2)
-#define mp_to_octal(M, S, N)   mp_to_radix((M), (S), (N), 8)
-#define mp_to_decimal(M, S, N) mp_to_radix((M), (S), (N), 10)
-#define mp_to_hex(M, S, N)     mp_to_radix((M), (S), (N), 16)
+#define mp_to_binary(M, S, N)  mp_to_radix((M), (S), (N), NULL, 2)
+#define mp_to_octal(M, S, N)   mp_to_radix((M), (S), (N), NULL, 8)
+#define mp_to_decimal(M, S, N) mp_to_radix((M), (S), (N), NULL, 10)
+#define mp_to_hex(M, S, N)     mp_to_radix((M), (S), (N), NULL, 16)
 
 #ifdef __cplusplus
 }
--- a/demo/shared.c
+++ b/demo/shared.c
@@ -2,7 +2,7 @@
 
 void ndraw(mp_int *a, const char *name)
 {
-   char *buf;
+   char *buf = NULL;
    int size;
 
    mp_radix_size(a, 10, &size);
--- a/doc/bn.tex
+++ b/doc/bn.tex
@@ -100,7 +100,7 @@ also build in MSVC, Borland C out of the box.  For any other ISO C compiler a ma
 developer. Please consider to commit such a makefile to the LibTomMath developers, currently residing at
 \url{http://github.com/libtom/libtommath}, if successfully done so.
 
-Intel's C-compiler (ICC) is sufficiently compatible with GCC, at least the newer versions, to replace GCC for building the static and the shared library. Editing the makfiles is not needed, just set the shell variable \texttt{CC} as shown below.
+Intel's C-compiler (ICC) is sufficiently compatible with GCC, at least the newer versions, to replace GCC for building the static and the shared library. Editing the makefiles is not needed, just set the shell variable \texttt{CC} as shown below.
 \begin{alltt}
 CC=/home/czurnieden/intel/bin/icc make
 \end{alltt}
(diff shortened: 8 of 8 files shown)


czurnieden:
First problem is, as you can see, are the macros but that was expected and is, I think, solvable.
The problms with the MS-compiler are&hellip;interresting.

czurnieden:
I find it a bit awkward that way (there are alternatives which I find equally or even more odd, but YMMV, so drop a note, please).

A cleaner way, I think, would be to add another parameter, maybe
```c
mp_err mp_to_radix(const mp_int *a, char *str, size_t maxlen, size_t *used_length, int radix)
```
or something in that line.

bn_deprecated.c
```diff
@@ -232,16 +232,20 @@ mp_err mp_n_root(const mp_int *a, mp_digit b, mp_int *c)
...
+   return mp_to_radix(a, str, length, radix);
 }
 #endif
 #ifdef BN_MP_TORADIX_C
 mp_err mp_toradix(const mp_int *a, char *str, int radix)
 {
-   return mp_to_radix(a, str, SIZE_MAX, radix);
+   size_t length[1];
+   *length = SIZE_MAX;
+   return mp_to_radix(a, str, length, radix);
```
minad: size_t len = SIZE_MAX; ...&len) please

minad: No awkwardness then

sjaeckel:
> I find it a bit awkward that way

what do you find awkward?

What do you think about replacing `mp_radix_size()` in favor of `mp_to_radix()`?

* `len = 0; mp_to_radix(x, NULL, &len, radix);` where len will then be filled with the approximated-to-be-allocated length... this wouldn't even need a separate error code...
* `len = 4567; mp_to_radix(x, buf, &len, radix);` (where 4567 is enough) and len get's decreased to 4566 because the approximation overshot
* an option would be `len = 123; mp_to_radix(x, buf, &len, radix);` (where 123 is too small to hold x) we would need a separate error code to tell the user to increase the buffer size

minad:
Removing radix_size is a good idea. We could still branch to an O(1) size computation function internally if buf==NULL.

minad:
The function in #343 could be used internally.

minad:
@sjaeckel what shall we do about the to_signed_bin functions? They also need a size argument. These functions should probably be treated differently since size computation is cheaper. But I don't like the difference in API now.

sjaeckel:
> The function in #343 could be used internally.

I'd say either we drop the original one or we keep them both through a compile-time setting if someone doesn't want to spend the space for the tables

> @sjaeckel what shall we do about the to_signed_bin functions? They also need a size argument. These functions should probably be treated differently since size computation is cheaper. But I don't like the difference in API now.

very good point! IMO we should deprecate the ones w/o length
... and probably can we then rename `mp_to_unsigned_bin_n` to `mp_to_unsigned_bin` for 2.0 (and provide a `#define mp_to_unsigned_bin_n mp_to_unsigned_bin`)  ... not sure though ...

minad:
Just introduce functions with new names mp_to_ubin and mp_to_sbin?

czurnieden:
>  >  The function in #343 could be used internally.
> I'd say either we drop the original one or we keep them both through a compile-time setting if someone doesn't want to spend the space for the tables

Tables has about 14k (depending a bit on MP_xBIT but not much), original about 3k , and the one with `mp_ilogb` has about 6k.
Oh, I though it would be more!

> what do you find awkward?

My social interactions.
But making fun of myself aside: 

> What do you think about replacing mp_radix_size() in favor of mp_to_radix()?

That would be quite radical aaaaaand I like it! ;-)
\*rolls up sleeves\*

bn_deprecated.c
czurnieden: I really should use a bit less of that dreaded C&P, or what do you think? ;-)

minad:
Hmm, I worked a bit on mp_to_radix and I am not sure if deprecating mp_radix_size is the right way.

We have the following scenarios:

* str==NULL, *len=returns needed length
* str len too small, *len returns written chars or needed length?
* str len large enough, *len returns written chars

In particular the overloading in the case of buffer overflow is not nice. We could always return the needed length, but then the division would have to continue even if no chars are written anymore.

czurnieden:
> Hmm, I worked a bit on `mp_to_radix` and I am not sure if deprecating `mp_radix_size` is the right way.

The exact details are debatable, of course, but in general it is a good idea to get rid of it.

Since we got `mp_ilogb` the function `mp_radix_size` is reduced to not much more than a convenient little wrapper for `mp_ilogb` and every user who needs an exact size without actually printing the number can do it themselves, a note in the docs is more than sufficient here.

No `to_string` function I know needs to get given the size in advance, to the contrary, most of them return the number of characters needed. They only need a pointer and do the allocating themself, only freeing is left to the user. Used internally is in most cases a kind of log-function that overshoots by a couple and the actual buffer gets cut at the end as needed, if at all.
Although we cannot do it *exactly* in that way, it is not the worst solution, I think.

> In particular the overloading in the case of buffer overflow is not nice.

I understood it in the way that that case is meant to replace `mp_to_radix_n`?
I still think that something like

```c
mp_err mp_to_radix(const mp_int *a, char *str, size_t *length_written, size_t size, int radix) 
```
with

 -  `str == NULL` -> FAIL because you cannot free that buffer after use anymore
 - `length_written == NULL` -> OK (if the caller doesn't want/need to know, who are we to insist?)
 - `size == 0` -> OK, do the "computing size and allocating buffer" internally
 - `size != 0` -> OK, do the "allocating buffer" internally and convert only the first `size` digits output (like the old `mp_toradix_n`).

would be better.
It would also be quite easy to the the old shortcut macros like `mp_todecimal`, `mp_tohex` etc.

@sjaeckel 
> But I don't like the difference in API now.

Could you be a bit more specific, please?

minad:
@czurnieden I think a function which allocates would also be acceptable, similar to what gmp offers.

minad:
~~~
size_t mp_ubin_size(const mp_int *a) MP_WUR;
mp_err mp_from_ubin(mp_int *a, const unsigned char *buf, size_t size) MP_WUR;
mp_err mp_to_ubin(const mp_int *a, unsigned char *buf, size_t size, size_t* written) MP_WUR;

size_t mp_sbin_size(const mp_int *a) MP_WUR;
mp_err mp_from_sbin(mp_int *a, const unsigned char *buf, size_t size) MP_WUR;
mp_err mp_to_sbin(const mp_int *a, unsigned char *buf, size_t size, size_t* written) MP_WUR;

mp_err mp_radix_size(const mp_int *a, size_t *size, int radix) MP_WUR;
mp_err mp_from_radix(mp_int *a, const char *str, int radix) MP_WUR;
mp_err mp_to_radix(const mp_int *a, char *str, size_t size, size_t* written, int radix) MP_WUR;
~~~

As @czurnieden proposed above The `mp_to_radix`, `mp_to_sbin` and `mp_to_ubin` functions will allocate when `str == NULL`. written bytes will be returned if `written!=NULL`.

czurnieden:
A quick try at the propositions. To be debated.
(`*bin*` not ignored, just not touched yet)

czurnieden:
Took the liberty to add "work in progress" because that's what it is, isn't it?

bn_mp_to_radix.c
```diff
@@ -5,66 +5,147 @@
...
+               return MP_MEM;
+            }
+         }
+      }
+      **str = '0';
+      (*str)++;
+      **str = '\0';
+      if (written != NULL) {
+         *written = 2u;
+      }
```
minad: Looks overly complex. No realloc since the user could pass static array. size==0&&*str!=0 and vice versa should be an error.

sjaeckel:
I'm skeptic about the pattern-change from "alloc outside" to "alloc inside", but I somehow like it. I'd say if we do it like that we won't need the `written` arg anyways, right?

```c
size_t mp_ubin_size(const mp_int *a) MP_WUR;
mp_err mp_from_ubin(mp_int *a, const unsigned char *buf, size_t size) MP_WUR;
mp_err mp_to_ubin(const mp_int *a, unsigned char **buf, size_t *len) MP_WUR;

size_t mp_sbin_size(const mp_int *a) MP_WUR;
mp_err mp_from_sbin(mp_int *a, const unsigned char *buf, size_t size) MP_WUR;
mp_err mp_to_sbin(const mp_int *a, unsigned char **buf, size_t *len) MP_WUR;

mp_err mp_radix_size(const mp_int *a, size_t *size, unsigned radix) MP_WUR;
mp_err mp_from_radix(mp_int *a, const char *str, unsigned radix) MP_WUR;
mp_err mp_to_radix(const mp_int *a, char **str, size_t *len, unsigned radix) MP_WUR;
```

where 
1. `*len` is always ignored as input\
2. we use the internal size-approximation 
3.a to do either the `malloc` if `*str` is NULL 
3.b or `realloc` to that size.

minad:
No realloc! This is a bad idea!

sjaeckel:
> No realloc! This is a bad idea!

then no allocation inside IMO

sjaeckel:
> > No realloc! This is a bad idea!
> 
> then no allocation inside IMO

we can also do a `free` + `malloc` as we don't need the copy

bn_mp_to_radix.c
sjaeckel: > No realloc since the user could pass static array. size==0&&*str!=0 and vice versa should be an error.

I'd say either we do the alloc or the user does the alloc, I'm no fan of having both as an option through the same API

minad:
>  then no allocation inside IMO 

Imo we can provide a function which allocates if no buffer is supplied. Only the idea of reallocating a given buffer is a bad idea because of static arrays passed in. This will only lead to user errors. I will update my proposal.

minad:
~~~ c
size_t mp_ubin_size(const mp_int *a) MP_WUR;
mp_err mp_from_ubin(mp_int *a, const unsigned char *buf, size_t size) MP_WUR;
mp_err mp_to_ubin(const mp_int *a, unsigned char **buf, size_t *maxsize) MP_WUR;

size_t mp_sbin_size(const mp_int *a) MP_WUR;
mp_err mp_from_sbin(mp_int *a, const unsigned char *buf, size_t size) MP_WUR;
mp_err mp_to_sbin(const mp_int *a, unsigned char **buf, size_t *maxsize) MP_WUR;

mp_err mp_radix_size(const mp_int *a, size_t *size, unsigned radix) MP_WUR;
mp_err mp_radix_size_overestimate(const mp_int *a, size_t *size, unsigned radix) MP_WUR;
mp_err mp_from_radix(mp_int *a, const char **str, unsigned radix) MP_WUR;
mp_err mp_to_radix(const mp_int *a, char **str, size_t *maxsize, unsigned radix) MP_WUR;
~~~

* There will be mp_radix_size (exact) and mp_radix_size_overestimate (fast and useful for allocations, basically a wrapper around log).
* For the mp_to_x functions behave as follows:

    * `str==NULL` is an error (no destination given)
    * `str!=NULL`:
        * `maxsize==NULL`:
             *  `*str==NULL`: function allocates buffer and writes there, user has to free buffer
             * `*str!=NULL` is an error (destination buffer given, but no size)
        * `maxsize!=NULL`:
             * `*str==NULL`:
                * `*maxsize==0`: function allocates and writes written bytes to *maxsize
                * `*maxsize>0` is an error (size given, but no destination buffer)
             * `*str!=NULL`:
                * `*maxsize==0` is an error (destination buffer given, but no size)
                * `*maxsize>0`: function writes to user supplied buffer and writes written bytes to *maxsize

This means the only allowed cases are:

~~~ c
str!=NULL && *str==NULL && maxsize==NULL
str!=NULL && *str==NULL && maxsize!=NULL && *maxsize==0
str!=NULL && *str!=NULL && maxsize!=NULL && *maxsize>0
~~~

sjaeckel:
Yeah, I can see that working and I can see me using it, but I'm not sure if it fits in the current way our API's are designed... it's suddenly a paradigm-shift...

minad:
Well for me it is also ok to keep the API as simple as possible (no allocations, no returning of written bytes). This is what we have now in develop, except for the missing maxsize argument which should be there for safety.

sjaeckel:
> no returning of written bytes

we should definitely have that, as we will have now a `mp_radix_size_overestimate()`

> except for the missing maxsize argument which should be there for safety

and we should have that as well

czurnieden:
@minad 

> [using one parameter for in- and output and as a flag, too]

The `size` given might be a constant. I don't think that disallowing something like
```c
mp_to_radix(&a, &buf, 5u, 10);
```
would be very user-friendly.
What are the disadvantages of having two parameters for in- and output instead of one?

> no returning of written bytes

That information is too useful to be dropped that easily. See `mp_fwrite` for example which would need some kind of `strlen` without that information or would need to run `mp_radix_size` again. (assuming internal allocation)

> except for the missing maxsize argument which should be there for safety.

We don't need it for safety if we do all of the measurements and allocations internally.
It is not useful for safety if it is user supplied.

> Only the idea of reallocating a given buffer is a bad idea because of static arrays passed in.

Ability to pass static arrays would make internal allocation a bit more difficult and, as a side effect, would make the whole function less safe than the "whole allocation managed internally" because you relay on user input.
Yes, I'm aware of the runtime bounds-checking modern compilers offer but it has an overhead and older compilers don't have it.

@sjaeckel 
> mp_to_radix(const mp_int *a, char **str, size_t *len, unsigned radix)
> `*len` is always ignored as input

That would drop the `mp_to_radix_n` functionality.
But it would also simplify a lot.

> we can also do a free + malloc as we don't need the copy

Yepp, that was the alternative I mentioned. Not very elegant but it works.

I think it all boils down to:
 - everything regarding the buffer done internally, no outside memory accepted: safe and easy to use but changes the way LTM functions normally work.
 - some outside buffer allowed: less safe for the part that relays on user supplied data. Allows for static arrays but also changes the way LTM functions normally work.
 - just the combination of `mp_to_radix + mp_to_radix_n` into one function with the single addition of a way to get the number of written characters which became necessary because of `mp_radix_sizeinbase`. Buffer must be allocated outside. No change in expectations but also completely unsafe because everything relays on user input.

I'm all for the safest variation, if that wasn't already obvious, but would also accept some exceptions if @minad can make a good case for the need of static arrays.

minad:
Allocating internally is also ok for me. However this is also very different from what we have now.

sjaeckel:
I realized only now that those three sets of functions are the only ones dealing with the reading or writing of mp_int's and they are the only ones allowing to write to a user-supplied buffer!
I really like the flexibility of the proposal of @minad, but I'd prefer to stick with the safe approach and only do allocations internally (even though it's very different from what we have now).

The background on why is: all ltm functionality can and will do allocations if required, so why shouldn't we do it here as well?

This should leave us with the same API of what @minad worked out already with the following diff:

* a small fix in `mp_err mp_from_radix(mp_int *a, const char *str, unsigned radix) MP_WUR;` where `*str` was a `**str`.
* `mp_radix_size_overestimate()` could (maybe should) become a private function.

And the modified behavior

* For the mp_to_x functions behave as follows:

    * `str==NULL` is an error (no destination given)
    * `str!=NULL`:
        * function allocates overestimated buffer and writes there, user has to free buffer
        * `maxsize!=NULL`:
             * function writes written bytes to *maxsize

This means the only allowed cases are:

~~~ c
str!=NULL
~~~

PS: after writing this and checking the sources again I realized that `mp_{ex,im}port()` exist as well ... and `mp_export()` doesn't even ... well have a look yourself ... maybe the behavior of `mp_export()` should be changed as well?

czurnieden:
> where `*str` was a `**str`
I think we should give @minad the benefit of doubt here and assume this to be a C&P induced typo.

> `mp_radix_size_overestimate()` could (maybe should) become a private function.
Are all OK with that name? Because in that case I'll rename it.

I [searched Github](https://github.com/search?q=in%3Afile+mpz_sizeinbase+NOT+%22GNU+MP+Library%22&type=Code) with about 44k hits. A lot of those are probably from included libraries that use that function but I think it could be taken as a sign that an approximation with such a small and well defined error (-0/+1 for GMP and LTM after dropping 8-bit support) has it's uses outside some fringe cases.

I'd rather make the exact `mp_radix_size` a private function or even drop it completely and replace it with a note in the documentation to use `mp_ilogb` instead. (and use `mp_ilogb`/`s_mp_radix_size_exact` in the `mp_to_radix` internals in case of memory scarcity).

> function writes written bytes to *maxsize

I'm still not that big of a fan of overusing a single parameter, in this case for both in- and output and as a flag, too.
What exactly makes you uncomfortable with something like
```c
mp_err mp_to_radix(const mp_int *a, char **str,  unsigned radix, size_t maxsize, size_t *written)
```
?
(order of parameters debatable, of course, but the first four are in the same order as in the old `mp_toradix_n`)

With unlimited size by setting `maxsize` to either `SIZE_MAX` (the one from `stdint.h`) or `0u` which is also common as an indicator for "limitless" (yes, I know what I said about "overusing a single parameter" above ;-) ). That also shows a disadvantage of `size_t *maxsize`: you cannot easily use a constant.

It would also be better for 3rd-party code using `mp_toradix_n`. It would be a shell one-liner to update these calls to the new version (only the line with the call itself needs change) instead of something a bit more complicated and hence harder to automate.

Oh, before there are any misunderstandings: I'll follow the majority (if technically possible but that shouldn't a problem), so if you all come to a different but common definition, I'll implement that, no worry!

> `mp_export()` should be changed as well?

I once took a quick look and immediately retracted slowly, trying to make as little noise as possible ;-)
But you are right, of course, it needs an overhaul, too.

BTW: it also uses two different parameters to pass the values of "size given" and "length written".

czurnieden:
Waited WE if somebody complains, nobody did, so here is the first attempt, such that you can complain now.
(`*_bin*` methods are in a different PR they seem to have problems different from this)

czurnieden:
For why "for now" see comment in `bn_mp_fwrite.c` regarding `bn_mp_radix_overestimate`.

bn_deprecated.c
```diff
@@ -230,18 +230,85 @@ mp_err mp_n_root(const mp_int *a, mp_digit b, mp_int *c)
...
+   mp_digit d;
+   char   *_s = str;
+
+   /* check range of the maxlen, radix */
+   if ((maxlen < 2) || (radix < 2) || (radix > 64)) {
       return MP_VAL;
    }
-   return mp_to_radix(a, str, (size_t)maxlen, radix);
+
+   /* quick out if its zero */
```
sjaeckel: similar to `mp_toradix()`

```
char *_s;
size_t l;
mp_err err = mp_to_radix(a, &_s, &l, radix);
if (l > maxlen) l = maxlen;
memcpy(str, _s, l);
free(_s);
```

bn_mp_fwrite.c
```diff
@@ -8,29 +8,19 @@ mp_err mp_fwrite(const mp_int *a, int radix, FILE *stream)
...
-   if (fwrite(buf, (size_t)len, 1uL, stream) != 1uL) {
+   if (fwrite(buf, length, 1uL, stream) != 1uL) {
       err = MP_ERR;
       goto LBL_ERR;
    }
-   err = MP_OKAY;
 
 LBL_ERR:
-   MP_FREE_BUFFER(buf, (size_t)len);
+   /* "length" can be smaller than the allocated buffer! */
```
sjaeckel: is that a problem @minad ?

as of https://github.com/libtom/libtommath/pull/192#issue-267356213 it should be a problem.

bn_mp_to_radix.c
```diff
@@ -5,66 +5,118 @@
...
+         return MP_MEM;
+      }
+      _s = *str;
+      **str = '0';
+      (*str)++;
+      **str = '\0';
+      if (length != NULL) {
+         *length = 2u;
+      }
+      *str = _s;
```
sjaeckel: just work with `_s` above and you don't need to fixup `str` here

bn_mp_radix_size.c
```diff
@@ -4,33 +4,33 @@
 /* SPDX-License-Identifier: Unlicense */
 
 /* returns size of ASCII reprensentation */
-mp_err mp_radix_size(const mp_int *a, int radix, int *size)
+mp_err mp_radix_size(const mp_int *a, int radix, size_t *size)
```
sjaeckel: we currently can't change the signature of this function

demo/main.c
```diff
@@ -5,20 +5,13 @@ int unit_tests(int argc, char **argv);
...
-   }
+   char *buf = NULL;
+   size_t written;
 
    printf("%s: ", name);
-   mp_to_decimal(a, buf, (size_t) size);
+   mp_to_radix(a, &buf, &written, 10);
    printf("%s\n", buf);
-   mp_to_hex(a, buf, (size_t) size);
+   mp_to_radix(a, &buf, &written,  16);
```
sjaeckel: this needs a `free()` before `buf` can be passed again

demo/test.c
```diff
@@ -1221,10 +1222,11 @@ static int test_mp_montgomery_reduce(void)
             if (mp_cmp(&c, &d) != MP_EQ) {
 /* *INDENT-OFF* */
                printf("d = e mod a, c = e MOD a\n");
-               mp_to_decimal(&a, buf, sizeof(buf)); printf("a = %s\n", buf);
-               mp_to_decimal(&e, buf, sizeof(buf)); printf("e = %s\n", buf);
-               mp_to_decimal(&d, buf, sizeof(buf)); printf("d = %s\n", buf);
-               mp_to_decimal(&c, buf, sizeof(buf)); printf("c = %s\n", buf);
+               mp_to_decimal(&a, &buf); printf("a = %s\n", buf);
```
sjaeckel: all of those need to `free(buf)` before calling `mp_to_decimal()` again

bn_s_mp_reverse.c
```diff
@@ -4,13 +4,13 @@
 /* SPDX-License-Identifier: Unlicense */
 
 /* reverse an array, used for radix code */
-void s_mp_reverse(unsigned char *s, int len)
+void s_mp_reverse(unsigned char *s, size_t len)
```
sjaeckel: this signature also can't be changed

bn_deprecated.c
```diff
@@ -230,18 +230,85 @@ mp_err mp_n_root(const mp_int *a, mp_digit b, mp_int *c)
...
+   p = s;
+   while (*p != '\0') {
+      p++;
+   }
+   return (int)(p - s);
+}
 mp_err mp_toradix(const mp_int *a, char *str, int radix)
 {
-   return mp_to_radix(a, str, SIZE_MAX, radix);
+   MP_FREE_BUFFER(str, (size_t)(s_mp_strlen(str) + 1));
```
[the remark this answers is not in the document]
czurnieden: > What if it's located on the stack?

It will crash and burn and die with a whimper as it is fitting for a deprecated function! ;-)

Actually: `helper.pl` complained so I made it a Q&vD at that time (which didn't hinder `helper.pl` from complaining again ;-) ); should have put a `TODO` comment there at least, sorry.

Make an exemption for `bn_deprectated.c` or use an internal `memcpy`?
Something cheap and small like e.g.:

```c
static void *s_memcpy (void *dst, const void *src, size_t len)
{
   char *d = (char *) dst;
   char *s = (char *) src;

   void *start = dst;

   while (len--)  {
       *d = *s;
       d++; s++;
   }
   return start;
}
```

or similar would suffice, so it isn't much work and doesn't use much memory.
There are also the `*_bin*` functions which will have similar (and other) problems in `bn_deprectated.c`, so it won't be a single-use function.
Yeah, I think I'll go with an internal function if nobody complains.

bn_deprecated.c
czurnieden: See above.

bn_mp_fwrite.c
czurnieden: Yes, that can be a showstopper.

That's one of the two reasons I waited the whole WE instead of the obligatory 24h but @minad seeems to enjoy his hiking tour a bit longer. (Yes, I admit, I am a bit envious ;-) )

The other one is that it is not possible anymore to use your own chunk of memory. I think @minad dropped a hint or two regarding static arrays, IIRC?

bn_mp_to_radix.c
czurnieden: Ah, my beloved leftovers! ;-)
Thanks!

No, I don't know what is was that left this, too many variations over the length of this thread.
I know, I really should stop trying to keep up with every single move \*sigh\*

bn_mp_radix_size.c
czurnieden: You wanted and said so?
But you're right, you didn't say when (Or did I miss it?) and it was in another context (`mp_count_bits` I think), so, yepp, my fault, will change it back.
And remove the superfluous "n", too, while I'm there ;-)

Are you willing to set a milestone for it?
LTM 2.0.0 maybe?

demo/main.c
czurnieden: What?
Argh, great, forgot to clean up `demo/*c`.
Sorry!

Mmh, `valgrind` can't detect that? Ah, it isn't used in the `valgrind` runs, that's why it slipped through and made that gaping hole in my foot.
Ouch!

Will check the docs if I can use a big hammer there instead of the little note it has now.
Won't help those, who don't read the documentation&mdash;the majority, I'm afraid&mdash;but will COAs.

demo/test.c
czurnieden: See above.

bn_s_mp_reverse.c
czurnieden: See above.

bn_mp_fwrite.c
minad: Sorry I don't have much time right now. But from what I see this looks very much like a problem :)

minad:
Could someone tell me what the plan is now here? Are we going with allocations internally as suggested by @sjaeckel? If things turn out the be too complicated, too intrusive, require too many changes through the code bases, I think we could still keep the option to go with the solution of just adding "size_t size, size_t* written" parameter to the conversion functions to make them safer. But maybe @sjaeckel already made his decision?

czurnieden:
Hope I didn't forgot any and introduced not too many new bugs.

sjaeckel:
> Could someone tell me what the plan is now here? Are we going with allocations internally as suggested by @sjaeckel?

I would've liked to go with internal allocations, but see below...

> But maybe @sjaeckel already made his decision?

I made the decision to let it evolve a bit a see what happens :-) 

>  If things turn out the be too complicated, too intrusive, require too many changes through the code bases, I think we could still keep the option to go with the solution of just adding "size_t size, size_t* written" parameter to the conversion functions to make them safer. 

IIUC `size_t size, size_t* written` won't help us, it would have to be `size_t *allocated, size_t *written` ...

I'm not so sure if that's an API we want...

```
mp_err mp_to_radix(const mp_int *a, char **str, size_t *allocated, size_t *written, int radix);
```

or do we?

maybe it'd make more sense to over-allocate in a deterministic way and provide an accompanying function to free?

```
mp_err mp_to_radix(const mp_int *a, char **str, size_t *written, int radix);
/* this re-calculates the allocated size from 'written', frees the buffer and sets *str to NULL */
void mp_free_str(char **str, size_t written);
```

@czurnieden would that be possible from the way the over-estimation works?

minad:
It seems to me that we are overcomplicating things here. In particular if we are adding another mp_free_str function to be used for deallocation. This doesn't look good.

> IIUC size_t size, size_t* written won't help us, it would have to be size_t *allocated, size_t *written ...

Why are we not keeping the API as is - with an array passed by the user, allocated on the stack or the heap by the user. My suggestion would be to just add `size_t size, size_t* written`. This will be safer in contrast to what we have now since we won't write too much if the user passes an array which is too small. Furthermore we give the user some feedback on how much has actually been written. I think this is the simplest, safe extension of the current api.

For sure - this will not be as safe as an internally allocated array. But by requiring a size argument, we make clear what we mean. This is similar to memcpy, memset and all the familiar functions.
I would prefer that to a non-obvious function with many arguments, additional special free function etc. An obvious api is better!

minad:
@sjaeckel If you still want to consider the internal allocation solution, I would propose to introduce a `mp_buf` struct, which contains size, written and data fields. Then we could add a function `mp_free_buf` which is again obvious.

But let's go with the simpler solution...

sjaeckel:
okay, no allocation inside

```
/* written can be NULL */
mp_err mp_to_radix(const mp_int *a, char *str, size_t maxsize, size_t *written, int radix);
```

minad:
Yes and let's do the same for the bin functions. For export we can also relatively easily add these additional params. @czurnieden what do you think?

czurnieden:
> maybe it'd make more sense to over-allocate in a deterministic way and provide an accompanying function to free?
[...]
> would that be possible from the way the over-estimation works?

No, not really, sorry. The one-off is a rounding error that can only be avoided by allowing more precision and, even without any error-analysis done, I can tell you that the tables would get much larger.

We could use the MSB of `*written` to signal if it is one-off. But only without MP_8BIT where it can be off-by-two and only as long as the length of the bigint array is measured in `int` which also assumes that `sizeof(int) < sizeof(size_t)`.

Nah, I don't know&hellip;

> `mp_err mp_to_radix(const mp_int *a, char *str, size_t maxsize, size_t *written, int radix);`
> @czurnieden what do you think?

Well, at first: let me&hellip;
> I know, I really should stop trying to keep up with every single move *sigh*
&hellip;take that back ;-)

You seem to have settled on that, it was one of my proposals, so: let's do it!

> Yes and let's do the same for the bin functions. For export we can also relatively easily add these additional params.

OK. (will handle the `*_bin*` functions in another PR, don't want to clobber this PR too much)

czurnieden:
Sorry for the delay, was hunting a memory-leak that wasn't one because the mixture of "installed from the distribution's repository" and "build from source" seems to make clang acting a bit weirdly, to say the least. That together with the ~30 minutes runtime of `valgrind` with `--track-origins=yes` on my machine&hellip;

I'm sure there are better ways to spend a Saturday night ;-)

Ah, great, forgot to rename the call  `mp_radix_sizeinbase` to `mp_radix_size_overestimate` in `bn_mp_fwrite.c`. But I can do that, when I squash it.

Apropos squashing: if you are okay with this in general (but see the TODO comment in `bn_mp_to_radix.c`) I would repair all of the "typos", bad formatting (get the `goto`s in line) and all the other things you'll find, and squash it down to commit de06f81be2398b887e5e47d8935bc4d1b3436ab9 "make mp_to_radix return the count of characters of the converted number".

If that is in `develop` I can rebase #343 to get merged, close those two PRs and go get a nice cold one before I attack the `*_bin*` functions which don't even have unit-tests yet.

And take a look at `helper.pl` because it starts to get a bit annoying.

tommath.h
```diff
@@ -729,10 +729,10 @@ mp_err mp_fwrite(const mp_int *a, int radix, FILE *stream) MP_WUR;
...
 #define mp_tohex(M, S)     (MP_DEPRECATED_PRAGMA("replaced by mp_to_hex")     mp_toradix((M), (S), 16))
 
-#define mp_to_binary(M, S, N)  mp_to_radix((M), (S), (N), 2)
-#define mp_to_octal(M, S, N)   mp_to_radix((M), (S), (N), 8)
-#define mp_to_decimal(M, S, N) mp_to_radix((M), (S), (N), 10)
-#define mp_to_hex(M, S, N)     mp_to_radix((M), (S), (N), 16)
+#define mp_to_binary(M, S, N)  mp_to_radix((M), (S), (N), NULL, 2)
+#define mp_to_octal(M, S, N)   mp_to_radix((M), (S), (N), NULL, 8)
+#define mp_to_decimal(M, S, N) mp_to_radix((M), (S), (N), NULL, 10)
+#define mp_to_hex(M, S, N)     mp_to_radix((M), (S), (N), NULL, 16)
```
minad: M, S, N, W?

czurnieden: (The squash I just did is just a clean-up, not an end-it-all!)

> M, S, N, W?
Yeah, that's a good question.
I didn't add it at the end because it is a macro that is meant, in my opinion, to be a *simple* shortcut with intentionally reduced functionality. But if you want it: no problem.

czurnieden:
> and go get a nice cold one before I attack the *_bin* functions which don't even have unit-tests yet.

Done in #354 

@minad I couldn't get the test running, it says `NOP mp_read_write_binaries`. Probably just some idiocy done by me, but I would it appreciate highly if you take a look, too, if you don't mind.

minad:
@czurnieden Did you manage to get this NOP issue fixed? I think @sjaeckel introduced this during some tweaking of the test harness (filtering etc). But I didn't use this yet.

tommath.h
sjaeckel: I already thought about overloading macros with different number of arguments, but unfortunately that's only doable for gcc etc. ... I could live with both MSN or MSNW

minad: Ok, @czurnieden is right. The written argument can be omitted here for the simplified/specialised macros.

sjaeckel:
> Did you manage to get this NOP issue fixed?

I can't see any more NOP's in either of the two PR's

czurnieden:
> Did you manage to get this NOP issue fixed?

Fixed. It was my fault: a typo that I wasn't able to recognize for some reason, despite it&mdash;the typo, that is&mdash;screaming at me the whole time.
What did I do: well, instead of using the inner part `MP_FOOBAR` of the defining macro `BN_MP_FOOBAR_C` I C&P'd it from the end, so I had `MP_FOOBAR_C`.

Yes, I'm an idiot, sue me ;-)

minad:
Status? Could you please squash?

czurnieden:
We are only waiting for your OK, so if you are OK with it, drop me a note and I will squash&finish.
Yes, I C&P'ed ;-)

demo/test.c
```diff
@@ -1221,10 +1223,11 @@ static int test_mp_montgomery_reduce(void)
             if (mp_cmp(&c, &d) != MP_EQ) {
 /* *INDENT-OFF* */
                printf("d = e mod a, c = e MOD a\n");
-               mp_to_decimal(&a, buf, sizeof(buf)); printf("a = %s\n", buf);
-               mp_to_decimal(&e, buf, sizeof(buf)); printf("e = %s\n", buf);
-               mp_to_decimal(&d, buf, sizeof(buf)); printf("d = %s\n", buf);
-               mp_to_decimal(&c, buf, sizeof(buf)); printf("c = %s\n", buf);
+               mp_to_decimal(&a, buf, SIZE_MAX); printf("a = %s\n", buf);
```
minad: SIZE_MAX?

minad:
Looks good

bn_deprecated.c
```diff
@@ -113,7 +113,7 @@ mp_err mp_toom_sqr(const mp_int *a, mp_int *b)
 #ifdef S_MP_REVERSE_C
 void bn_reverse(unsigned char *s, int len)
 {
-   s_mp_reverse(s, len);
+   s_mp_reverse(s, (size_t)len);
```
sjaeckel: please check for `len < 0`

bn_s_mp_reverse.c
```diff
@@ -6,7 +6,7 @@
 /* reverse an array, used for radix code */
 void s_mp_reverse(unsigned char *s, int len)
```
sjaeckel: oh I just realized that it isn't done here... can you please change `len` to be `size_t`

bn_mp_to_radix.c
```diff
@@ -43,11 +53,12 @@ mp_err mp_to_radix(const mp_int *a, char *str, size_t maxlen, int radix)
...
       --maxlen;
    }
-
-   digs = 0;
+   digs = 0u;
    while (!MP_IS_ZERO(&t)) {
       if (--maxlen < 1u) {
          /* no more room */
+         /* TODO: It could mimic mp_to_radix_n if that is not an error
+                  or at least not this error (MP_ITER or a new one?). */
```
sjaeckel: maybe add an `MP_BOF`? But we should add that in another PR as this will be used in more functions than this one!

demo/test.c
sjaeckel: I guess that was a 'find&replace all' going haywire

bn_mp_to_radix.c
minad: Good idea! But I would prefer MP_OVF or sth else. BOF reminds me of EOF.

sjaeckel: well a regular overflow shouldn't be possible in an MPI library *rimshot* 
... and `MP_BUFFER_OVERFLOW` is too long ... `MP_BOVF`?

minad: Ok or MP_BUF?

sjaeckel: I thought about that too... fine by me.

demo/test.c
czurnieden: > I guess that was a 'find&replace all' going haywire

After all of that to&fro it makes me wonder that it isn't more ;-)

bn_mp_to_radix.c
czurnieden: > BOF reminds me of EOF but that doesn't make sense
BOF reminds me more of BOFH ;-)

So `MP_BUF` it is?

> we should add that in another PR

There isn't that much else I think, but as it doesn't cost anything: new PR.

minad: See #357

minad:
Squash, rebase and fix CI? But besides that I approve.

czurnieden:
>  fix CI?

CI?

czurnieden:
@minad 
> fix CI?

Oh, you mean Travis (**c**ontinuous **i**ntegration)?
I cannot repair it, I don't work here.

Because I just cannot resist: it got a bit down here with the new owner, didn't it? >;->
\*ducks&runs\*

czurnieden:
> I cannot repair it, I don't work here.

Ah, no that was over at #348 
(I really shouldn't do that many PRs in parallel!)

minad:
@sjaeckel your turn :)

minad:
@czurnieden please rebase

minad:
@czurnieden thanks :) @sjaeckel from my side this looks ready.

PR #351, opened 2019-09-09
Add message that MP-8BIT is deprecated

czurnieden:
As suggested by @sjaeckel in #349

Maybe an "issue", too (mainly for Google) and a note in the Wiki?


sjaeckel:
as I thought, can you please also remove all the 8bit builds from travis?

czurnieden:
Ah, forgot: warnings are taken as errors.

PR #352, opened 2019-09-10
Deduplicate the test binary

sjaeckel:
Create `mtest_opponent` instead of alternative version of `test`

### DIFF
--- a/makefile.msvc
+++ b/makefile.msvc
@@ -67,14 +67,15 @@ $(OBJECTS): $(HEADERS)
 $(LIBMAIN_S): $(OBJECTS)
 	lib /out:$(LIBMAIN_S) $(OBJECTS)
 
-#Build test_standalone suite
-test.exe: $(LIBMAIN_S) demo/main.c demo/opponent.c demo/test.c
-	cl $(LTM_CFLAGS) $(TOBJECTS) $(LIBMAIN_S) $(LTM_LDFLAGS) demo/main.c demo/opponent.c demo/test.c /DLTM_DEMO_TEST_VS_MTEST=0 /Fe$@
+#Build test suite
+test.exe: $(LIBMAIN_S) demo/shared.obj demo/test.obj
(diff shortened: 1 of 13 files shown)


makefile.msvc
```diff
@@ -68,11 +68,12 @@ $(LIBMAIN_S): $(OBJECTS)
...
 
 #Build test_standalone suite
-test.exe: $(LIBMAIN_S) demo/main.c demo/opponent.c demo/test.c
-	cl $(LTM_CFLAGS) $(TOBJECTS) $(LIBMAIN_S) $(LTM_LDFLAGS) demo/main.c demo/opponent.c demo/test.c /DLTM_DEMO_TEST_VS_MTEST=0 /Fe$@
+test.exe: $(LIBMAIN_S) demo/shared.obj demo/test.obj
+	cl $(LTM_CFLAGS) $(TOBJECTS) $(LIBMAIN_S) $(LTM_LDFLAGS) demo/shared.c demo/test.c /Fe$@
 	@echo NOTICE: start the tests by launching test.exe
 
 test_standalone: test.exe
+	@echo Please use make-target 'test.exe'
```
minad: deprecated make command? Is this really necessary. I think not, but if you insist - maybe add deprecated to the message for easier grepping. This will be useful when the time comes to remove the cruft :)

makefile.shared
```diff
@@ -79,8 +79,11 @@ uninstall:
 	rm $(HEADERS_PUB:%=$(DESTDIR)$(INCPATH)/%)
 	rm $(DESTDIR)$(LIBPATH)/pkgconfig/libtommath.pc
 
-test test_standalone: demo/main.o demo/opponent.o demo/test.o $(LIBNAME)
-	$(LTLINK) $(LTM_LDFLAGS) $^ -o test
+test_standalone: test
+	@echo "Please use make-target 'test'"
```
minad: same here

minad:
Not really related but could we also rename demo/ to test/?

sjaeckel:
> Not really related but could we also rename demo/ to test/?

I think after 1.2.0 is out would be a good time, okay?

PR #355, opened 2019-09-27
mp_export: Addition of buffer size and written bytes

czurnieden:
First sketch.

### DIFF
--- a/bn_mp_pack.c
+++ b/bn_mp_pack.c
@@ -1,49 +1,60 @@
 #include "tommath_private.h"
-#ifdef BN_MP_EXPORT_C
+#ifdef BN_MP_PACK_C
 /* LibTomMath, multiple-precision integer library -- Tom St Denis */
 /* SPDX-License-Identifier: Unlicense */
+
+
 
 /* based on gmp's mpz_export.
  * see http://gmplib.org/manual/Integer-Import-and-Export.html
  */
-mp_err mp_export(void *rop, size_t *countp, int order, size_t size,
-                 int endian, size_t nails, const mp_int *op)
+mp_err mp_pack(void *rop, size_t *countp, mp_order order, size_t size,
+               mp_endian endian, size_t nails, const mp_int *op,
+               size_t maxsize)
 {
    mp_err err;
-   size_t odd_nails, nail_bytes, i, j, bits, count;
+   size_t odd_nails, nail_bytes, i, j, count, written;
    unsigned char odd_nail_mask;
 
    mp_int t;
+
+   if (rop == NULL) {
+      return MP_MEM;
+   }
+
+   if (maxsize == 0u) {
+      return MP_VAL;
+   }
 
    if ((err = mp_init_copy(&t, op)) != MP_OKAY) {
       return err;
    }
 
-   if (endian == 0) {
-      union {
-         unsigned int i;
-         char c[4];
-      } lint;
-      lint.i = 0x01020304;
-
-      endian = (lint.c[0] == '\x04') ? -1 : 1;
+   if (endian == MP_NATIVE_ENDIAN) {
+      MP_GET_ENDIANNESS(endian);
    }
 
    odd_nails = (nails % 8u);
    odd_nail_mask = 0xff;
-   for (i = 0; i < odd_nails; ++i) {
+   for (i = 0u; i < odd_nails; ++i) {
       odd_nail_mask ^= (unsigned char)(1u << (7u - i));
    }
    nail_bytes = nails / 8u;
 
-   bits = (size_t)mp_count_bits(&t);
-   count = (bits / ((size * 8u) - nails)) + (((bits % ((size * 8u) - nails)) != 0u) ? 1u : 0u);
+   count = mp_pack_count(&t, nails, size);
 
-   for (i = 0; i < count; ++i) {
-      for (j = 0; j < size; ++j) {
+   written = 0u;
+   for (i = 0u; i < count; ++i) {
+      for (j = 0u; j < size; ++j) {
          unsigned char *byte = (unsigned char *)rop +
-                               (((order == -1) ? i : ((count - 1u) - i)) * size) +
-                               ((endian == -1) ? j : ((size - 1u) - j));
+                               (((order == MP_LSB_FIRST) ? i : ((count - 1u) - i)) * size) +
+                               ((endian == MP_LITTLE_ENDIAN) ? j : ((size - 1u) - j));
+         if (maxsize == 0u) {
+            err = MP_VAL;
+            break;
+         }
+         maxsize--;
+         written++;
 
          if (j >= (size - nail_bytes)) {
             *byte = 0;
@@ -55,11 +66,12 @@ mp_err mp_export(void *rop, size_t *countp, int order, size_t size,
          if ((err = mp_div_2d(&t, (j == ((size - nail_bytes) - 1u)) ? (int)(8u - odd_nails) : 8, &t, NULL)) != MP_OKAY) {
             goto LBL_ERR;
          }
+
       }
    }
 
    if (countp != NULL) {
-      *countp = count;
+      *countp = written;
    }
    err = MP_OKAY;
 
@@ -67,5 +79,6 @@ LBL_ERR:
    mp_clear(&t);
    return err;
 }
+
 
 #endif
--- a/demo/test.c
+++ b/demo/test.c
@@ -2345,6 +2345,53 @@ LTM_ERR:
    return EXIT_FAILURE;
 }
 
+static int test_mp_pack_unpack(void)
+{
+   mp_int a, b;
+   int err;
+   size_t countp, size, bits, bytes;
+   unsigned char *buf = NULL;
+
+   mp_order order = MP_LSB_FIRST;
+   mp_endian endianess = MP_NATIVE_ENDIAN;
+
+   if ((err = mp_init_multi(&a, &b, NULL)) != MP_OKAY)                       goto LTM_ERR;
+   if ((err = mp_rand(&a, 15)) != MP_OKAY)                                   goto LTM_ERR;
+
+   bits = (size_t)mp_count_bits(&a);
+   bytes = (size_t)(((bits + 1u) + (size_t)(CHAR_BIT - 1)) / (size_t)CHAR_BIT);
+   size = bytes * mp_pack_count(&a, 0, 1);
+
+   buf = MP_MALLOC(size);
+   if (buf == NULL) {
+      fprintf(stderr, "test_pack_unpack failed to allocate %zu bytes\n",
+              sizeof(*buf) * size);
+      goto LTM_ERR;
+   }
+
+   if ((err = mp_pack((void *)buf, &countp, order, 1,
+                      endianess, 0, &a, size)) != MP_OKAY)                   goto LTM_ERR;
+   printf("size = %zu, countp = %zu, countp * byte_size = %zu\n",
+          size, countp, countp * bits);
+   if ((err = mp_unpack(&b, countp, order, 1,
+                        endianess, 0, (const void *)buf)) != MP_OKAY)        goto LTM_ERR;
+
+   if (mp_cmp(&a, &b) != MP_EQ) {
+      fprintf(stderr, "pack/unpack cycle failed\n");
+      goto LTM_ERR;
+   }
+
+   MP_FREE(buf, size);
+   mp_clear_multi(&a, &b, NULL);
+   return EXIT_SUCCESS;
+LTM_ERR:
+   if (buf != NULL) {
+      MP_FREE(buf, size);
+   }
+   mp_clear_multi(&a, &b, NULL);
+   return EXIT_FAILURE;
+}
+
 static int unit_tests(int argc, char **argv)
 {
    static const struct {
@@ -2364,6 +2411,7 @@ static int unit_tests(int argc, char **argv)
       T1(mp_decr, MP_DECR),
       T1(mp_div_3, MP_DIV_3),
       T1(mp_dr_reduce, MP_DR_REDUCE),
+      T2(mp_pack_unpack,MP_PACK, MP_UNPACK),
       T2(mp_fread_fwrite, MP_FREAD, MP_FWRITE),
       T1(mp_get_u32, MP_GET_I32),
       T1(mp_get_u64, MP_GET_I64),
--- a/doc/bn.tex
+++ b/doc/bn.tex
@@ -2176,12 +2176,38 @@ They operate essentially the same as the unsigned copies except they prefix the 
 byte depending on the sign.  If the sign is zpos (e.g. not negative) the prefix is zero, otherwise the prefix
 is non--zero.
 
-The two functions \texttt{mp\_import} and \texttt{mp\_export} implement the corresponding GMP functions as described at \url{http://gmplib.org/manual/Integer-Import-and-Export.html}.
-\index{mp\_import} \index{mp\_export}
+The two functions \texttt{mp\_unpack} (get your gifts out of the box, import binary data) and \texttt{mp\_pack} (put your gifts into the box, export binary data) implement the similarly working GMP functions as described at \url{http://gmplib.org/manual/Integer-Import-and-Export.html} with the exception that \texttt{mp\_pack} will not allocate memory if \texttt{rop} is \texttt{NULL}.
+\index{mp\_unpack} \index{mp\_pack}
 \begin{alltt}
-int mp_import(mp_int *rop, size_t count, int order, size_t size, int endian, size_t nails, const void *op);
-int mp_export(void *rop, size_t *countp, int order, size_t size, int endian, size_t nails, const mp_int *op);
+int mp_unpack(mp_int *rop, size_t count, mp_order order, size_t size,
+             mp_endian endian, size_t nails, const void *op, size_t maxsize);
+int mp_pack(void *rop, size_t *countp, mp_order order, size_t size,
+             mp_endian endian, size_t nails, const mp_int *op);
 \end{alltt}
+The function \texttt{mp\_pack} has the additional variable \texttt{maxsize} which must hold the size of the buffer \texttt{rop} in bytes. Use
+\begin{alltt}
+/* Parameters "nails" and "size" are the same as in mp_pack */
+size_t mp_pack_size(mp_int *a, size_t nails, size_t size);
+\end{alltt}
+To get the size in bytes necessary to be put in \texttt{maxsize}).
+
+To enhance the readability of your code, the following enums have been wrought for your convenience.
+\begin{alltt}
+#ifdef MP_USE_ENUMS
+typedef enum {
+   MP_LSB_FIRST = -1,
+   MP_MSB_FIRST =  1
+} mp_order;
+typedef enum {
+   MP_LITTLE_ENDIAN  = -1,
+   MP_NATIVE_ENDIAN  =  0,
+   MP_BIG_ENDIAN     =  1
+} mp_endian;
+\end{alltt}
+Also available as preprocessor macros.
+
+
+
 
 \chapter{Algebraic Functions}
 \section{Extended Euclidean Algorithm}
--- a/tommath.h
+++ b/tommath.h
@@ -136,6 +136,15 @@ typedef enum {
    MP_VAL   = -3,
    MP_ITER  = -4
 } mp_err;
+typedef enum {
+   MP_LSB_FIRST = -1,
+   MP_MSB_FIRST =  1
+} mp_order;
+typedef enum {
+   MP_LITTLE_ENDIAN  = -1,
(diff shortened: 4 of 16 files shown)


bn_mp_export.c
```diff
@@ -7,14 +7,23 @@
  * see http://gmplib.org/manual/Integer-Import-and-Export.html
  */
 mp_err mp_export(void *rop, size_t *countp, int order, size_t size,
```
minad: countp == written??

minad:
See my comment about written. Furthermore if we really want to improve this function, we should add enums for order and endian. How can we handle the deprecation?

bn_mp_export.c
minad: Maybe rename countp to written? But then we are diverging from the gmp api. But the API is already not 100% similar, since GMP also does internal allocation for rop==NULL. But we should keep it simple as it is.

czurnieden:
> we should add enums for order and endian.

Yepp, would look better.

> How can we handle the deprecation?

Good question. The way `bn_deprecated.c` works it would need a different function-name, so: rename?

bn_mp_export.c
minad: @czurnieden I don't think we need both countp and written?

czurnieden: > countp == written??

No, they are not equal, they count two different things.

minad: written=count*size? We don't need this. mp_export is already complicated enough.

sjaeckel:
> > How can we handle the deprecation?
> 
> Good question. The way `bn_deprecated.c` works it would need a different function-name, so: rename?

I first thought about renaming it to `mp_{ex,im}port_gmp()`, but that'd be lying as the current implementation is exactly that (besides the parts where it isn't...) ... this would have been a good name in the first place for those functions... probably `mp_{ex,im}port_safe()` ?
Or maybe `mp_{ex,im}port_gmp_style()` or `mp_{ex,im}port_gmp_compat()`?

I also already thought about having one of those replacements for 1.2 and after 1.2 is out change the API of `mp_{ex,im}port()`? Would that be acceptable?

minad:
Hmm, I would rather introduce new names. For example mp_pack/mp_unpack?

minad:
mp_to_packed/mp_from_packed would be more consistent with the from_to_bin functions?

sjaeckel:
> For example mp_pack/mp_unpack?

it seems you're also one of the goto guys if one has problems with the second of the two hard problems of CS ;-) 

> mp_to_packed/mp_from_packed would be more consistent with the from_to_bin functions?

I like the first better... they sound natural, whereas those sound pushed to be consistent...

minad:
:-P

czurnieden:
> mp_pack/mp_unpack

Yepp, sounds good.
Do it?

bn_mp_export.c
minad: We should also use enums for the flags. This will make things more clear. Maybe add this to this PR?

czurnieden: > We should also use enums for the flags

As you are now our semi-official "namer of functions and stuff": any suggestions?

tommath.h
```diff
@@ -352,11 +368,20 @@ mp_err mp_init_copy(mp_int *a, const mp_int *b) MP_WUR;
 void mp_clamp(mp_int *a);
 
 /* import binary data */
-mp_err mp_import(mp_int *rop, size_t count, int order, size_t size, int endian, size_t nails, const void *op) MP_WUR;
+
+MP_DEPRECATED(mp_unpack) mp_err mp_import(mp_int *rop, size_t count, int order,
+      size_t size, int endian, size_t nails,
+      const void *op) MP_WUR;
+
+mp_err mp_unpack(mp_int *rop, size_t count, int order, size_t size, int endian,
```
minad: Why not use mp_endian etc here?

czurnieden: > Why not use mp_endian etc here?

Because I forgot? ;-)

Thanks!

minad:
What's the status here? Ready?

czurnieden:
> What's the status here? Ready?

If you are OK with it, drop me a note and I will squash&finish.

demo/test.c
```diff
@@ -2214,6 +2214,50 @@ static int test_s_mp_toom_sqr(void)
...
+                      endianess, 0, &a, size)) != MP_OKAY)                   goto LTM_ERR;
+   if ((err = mp_unpack(&b, countp, order, 1,
+                        endianess, 0, (const void *)buf)) != MP_OKAY)        goto LTM_ERR;
+
+   if (mp_cmp(&a, &b) != MP_EQ) {
+      fprintf(stderr, "pack/unpack cycle failed\n");
+      goto LTM_ERR;
+   }
+
+   free(buf);
```
minad: Free or mp_free? Is this consistent in the test suite

minad:
There is an endian==0 comparison

minad:
Countp should probably be set to the written count? Maybe add a function mp_pack_size which does this mp_count_bits?

minad:
An internal helper function for the native endian test woul be nice. Are there more functions which check endianness?

doc/bn.tex
```diff
@@ -2185,12 +2185,41 @@ \section{Binary Conversions}
...
+   MP_MSB_FIRST =  1
+} mp_order;
+typedef enum {
+   MP_LTTLE_ENDIAN  = -1,
+   MP_NATIVE_ENDIAN =  0,
+   MP_BIG_ENDIAN    =  1
+} mp_endian;
+\end{alltt}
+Also available as preprocessor macros.
+\begin{alltt}
```
minad: This is not 100% correct. I would not describe them as enums/ints. After 1.2 we will make MP_USE_ENUMS the default.

minad: Maybe also a bit too verbose docs replicating the code of a triviality.

demo/test.c
```diff
@@ -2214,6 +2214,50 @@ static int test_s_mp_toom_sqr(void)
...
+
+   mp_order order = MP_LSB_FIRST;
+   mp_endian endianess = MP_NATIVE_ENDIAN;
+
+   if ((err = mp_init_multi(&a, &b, NULL)) != MP_OKAY)                       goto LTM_ERR;
+   if ((err = mp_rand(&a, 15)) != MP_OKAY)                                   goto LTM_ERR;
+
+   size = (size_t)(((mp_count_bits(&a) + 1) + 7) / 8);
+
+   buf = MP_MALLOC(size);
```
minad: mp_malloc or malloc? See below.

doc/bn.tex
czurnieden: > Maybe also a bit too verbose docs replicating the code for a triviality.

I can squeeze it into a sentence but I find code snippets are quicker to read for a programmer. But the snippet after "Also available as preprocessor macros." can go, that's right, the sentence introducing them is sufficient.

demo/test.c
czurnieden: I think `MP_MALLOC/MP_FREE` is better to make it more consistent with the rest.

czurnieden:
> Countp should probably be set to the written count? Maybe add a function mp_pack_size which does this mp_count_bits?

We have `written` to mean the size of amount of bytes throughout for now,  `countp` is different in that regard.

But I do support the proposal to put `size = (size_t)(((mp_count_bits(&a) + 1) + 7) / 8);` into an extra function and I don't know why I haven't done it already.

tommath.h
```diff
@@ -167,9 +168,10 @@ typedef int mp_order;
 #define MP_LSB_FIRST -1
 #define MP_MSB_FIRST  1
 typedef int mp_endian;
-#define MP_LTTLE_ENDIAN  -1
+#define MP_LITTLE_ENDIAN  -1
 #define MP_NATIVE_ENDIAN  0
 #define MP_BIG_ENDIAN     1
+#define MP_UNKNOWN_ENDIAN 1
```
minad: Please no unknown endian!

bn_s_mp_check_endianness.c
```diff
@@ -0,0 +1,25 @@
...
+{
+   mp_endian endian;
+   int16_t n;
+   char *p;
+   /*
+       This sorts out many, if not all of the "odd" endiannesses
+       but also modern DSPs with CHAR_BIT > 8.
+    */
+   if (CHAR_BIT != 8) {
+      return MP_UNKNOWN_ENDIAN;
```
minad: No. Rather add a static_assert for charbit==8

minad:
Regarding countp - since we added the maxsize we introduced an additional behavior. In that case we want countp to return the actually written units. This mirrors the behavior of the other functions with written argument.

bn_s_mp_check_endianness.c
```diff
@@ -0,0 +1,25 @@
...
+mp_endian s_mp_check_endianness(void)
+{
+   mp_endian endian;
+   int16_t n;
+   char *p;
+   /*
+       This sorts out many, if not all of the "odd" endiannesses
+       but also modern DSPs with CHAR_BIT > 8.
+    */
+   if (CHAR_BIT != 8) {
```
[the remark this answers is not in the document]
czurnieden: You can never be absolutely sure, could still be 16 (without a real `[u]?int8_t` only `[u]?int_least8_t` and `[u]?int_fast8_t`) or even 32 but I don't work with these MCUs (don't program DSPs myself) so I don't know for sure and if I don't know for sure I check.

tommath.h
czurnieden: The rest of LTM is endianness agnostic.
I think?

bn_s_mp_check_endianness.c
czurnieden: Yes, I forgot that we even have a macro for that.

czurnieden:
> I think MP_GET_ENDIAN should be a macro/inline function in tommath_private since it will get optimized away completely

That's what I started with but than I read that you wanted a function.
\*grr\*!
;-)

demo/test.c
czurnieden: > Is this consistent in the test suite
(I'll take it as a question) The rest of the tests is done with static buffers didn't want to do it here, although it is possible.

demo/test.c
minad: It should just be consistent in the test suite.

minad: But I think malloc makes more sense since we don't want to use a custom malloc in the test suite. This is different from the library code. I think the other tests also use plain malloc.

tommath.h
minad: Yes but we shouldn't added this to the mp_,endian enum values.

tommath_private.h
```diff
@@ -236,8 +236,13 @@ MP_DEPRECATED(s_mp_toom_mul) mp_err mp_toom_mul(const mp_int *a, const mp_int *b
 MP_DEPRECATED(s_mp_toom_sqr) mp_err mp_toom_sqr(const mp_int *a, mp_int *b);
 MP_DEPRECATED(s_mp_reverse) void bn_reverse(unsigned char *s, int len);
 
-/* From the C-FAQ */
-#define MP_ENDIANESS ((0x41424344 == *((uint32_t*)"ABCD"))?MP_BIG_ENDIAN:MP_LITTLE_ENDIAN)
+/* Thanks to PSkocik at stackoverflow for pointing out that C++ has no compound literals */
```
minad: Urks. Why not use only the c++ version. I really dislike this __cplusplus macro.

minad: Hmm but I also don't like this static. Why not keep the ABCD version?

minad: Is this also not c++ compatible? This is a mess. It is not necessary for tommath to be c++ compilable. We also don't test this and compile with g++.

czurnieden: Yepp, it's not the most beautiful of all the macros, so if you insist, I'll delete.

czurnieden: > Why not keep the ABCD version?

Not standard compatible (alignment problems). [I asked at stackoverflow](https://stackoverflow.com/questions/58224612/standard-compliance-of-a-c-macro) to get some hints because I'm not very fluent in C++.
clang complained, too.

> It is not necessary for tommath to be c++ compilable.

If it is cheap to have&hellip;

> We also don't test this and compile with g++ in the CI.

Mmh&hellip; should we?
But now that you mentioned it, I have never tried, so
`CC=g++ make test && ./test`
A lot of warnings of the kind
```text
cc1plus: warning: command line option ‘-Wdeclaration-after-statement’ is valid for C/ObjC but not for C++
cc1plus: warning: command line option ‘-Wbad-function-cast’ is valid for C/ObjC but not for C++
cc1plus: warning: command line option ‘-Wstrict-prototypes’ is valid for C/ObjC but not for C++
```
but otherwise works.

clang++-7 complains with a lot of
```text
clang: warning: treating 'c' input as 'c++' when in C++ mode, this behavior is deprecated [-Wdeprecated]
```

and throws one error

```text
   * clang++-7 demo/test.o
clang: warning: treating 'c' input as 'c++' when in C++ mode, this behavior is deprecated [-Wdeprecated]
demo/test.c:2277:22: error: invalid suffix on literal; C++11 requires a space between literal and identifier
      [-Wreserved-user-defined-literal]
   printf("SEED: 0x%"PRIx64"\n\n", t)
```

So it works in general with a decent C++ compiler and I'd like to keep LTM in that state if possible.

minad: I don't think the c++ variant with the static variable is a good idea. You could also write a different macro which is not an expression.

MP_GET_ENDIAN(endian) do...while etc

czurnieden: > You could also write a different macro which is not an expression

Of course I could.
Shoudl I?
I think I should, yes.

tommath.h
minad: Ping @czurnieden - could you remove unknown_endian everywhere?

bn_mp_pack.c
```diff
@@ -55,11 +65,12 @@ mp_err mp_export(void *rop, size_t *countp, int order, size_t size,
          if ((err = mp_div_2d(&t, (j == ((size - nail_bytes) - 1u)) ? (int)(8u - odd_nails) : 8, &t, NULL)) != MP_OKAY) {
             goto LBL_ERR;
          }
+
       }
    }
 
    if (countp != NULL) {
-      *countp = count;
+      *countp = count * size;
```
minad: No I meant countp should return the number of actually written units. Not the total number of units compute

bn_mp_pack_size.c
```diff
@@ -0,0 +1,10 @@
+#include "tommath_private.h"
+#ifdef BN_MP_PACK_SIZE_C
+/* LibTomMath, multiple-precision integer library -- Tom St Denis */
+/* SPDX-License-Identifier: Unlicense */
+
+size_t mp_pack_size(mp_int *a)
+{
+   return (size_t)(((mp_count_bits(a) + 1) + (CHAR_BIT - 1)) / CHAR_BIT);
```
minad: This should be called mp_pack_count and compute the needed count as computed in mp_pack.

minad: count = (bits / ((size * 8u) - nails)) + (((bits % ((size * 8u) - nails)) != 0u) ? 1u : 0u);

czurnieden: > This should be called mp_pack_count

Someone&mdash;wasn't it you?&mdash;wanted `mp_pack_size`, so I did.

minad: Yes, but I thought a bit about it. If we go with these units of size we should change it. Maxsize should also be maxcount instead? The user should allocate malloc(size*maxcount). countp is also written units.

Alternative would be to go with bytes? countp in bytes, maxsize, pack_size in bytes?

What do you think?

bn_mp_pack.c
```diff
@@ -36,16 +36,17 @@ mp_err mp_pack(void *rop, size_t *countp, mp_order order, size_t size,
...
    odd_nails = (nails % 8u);
    odd_nail_mask = 0xff;
-   for (i = 0; i < odd_nails; ++i) {
+   for (i = 0u; i < odd_nails; ++i) {
       odd_nail_mask ^= (unsigned char)(1u << (7u - i));
    }
    nail_bytes = nails / 8u;
 
    bits = (size_t)mp_count_bits(&t);
    count = (bits / ((size * 8u) - nails)) + (((bits % ((size * 8u) - nails)) != 0u) ? 1u : 0u);
```
minad: Use mp_pack_size/count here?

czurnieden: > Use mp_pack_size/count here?

Yes, of course.

So you are OK with what I did?

minad: Yes, but we still have to decide bytes or units.

Right now countp in units, maxsize in bytes, pack_size in bytes? The same applies to mp_unpack.

czurnieden: I shuffled it a bit around, probably more to what you wanted in the first place.
I hope ;-)

Argh, documentation not yet updated!

> Yes, but we still have to decide bytes or units.

Do you want to name them (`enum`)?
Otherwise the types can stay as they are now.

minad: Hmm I don't like it like this. We have count for unpack which is used to pass in the buffer size and for pack we have both maxsize and countp. Either we also change countp/count to bytes or we change maxsize to units. Mixing the units is not good as we have it now.

czurnieden: (I hope I'm up-to-date now with the rebasing)

> Hmm I don't like it like this.
Yeah, it's a bit of a mess, admitted.

> Either we also change countp/count to bytes or we change maxsize to units.

`maxsize` (I think it's `maxlen` elsewhere but I always mix them up) is in all other cases meant to measure the given buffer-size in bytes, a `size_t` type.

Let me restrict myself to `mp_pack` for now:
`countp` does, as not only the name suggests, some (ac)counting. The size in bytes behind `countp` depends on a second variable (it formerly returned `count` which were units computed from bitsize and nails) and can be different from the amount of actually used up bytes. so I would say that the expectation is until now that `countp` returns units.
You won't have a port to get the amount of written bytes if you use `countp` for units.

On the other side: we have an external function to compute `countp` so the information that`countp` was once needed for is now available externally and in advance, `countp` is now free for other uses, e.g.: returning the number bytes written.
It changes the API but it is not much work to write a little macro as a wrapper if so desired.

So for `mp_pack` I'd propose:
 - `countp` (maybe change the name) are the bytes written
 - `maxsize` is the size in bytes of the buffer given
 - `count` as computed by `mp_pack_count` returns the number of units you get with the combination of bit-size and nails given to `mp_pack_count`. That means that there are no accounts of units send in or out of `mp_pack`.

And for `mp_unpack`
 - `count` is not the number of bytes but the number of units, the actual byte-size of the input is only of internal interest but can be computed from `count` and `nails` if so desired although I would just measure the input.

minad: pack/unpack should also be symmetric. This means count should be bytes/unit for both. However we should also consider that changing from units/bytes will introduce errors for user who do not expect that change when they move to the new function.

minad: My suggestion would therefore be to call it mp_pack_count, count counts unit, maxcount etc. This also makes sense. The user is not operating on byte arrays but on arrays of unit size (like calloc for example). The only downside is that its different from the sbin/ubin functions which operate on bytes.

minad:
@sjaeckel what is your opinion on the bytes vs size-units discussion?

sjaeckel:
> @sjaeckel what is your opinion on the bytes vs size-units discussion?

I would say we keep the semantics as it was.

i.e. the proposed API should change like that:

```
diff --git a/tommath.h b/tommath.h
index 2a37709..6b627ff 100644
--- a/tommath.h
+++ b/tommath.h
@@ -384,3 +384,3 @@ size_t mp_pack_count(mp_int *a, size_t nails, size_t size) MP_WUR;
 mp_err mp_pack(void *rop, size_t *countp, mp_order order, size_t size, mp_endian endian,
-               size_t nails, const mp_int *op, size_t maxsize) MP_WUR;
+               size_t nails, const mp_int *op, size_t maxcount) MP_WUR;
```

`mp_pack` related API's work with `count`'s of elements, buffers have to be `count * size`

IMO it makes sense as a call to `mp_pack()` with `MP_MSB_FIRST`, `MP_BIG_ENDIAN`, `nails=0` and `size=1` is equal to a call of `mp_to_ubin()`, in the result that is written out **and** in the number of written elements ;-)

minad:
I agree with @sjaeckel and I would like to add that mp_pack_count should also return a count. Not a size in bytes.

minad:
Maybe also rearrange the args. void*buf,size_t countmax,size_t*countp

sjaeckel:
> I would like to add that mp_pack_count should also return a count

that's what it does already.

> Maybe also rearrange the args.

there's no real style how ltm treats the order of args, so I'm somehow against reordering only here, but it's true... somehow it'd also make sense to order them like that `mp_err mp_pack(void *rop, size_t maxsize, size_t *countp, ...)`

sjaeckel:
> `mp_err mp_pack(void *rop, size_t maxsize, size_t *countp, ...)`

I just thought about it a bit more and yes, please re-order

minad:
superseded by #361

PR #361, opened 2019-10-08
add mp_pack/unpack and MP_BUF error

minad:
Based on #355 by @czurnieden. Buffer size measured by units as discussed in #355. @czurnieden I hope it is ok to base my PR on yours, since you might be busy with #343 ;) I also wanted to add something instead of only reviewing and nitpicking all the time.

### DIFF
--- a/bn_deprecated.c
+++ b/bn_deprecated.c
@@ -113,10 +113,9 @@ mp_err mp_toom_sqr(const mp_int *a, mp_int *b)
 #ifdef S_MP_REVERSE_C
 void bn_reverse(unsigned char *s, int len)
 {
-   if (len < 0) {
-      return MP_VAL;
+   if (len > 0) {
+      s_mp_reverse(s, (size_t)len);
    }
-   s_mp_reverse(s, (size_t)len);
 }
 #endif
 #ifdef BN_MP_TC_AND_C
@@ -314,6 +313,20 @@ mp_err mp_toradix_n(const mp_int *a, char *str, int radix, int maxlen)
 mp_err mp_toradix(const mp_int *a, char *str, int radix)
 {
    return mp_to_radix(a, str, SIZE_MAX, NULL, radix);
+}
+#endif
+#ifdef BN_MP_IMPORT_C
+mp_err mp_import(mp_int *rop, size_t count, int order, size_t size, int endian, size_t nails,
+                 const void *op)
+{
+   return mp_unpack(rop, count, order, size, endian, nails, op);
+}
+#endif
+#ifdef BN_MP_EXPORT_C
+mp_err mp_export(void *rop, size_t *countp, int order, size_t size,
+                 int endian, size_t nails, const mp_int *op)
+{
+   return mp_pack(rop, SIZE_MAX, countp, order, size, endian, nails, op);
 }
 #endif
 #endif
--- a/bn_mp_pack.c
+++ b/bn_mp_pack.c
@@ -16,7 +16,7 @@ mp_err mp_pack(void *rop, size_t maxcount, size_t *writtencount, mp_order order,
    mp_int t;
 
    if (maxcount == 0u) {
-      return MP_VAL;
+      return MP_BUF;
    }
 
    if ((err = mp_init_copy(&t, op)) != MP_OKAY) {
@@ -38,8 +38,8 @@ mp_err mp_pack(void *rop, size_t maxcount, size_t *writtencount, mp_order order,
 
    for (i = 0u; i < count; ++i) {
       if (i >= maxcount) {
-          err = MP_VAL;
-          break;
+         err = MP_BUF;
+         break;
       }
       for (j = 0u; j < size; ++j) {
          unsigned char *byte = (unsigned char *)rop +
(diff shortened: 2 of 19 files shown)


bn_deprecated.c
```diff
@@ -113,10 +113,9 @@ mp_err mp_toom_sqr(const mp_int *a, mp_int *b)
 #ifdef S_MP_REVERSE_C
 void bn_reverse(unsigned char *s, int len)
 {
-   if (len < 0) {
-      return MP_VAL;
```
minad: this was an error. I don't know how it could manage to get into develop?

bn_mp_to_radix.c
```diff
@@ -16,14 +16,11 @@ mp_err mp_to_radix(const mp_int *a, char *str, size_t maxlen, size_t *written, i
    mp_digit d;
    char   *_s = str;
 
-
-   /* If we want to fill a bucket we need a bucket in the first place. */
-   if (str == NULL) {
-      return MP_VAL;
-   }
```
minad: This check was wrong. It is a programming error, not an error we should handle. I had the same discussion with fperrad a while ago, where he suggested checking every mp_int* for NULL. It does not make sense since we have no means of detecting valid/invalid memory addresses, and NULL is just an invalid address, as is probably 42. This is different than checking maxlen!

bn_mp_to_ubin.c
```diff
@@ -10,12 +10,8 @@ mp_err mp_to_ubin(const mp_int *a, unsigned char *buf, size_t maxlen, size_t *wr
    mp_err  err;
    mp_int  t;
 
-   if (buf == NULL) {
-      return MP_MEM;
-   }
```
minad: Same invalid check.

bn_deprecated.c
minad: See here https://github.com/libtom/libtommath/blob/c5b901b917c613a8b21064f816e6a7c70ed2fa09/bn_deprecated.c#L117

sjaeckel: lol nice

I wonder how no compiler has complained about that!

bn_mp_to_radix.c
sjaeckel: I'd really love to see the architecture where NULL is valid memory, but I get the point.

tommath_private.h
```diff
@@ -236,6 +236,13 @@ MP_DEPRECATED(s_mp_toom_mul) mp_err mp_toom_mul(const mp_int *a, const mp_int *b
 MP_DEPRECATED(s_mp_toom_sqr) mp_err mp_toom_sqr(const mp_int *a, mp_int *b);
 MP_DEPRECATED(s_mp_reverse) void bn_reverse(unsigned char *s, int len);
 
+#define MP_GET_ENDIANNESS(x) \
+   do{\
+      int16_t n = 0x1;                                          \
```
sjaeckel: I already see the "shadowing variable name" warnings pop up ...

bn_mp_pack.c
```diff
@@ -1,49 +1,50 @@
...
    nail_bytes = nails / 8u;
 
-   bits = (size_t)mp_count_bits(&t);
-   count = (bits / ((size * 8u) - nails)) + (((bits % ((size * 8u) - nails)) != 0u) ? 1u : 0u);
+   count = mp_pack_count(&t, nails, size);
 
-   for (i = 0; i < count; ++i) {
-      for (j = 0; j < size; ++j) {
+   for (i = 0u; i < count; ++i) {
+      if (i >= maxcount) {
```
sjaeckel: I just asked myself if it wouldn't make more sense to do this check before allocating the copy? who's interested in partial results anyway?

minad: Sure, but this is how we defined it for the other functions returning MP_BUF too. We return how many bytes/units have been written. The issue is that there is no quick check if the result will fit for to_radix. So we could avoid writing partial results for ubin/sbin/pack. For to_radix we necessarily have to allow partial results. And this is a bit inconsistent. We could also say that in the case MP_BUF is returned, the buffer is in some kind of inconsistent state and the result should not be used. Then we can do whatever.

tommath_private.h
minad: Yes, I would really like if we could use inline functions. But given the limited use of the macro, I think it is ok. There are also no good alternatives. @czurnieden tried writing it first as expression but then there is a c++ incompatibility.

bn_deprecated.c
minad: My compiler does. Maybe the CI was not running for some reason on develop.

bn_mp_pack.c
sjaeckel: > We could also say that in the case MP_BUF is returned, the buffer is in some kind of inconsistent state and the result should not be used

I think we should do that if it makes our code easier! ... and to_radix can stay an exception with well-defined output ... @czurnieden your thoughts?

minad: Well this could even apply to to_radix. The difference is only that to_radix does not allow for an immediate bail out.

minad:
@sjaeckel Things are indeed simpler if we do what you suggested.

bn_mp_pack.c
sjaeckel: AFAIU it won't make the code of to_radix easier, so we simply leave it as it is.

PR #364, opened 2019-10-10
Update to list of number of Miller-Rabin trials

czurnieden:
- Pulled values in `bn_mp_prime_rabin_miller_trials.c` up one notch over the values recommended by the different governmental institutions of various nations.
 - Addition of a couple of lines regarding the differences between DSA and RSA. in the documentation.
 - Put PARI/GP code to calculate the number of MIller-Rabin rounds into the appendix of the documentation

### DIFF
--- a/bn_mp_prime_frobenius_underwood.c
+++ b/bn_mp_prime_frobenius_underwood.c
@@ -7,7 +7,7 @@
 /*
  *  See file bn_mp_prime_is_prime.c or the documentation in doc/bn.tex for the details
  */
-#ifndef LTM_USE_FIPS_ONLY
+#ifndef LTM_USE_ONLY_MR
 
 #ifdef MP_8BIT
 /*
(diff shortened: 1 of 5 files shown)


minad:
:+1: from me

sjaeckel:
If I'm not mistaken this table should reflect C.1 depending on whether `LTM_USE_FIPS_ONLY` is defined... or not?

sjaeckel:
and I'd say `LTM_USE_FIPS_ONLY` should better be called `LTM_PRIMETEST_USE_ONLY_MR`

czurnieden:
> If I'm not mistaken this table should reflect C.1 of FIPS 186.4 depending on whether LTM_USE_FIPS_ONLY is defined... or not?

One notch more than C.1 (and a couple of others) but: no, it is independent, you still have to call it manually. Do you want it automated?

If you hint at automating that choice: I don't know. You don't need these maximum, bordering paranoia values for every usecase, so I'd like to keep the manual input here.

> and I'd say `LTM_USE_FIPS_ONLY` should better be called [...] `LTM_USE_ONLY_MR`

Done.

PR #365, opened 2019-10-10
some linting

fperrad:
after merge of #361

### DIFF
--- a/tommath.h
+++ b/tommath.h
@@ -385,8 +385,8 @@ mp_err mp_unpack(mp_int *rop, size_t count, mp_order order, size_t size, mp_endi
 
 /* pack binary data */
 size_t mp_pack_count(const mp_int *a, size_t nails, size_t size) MP_WUR;
-mp_err mp_pack(void *rop, size_t maxcount, size_t *writtencount, mp_order order, size_t size, mp_endian endian,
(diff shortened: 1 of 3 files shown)


tommath.h
```diff
@@ -385,8 +385,8 @@ mp_err mp_unpack(mp_int *rop, size_t count, mp_order order, size_t size, mp_endi
 
 /* pack binary data */
 size_t mp_pack_count(const mp_int *a, size_t nails, size_t size) MP_WUR;
-mp_err mp_pack(void *rop, size_t maxcount, size_t *writtencount, mp_order order, size_t size, mp_endian endian,
-               size_t nails, const mp_int *op) MP_WUR;
+mp_err mp_pack(void *rop, size_t maxcount, size_t *written, mp_order order, size_t size,
```
sjaeckel: @minad `written` or `writtencount`?

fperrad: `written` is used by `mp_to_ubin`, `mp_to_sbin` and `mp_to_radix`.

sjaeckel: yes, but `written` there is the number of bytes written, whereas `written` here is the number of elements of size `size`

minad: written is ok I guess

PR #368, opened 2019-10-10
Minimal mp_radix_size_overestimate, very large error

czurnieden:
This is the version with a large error range `-0,+x` (with `x < 28,328`)
PR slicing as asked for in #343 
**EDIT:** corrected sign for thousands separator.

PR #369, opened 2019-10-10
mp_radix_size replacement O(1) for all bases, large tables

czurnieden:
**EDIT:** The test-rig had a bug. Result after repair:
<del>This is the version with error range `-0,+1`</del>
This version is exact for input `x` in the range `0 <= x <= 2^(2^31)`) with bases `b` in the range `2 <= b <= 64` and has been fully(!) tested.

PR slicing as asked for in #343

### DIFF
--- a/bn_mp_radix_size_overestimate.c
+++ b/bn_mp_radix_size_overestimate.c
@@ -145,11 +145,7 @@ mp_err mp_radix_size_overestimate(const mp_int *a, const int radix, int *size)
    int bit_count;
    mp_err err = MP_OKAY;
 
-   if (size == NULL) {
-      return MP_VAL;
-   } else {
-      *size = 0;
-   }
+   *size = 0;
 
    if ((radix < 2) || (radix > 64)) {
       return MP_VAL;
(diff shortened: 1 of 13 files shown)


sjaeckel:
how about calling this one `mp_radix_size_overestimate_gmp()`, the version of #368 `mp_radix_size_overestimate_small()`, then adding a `mp_radix_size_overestimate()` which branches depending on `MP_LOW_MEM`?

czurnieden:
I have a faint but nagging idea that there is no overestimation here but a bug in the test code. See comment inside the code of #371 
It will last about three days to check it fully. (one base is about half an hour on my machine and I can only use two cores at once or I'll loose the ability to watch&hellip;uhm&hellip;kitten videos in full 4k and I don't want to let it run unattended because the smoke-alarm is so loud it causes me headaches).

> [...] which branches depending on MP_LOW_MEM

Ah, yes, I shouldn't have ignored that, thanks!

Mmh&hellip;is it (#386 ) even suitable for `MP_LOW_MEM`? It is itself larger, although not much, but it also causes the caller to reserve more memory, although not much in the smaller regions.

czurnieden:
> I have a faint but nagging idea that there is no overestimation here but a bug in the test code.

Yes, three is a bug in the test code. Although both, the `log` and the division by the literal constant for `log(2)` are less than one ULP, both together can accumulate a slightly bigger error in a couple of cases (e.g.: 115 times for base 10, 92 times for base 7) for large input (`>2^26`) changing the result by one bit if it is too close to an integer which causes the final `ceil` to get irritated enough to allow for a wrong output.

Rerunning all(!) of the tests with binary96 (`long double`) to calculate the reference value resulted in zero errors, this table-based algorithm is exact for inputs `0 < x < 2^31`.

The moral of this story? It is not **always** the code you test that's wrong, there are some rare, very rare cases when it it is the reference.

Outch!

That changes the landscape a bit, I think?

czurnieden:
> @czurnieden ok to close this too?

You already did ;-)

So you don't want this table-thing at all?

minad:
Well, who am I to decide. But I think #371 is more reasonable for most use cases.

PR #370, opened 2019-10-11
Addition of fast division (recursive divrem only)

czurnieden:
Direct implementation of algorithms 1.8 "RecursiveDivRem" and 1.9 "UnbalancedDivision"
from:

   Brent, Richard P., and Paul Zimmermann. "Modern computer arithmetic"
   Vol. 18. Cambridge University Press, 2010
   [Available online](https://arxiv.org/pdf/1004.4710)

pages 19ff. in the above online document

### DIFF
--- a/demo/test.c
+++ b/demo/test.c
@@ -2327,6 +2327,139 @@ LBL_ERR:
 }
 
 
+/* Some larger values to test the fast division algorithm */
+static int test_s_mp_div_recursive(void)
+{
+   mp_int a, b, c_q, c_r, d_q, d_r;
+   int size, err;
+
+   if ((err = mp_init_multi(&a, &b, &c_q, &c_r, &d_q, &d_r, NULL)) != MP_OKAY) {
+      goto LBL_ERR;
+   }
+
+   for (size = MP_KARATSUBA_MUL_CUTOFF; size < 3 * MP_KARATSUBA_MUL_CUTOFF; size += 10) {
+      fprintf(stderr,"sizes = %d / %d\n", 10 * size, size);
+      /* Relation 10:1 */
+      if ((err = mp_rand(&a, 10 * size)) != MP_OKAY) {
+         goto LBL_ERR;
+      }
+      if ((err = mp_rand(&b, size)) != MP_OKAY) {
+         goto LBL_ERR;
+      }
+      if ((err = s_mp_div_recursive(&a, &b, &c_q, &c_r)) != MP_OKAY) {
+         goto LBL_ERR;
+      }
+      if ((err = s_mp_div_school(&a, &b, &d_q, &d_r)) != MP_OKAY) {
+         goto LBL_ERR;
+      }
+      if (mp_cmp(&c_q, &d_q) != MP_EQ) {
+         fprintf(stderr, "1. Recursive division failed at sizes %d / %d, wrong quotient\n",
+                 10 * size, size);
+         goto LBL_ERR;
+      }
+      if (mp_cmp(&c_r, &d_r) != MP_EQ) {
+         fprintf(stderr, "1. Recursive division failed at sizes %d / %d, wrong remainder\n",
+                 10 * size, size);
+         goto LBL_ERR;
+      }
+      fprintf(stderr,"sizes = %d / %d\n", 2 * size, size);
+      /* Relation 2:1 */
+      if ((err = mp_rand(&a, 2 * size)) != MP_OKAY) {
+         goto LBL_ERR;
+      }
+      if ((err = mp_rand(&b, size)) != MP_OKAY) {
+         goto LBL_ERR;
+      }
+      if ((err = s_mp_div_recursive(&a, &b, &c_q, &c_r)) != MP_OKAY) {
+         goto LBL_ERR;
+      }
+      if ((err = s_mp_div_school(&a, &b, &d_q, &d_r)) != MP_OKAY) {
+         goto LBL_ERR;
+      }
+      if (mp_cmp(&c_q, &d_q) != MP_EQ) {
+         fprintf(stderr, "2. Recursive division failed at sizes %d / %d, wrong quotient\n",
+                 2 * size, size);
+         goto LBL_ERR;
+      }
+      if (mp_cmp(&c_r, &d_r) != MP_EQ) {
+         fprintf(stderr, "2. Recursive division failed at sizes %d / %d, wrong remainder\n",
+                 2 * size, size);
+         goto LBL_ERR;
+      }
+      fprintf(stderr,"sizes = %d / %d\n", 3 * size, 2 * size);
+      /* Upper limit 3:2 */
+      if ((err = mp_rand(&a, 3 * size)) != MP_OKAY) {
+         goto LBL_ERR;
+      }
+      if ((err = mp_rand(&b, 2 * size)) != MP_OKAY) {
+         goto LBL_ERR;
+      }
+      if ((err = s_mp_div_recursive(&a, &b, &c_q, &c_r)) != MP_OKAY) {
+         goto LBL_ERR;
+      }
+      if ((err = s_mp_div_school(&a, &b, &d_q, &d_r)) != MP_OKAY) {
+         goto LBL_ERR;
+      }
+      if (mp_cmp(&c_q, &d_q) != MP_EQ) {
+         fprintf(stderr, "3. Recursive division failed at sizes %d / %d, wrong quotient\n",
+                 3 * size, 2 * size);
+         goto LBL_ERR;
+      }
+      if (mp_cmp(&c_r, &d_r) != MP_EQ) {
+         fprintf(stderr, "3. Recursive division failed at sizes %d / %d, wrong remainder\n",
+                 3 * size, 2 * size);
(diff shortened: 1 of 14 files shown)


MasterDuke17:
Should this change anything about my in-progress implementation of the faster to_radix?

czurnieden:
@MasterDuke17 
> Should this change anything about my in-progress implementation of the faster to_radix?

No.
I'm just offering it, it may or may not find it's way into LTM eventually.

minad:
I would like to have this :+1:

minad:
Could you refactor this a bit, introducing a s_mp_div_small function? Then disabling s_mp_div_recursive/school would result in the small function being used.

~~~ c
   if (MP_HAS(S_MP_DIV_RECURSIVE)
       && (b->used > MP_KARATSUBA_MUL_CUTOFF)
       && (b->used <= ((a->used)/3*2))) {
      err = s_mp_div_recursive(a, b, c, d);
   } else if (MP_HAS(S_MP_DIV_SCHOOL)) {
      err = s_mp_div_school(a, b, c, d);
   } else {
      err = s_mp_div_small(a, b, c, d);
   }
~~~

czurnieden:
> Could you refactor this a bit, introducing a s_mp_div_small function?

Yepp, no prob'.

> It seems BN_MP_DIV_SMALL is the only macro we have which enables an alternative version being built, circumventing the MP_HAS configuration system.

As we are working towards 2.0.0 now (Ignoring the bugfix-versions 1.2.x): we **could** change it and include it in the `MP_HAS` configuration system?

bn_s_mp_div_small.c
```diff
@@ -0,0 +1,72 @@
...
+      if (d != NULL) {
+         err = mp_copy(a, d);
+      } else {
+         err = MP_OKAY;
+      }
+      if (c != NULL) {
+         mp_zero(c);
+      }
+      return err;
+   }
```
minad: Code duplication?

bn_s_mp_div_small.c
```diff
@@ -0,0 +1,72 @@
+#include "tommath_private.h"
+#ifdef BN_S_MP_DIV_SMALL_C
+/* LibTomMath, multiple-precision integer library -- Tom St Denis */
+/* SPDX-License-Identifier: Unlicense */
+
+#ifdef BN_MP_DIV_SMALL
```
minad: Remove old guard

czurnieden: Just two weeks before he is eligible to his full pension?
Ow, c'mon, really?

czurnieden:
The snippet

```c
   if (MP_HAS(S_MP_DIV_RECURSIVE)
       && (b->used > MP_KARATSUBA_MUL_CUTOFF)
       && (b->used <= ((a->used)/3*2))) {
      err = s_mp_div_recursive(a, b, c, d);
   } else if (MP_HAS(S_MP_DIV_SCHOOL)) {
      err = s_mp_div_school(a, b, c, d);
   } else {
      err = s_mp_div_small(a, b, c, d);
   }
```

Does not seem to work as intended (with `LTM_NOTHING`), any suggestions besides bracketing them off the old preprocessor way?

tommath_superclass.h
```diff
@@ -67,7 +67,7 @@
 #   define BN_S_MP_REVERSE_C
 
 /* other modifiers */
-#   define BN_MP_DIV_SMALL                    /* Slower division, not critical */
+#   define BN_S_MP_DIV_SMALL                    /* Slower division, not critical */
```
minad: This line should go away completely.

bn_s_mp_div_school.c
```diff
@@ -0,0 +1,160 @@
+#include "tommath_private.h"
+#ifdef BN_S_MP_DIV_SCHOOL_C
+/* LibTomMath, multiple-precision integer library -- Tom St Denis */
+/* SPDX-License-Identifier: Unlicense */
+
+#ifndef BN_S_MP_DIV_SMALL
```
minad: Remove guard

bn_s_mp_div_recursive.c
```diff
@@ -0,0 +1,185 @@
+#include "tommath_private.h"
+#ifdef BN_S_MP_DIV_RECURSIVE_C
+
+/* LibTomMath, multiple-precision integer library -- Tom St Denis */
+/* SPDX-License-Identifier: Unlicense */
+
+#ifndef BN_S_MP_DIV_SMALL
```
minad: Remove guard

minad:
What do you mean - does not work as intended? If you compile with LTM_NOTHING, only s_mp_div_small will be used as I see it.

minad:
Yes, we should rework the configuration system. Optional dependencies (guarded by MP_HAS) should not be required automatically in tommath_class.h. I made such suggestions already in #301. But this is independent of this PR and we can address this later.

czurnieden:
> we can address this later.

OK.
So I removed *all* `BN_S_MP_DIV_SMALL` guards (including the definition in `tommath_superclass.h`).

> [...] in #301. 

It is still on the TODO list, no worry.
I think I'll boldly add the 2.0.0 milestone.

minad:
@czurnieden this PR seems almost ready. Can you rebase it?

czurnieden:
> this PR seems almost ready. 

Bugfixes not withstanding, of course, but yes.
Changed label accordingly.

demo/test.c
```diff
@@ -2294,37 +2294,112 @@ static int test_mp_radix_size(void)
    /* number to result in a different size for every base: 67^(4 * 67) */
    mp_set(&a, 67);
    if ((err = mp_expt_u32(&a, 268u, &a)) != MP_OKAY) {
-      goto LBL_ERR;
+      goto LTM_ERR;
```
minad: Use LBL_ for labels now. It is more consistent like this over the whole codebase.

demo/test.c
```diff
@@ -2294,37 +2294,112 @@ static int test_mp_radix_size(void)
...
+      fprintf(stderr,"sizes = %d / %d\n", 2 * size, size);
+      /* Relation 2:1 */
+      if ((err = mp_rand(&a, 2 * size)) != MP_OKAY) {
+         goto LTM_ERR;
+      }
+      if ((err = mp_rand(&b, size)) != MP_OKAY) {
+         goto LTM_ERR;
+      }
+      if ((err = s_mp_div_recursive(&a, &b, &c, NULL)) != MP_OKAY) {
+         goto LTM_ERR;
```
minad: Compare the mod too? Maybe write this function as a helper with function pointers and then call div_test(school, recursive)
div_test(school, small)?

demo/test.c
czurnieden: Leftover from rebasing. The merging process  insisted on mangling `test_mp_radix_size` and `test_mp_div` together and it needed quite some fuddling to get them apart. Don't know why it did it.

So `LBL_` is the prefix for a **L**A**B**E**L** now?
Can I put it in the [Wiki](https://github.com/libtom/libtommath/wiki/The-Sourcecode:-Formatting-and-Style)?

demo/test.c
czurnieden: > Compare the mod too?

Yepp, can do, no prob'.

> Maybe write this function as a helper with function pointers and then call div_test(school, recursive), div_test(school, small)?

Well, C&P with replace would be simpler, wouldn't it? ;-)
`div_small` is quite slow and as it does not get used with recursion (save a couple of bytes with `div_small` and than add a load full with recursion?) it might need too much runtime but haven't checked yet.

demo/test.c
```diff
@@ -2325,6 +2325,140 @@ static int test_mp_radix_size(void)
    return EXIT_FAILURE;
 }
 
+#ifndef S_MP_DIV_SMALL
```
minad: Remove the guard and test both functions test_s_mp_div_recursive, trst_s_mp_div_small. Additional guards are rarely necessary. MP_HAS takes care of it.

minad:
@czurnieden please squash @sjaeckel this looks ready from my side

minad:
I forgot - does it make sense to add an additional cut off here? Instead of using karatsuba?

czurnieden:
" out-of-date with the base branch"?
You are really busy the last couple of days!

> I forgot - does it make sense to add an additional cut off here? Instead of using karatsuba?

Found no differences above Karatsuba only below (it needs fast multiplication to function). The larger difference is in the relation numerator/denominator where it stops being faster approaching ~2/3 and even starts to get slower above 0.8.

I don't think that any kind of tuning would make sense.

minad:
Hmm ok, but I mean is there a theoretical reason why the cutoff should be the same? If there is none, I would prefer if you introduce another constant (even if the value is the same).

czurnieden:
> Hmm ok, but I mean is there a theoretical reason why the cutoff should be the same?

As I said: it needs fast multiplication to function and the lowest cutoff for that is the Karatsuba cutoff; there is a direct connection, not just coincidence.

We can ignore Comba here because the most likely reason for not wanting the Comba algorithm is lack of memory which means there is no other fast multiplication in that case and I'm pretty sure no space for fast division either.

minad:
Ok!

PR #372, opened 2019-10-15
Rename and fixups

sjaeckel:
* Rename `mp_ilogb()` to `mp_log_u32()` as discussed in #366 
* fix the build on multiple points

### DIFF
--- a/bn_mp_log_u32.c
+++ b/bn_mp_log_u32.c
@@ -1,5 +1,5 @@
 #include "tommath_private.h"
-#ifdef BN_MP_ILOGB_C
+#ifdef BN_MP_LOG_U32_C
 /* LibTomMath, multiple-precision integer library -- Tom St Denis */
 /* SPDX-License-Identifier: Unlicense */
 
@@ -70,7 +70,7 @@ static mp_digit s_digit_ilogb(mp_digit base, mp_digit n)
          as is the output of mp_bitcount.
          With the same problem: max size is INT_MAX * MP_DIGIT not INT_MAX only!
 */
-mp_err mp_ilogb(const mp_int *a, uint32_t base, mp_int *c)
+mp_err mp_log_u32(const mp_int *a, uint32_t base, mp_int *c)
 {
    mp_err err;
    mp_ord cmp;
--- a/bn_mp_log_u32.c
+++ b/bn_mp_log_u32.c
@@ -70,11 +70,11 @@ static mp_digit s_digit_ilogb(mp_digit base, mp_digit n)
          as is the output of mp_bitcount.
          With the same problem: max size is INT_MAX * MP_DIGIT not INT_MAX only!
 */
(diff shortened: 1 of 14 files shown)


minad:
mp_expt_u32 is also not released right? So we could also name it mp_exp_u32 (like the c exp function). About the integer logarithm, I wonder why the output is a mp_int? The output should be a uint32_t too. In mp_exp_u32 the exponent is an uint32_t too, while the base is an mp_int. So we are not very consistent here. Maybe we should also have a log function with an mp_int base? But the exponent definitely doesn't have to be an mp_int.

sjaeckel:
> mp_expt_u32 is also not released right? So we could also name it mp_exp_u32 (like the c exp function).

I thought it'd make sense to keep its name as for some reason unkown to me, the entire cryptographic world talks about `exptmod` instead of `expmod` ...

> About the integer logarithm, I wonder why the output is a mp_int? The output should be a uint32_t too.

c.f. the comment above `mp_log_u32()`

> Maybe we should also have a log function with an mp_int base?

I'd really like if one were able to do `mp_log(mp_exp())` but I'm not sure if that's doable without all parameters being an `mp_int()` ...

sjaeckel:
> I thought it'd make sense to keep its name as for some reason unkown to me, the entire cryptographic world talks about `exptmod` instead of `expmod` ...

forget what I just said... it seems only the part of the cryptographic world that took the ltc/ltm code talks about exptmod (WolfSSL, MatrixSSL etc.)

minad:
Good. Could we change it for now such that mp_log returns an uint32_t (base and output should be uint32_t)? I think such a log function is very useful (eg radix size). Later on we can also add a full bigint version.

minad:
Hmm regarding 2expt and exptmod - Should we also rename them, or just keep mp_expt despite it being a little weird?

sjaeckel:
> Hmm regarding 2expt and exptmod - Should we also rename them, or just keep mp_expt despite it being a little weird?

Yes, either all or nothing IMO, which one - I don't care. @czurnieden your preference?

> Could we change it for now such that mp_log returns an uint32_t (base and output should be uint32_t)?

looks like that shouldn't be a problem as it's internally already working with `unsigned int`

minad:
Ok, I think we should keep mp_expt. Just renaming without any other changes doesn't bringt enough advantages. A little idiosyncracies of the library probably don't hurt.

I think only the log uint32 output needs fixing, but then I think we are good to go with 1.2

sjaeckel:
> I think only the log uint32 output needs fixing

@czurnieden fine by you as well?

minad:
@sjaeckel I pushed my version here. I hope you are fine with it. It makes sense to do it like this since all the internal computations are done in uint32_t. There is really no point in returning an mp_int. It is just an unnecessary allocation/boxing.

sjaeckel:
can you please simply remove the two commit&immediate revert

minad:
done

PR #373, opened 2019-10-15
add back missing symbols

minad:
@sjaeckel Added back the missing symbols. The issue is that the functions have not been compiled due to the invalid ifdef guard. Maybe we can improve the generate perl script somehow to check that the ifdef guard matches the function being defined. We also had the issue before where some invalid code was in bn_deprecated.c, which I fixed in some recent PR.


minad:
:+1:

PR #375, opened 2019-10-15
fix s_mp_montgomery_reduce_fast

### LINKED ISSUE #62: The incorrect range in fast_mp_montgomery_reduce
```
  /* now we have to propagate the carries and
   * shift the words downward [all those least
   * significant digits we zeroed].
   */
  {
    mp_digit *tmpx;
    mp_word *_W, *_W1;

    /* nox fix rest of carries */

    /* alias for current word */
    _W1 = W + ix;

    /* alias for next word, where the carry goes */
    _W = W + ++ix;

    for (; ix <= ((n->used * 2) + 1); ix++) {
      *_W++ += *_W1++ >> ((mp_word) DIGIT_BIT);
    }
```

I think, in this code should be used

```
    for (; ix < ((n->used * 2) + 1); ix++) {
```
1. The result from last cycle not used
2. In last cycle " *_W++ += " works with uninitialized element from " W "

---

In general, this part of code can be reduced / simplified

sjaeckel:
as proposed by @dmitry-lipetsk

This fixes #62

### DIFF
--- a/bn_s_mp_montgomery_reduce_fast.c
+++ b/bn_s_mp_montgomery_reduce_fast.c
@@ -119,7 +119,7 @@ mp_err s_mp_montgomery_reduce_fast(mp_int *x, const mp_int *n, mp_digit rho)
       /* alias for next word, where the carry goes */
       _W = W + ++ix;
 
-      for (; ix <= ((n->used * 2) + 1); ix++) {
+      for (; ix < ((n->used * 2) + 1); ix++) {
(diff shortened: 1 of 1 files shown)


minad:
:thinking:

sjaeckel:
@minad your thoughts on this patch?

minad:
I agree with @czurnieden but I think the code needs some simplification.

PR #378, opened 2019-10-16
Remove deprecated functions

minad:
Supersedes #327. This removes all the deprecated functions. The empty file is removed in a separate commit, since we might want to keep it (or revert it) in the future if new deprecations come up.


minad:
@sjaeckel Note that the more intrusive 8 bit removal is not done here. It is still in #349.

minad:
@nijtmans Thanks again! I saw those but then I forgot.

PR #379, opened 2019-10-16
enable MP_PRIVATE (visibility=hidden)

minad:
In 1.2 we had MP_PRIVATE disabled by default. This should be changed now.


nijtmans:
Well, I took the liberty to fix the mingw build of libtommath.dll to respect the privacy too: The GCC attributes don't work in this situation, instead the tommath.def file needs to be used to indicate which symbols are exported.

Apart from that: Looks good!

minad:
@nijtmans thanks!

PR #381, opened 2019-10-16
WIP: release v1.2.0

sjaeckel:
Changes on the release-branch

Only for review
Not to be merged (via GitHub Web UI ;-) )!

PR #382, opened 2019-10-17
Don't use the (undocumented) XCALLOC/MP_CALLOC any-more

nijtmans:
Since XCALLOC is undocumented, I don't think it's wise to require users with their own memory allocator to implement 4 replacement functions. I prefer to use MP_MALLOC followed by MP_ZERO_DIGITS, as everywhere else in the source-code.

Also, fix some documentation regarding the no longer existing functions.

Candidate for 1.2.0. Please consider.

### DIFF
--- a/bn_mp_init.c
+++ b/bn_mp_init.c
@@ -7,7 +7,8 @@
 mp_err mp_init(mp_int *a)
 {
    /* allocate memory required and clear it */
-   a->dp = (mp_digit *) MP_CALLOC((size_t)MP_PREC, sizeof(mp_digit));
+   a->dp = (mp_digit *) MP_MALLOC((size_t)MP_PREC * sizeof(mp_digit));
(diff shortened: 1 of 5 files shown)


minad:
I am against merging this. calloc is standard since calloc can be implemented more efficiently than malloc in some cases (Otherwise the question would be why it exists in the c standard after all).

@nijtmans What is the problem you are having? It is no issue to implement calloc for your custom allocator.

I propose instead to **document** MP_CALLOC.

sjaeckel:
c.f. #383

nijtmans:
At least the changelog should mention the move from XMALLOC and friends to MP_MALLOC and friends, since it's a change which libtommath embedders (like Tcl) should be aware of!

minad:
Agree, it should be added there. The changelog and the release are still not finalized as far as I am aware (see #381).

PR #384, opened 2019-10-18
Fix msvc build warnings

nijtmans:
Proposed solution for MSVC warning. It appears an additional pair of braces helps.

### DIFF
--- a/tommath_private.h
+++ b/tommath_private.h
@@ -142,8 +142,8 @@ extern void MP_FREE(void *mem, size_t size);
 
 /* feature detection macro */
 #define MP_STRINGIZE(x)  MP__STRINGIZE(x)
-#define MP__STRINGIZE(x) ""#x""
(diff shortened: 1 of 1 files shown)


nijtmans:
Apparently, this doesn't work :-(

minad:
Uh no. But adding a pragma to tommath_private.h should do it. I guess you don't include the private header anywhere in tcl? This would restrict the disabled warning to the relevant files only.

nijtmans:
I would welcome such a pragma in tommath_private.h. Yes, Tcl includes this, so that - indeed - should work too.

sjaeckel:
> If you have time to prepare a PR I would appreciate it. I guess @sjaeckel could still add it to 1.2.

@nijtmans feel free to open a PR based on release/1.2.0 or if "you're sure it works"^TM simply commit it to the release/1.2.0 branch ;-)

PR #388, opened 2019-10-19
Various fixes + split log_u32

minad:
Some small oversights

Edit: I splitted the log function for more configurability and better readability


minad:
There are some bn_* files reintroduced here which should be renamed.

sjaeckel:
I've seen it just now. Will do the rename.

Issue #363, opened 2019-10-08
use size_t where appropriate


minad:
In particular in mp_int. 2.0 refactoring!

dmitry-lipetsk:
https://github.com/dmitry-lipetsk/libtommath-for-ibprovider--set01/blob/master/ibp_external__libtommatch_set01__tommath.h

:)

minad:
@dmitry-lipetsk interesting! Are there issues/mistakes you found while doing the porting? You mention that in your readme. It would be nice to backport them here.

dmitry-lipetsk:
If I remember correcly, I described all detected issues here (in Issues).

minad:
Ok thank you!

czurnieden:
Are there plans to tackle the carefully ignored integer overflow in `mp_count_bits`? That problem is theoretical only for 32/64-bit large `size_t` but not so much for the LibCs for some 16-bit architectures with 16-bit `size_t` (e.g.: [AVR Libc](https://www.microchip.com/webdoc/AVRLibcReferenceManual/group__avr__stdint_1ga3c75bb398badb69c7577b21486f9963f.html) ) with enough memory.

Adding an overflow check in `mp_count_bits` is not a problem, signaling the error is. We could return `SIZE_MAX`  (as the error `MP_SOF` maybe?) which is safe because `(sizeof(size_t) * CHAR_BIT)/MP_DIGIT_BIT < SIZE_MAX` for all of the current `MP_nBIT` sizes.

E.g.: something like this:

```c
/* Useful elsewhere, too? */
#define MP_USED_MAX (SIZE_MAX/MP_DIGIT_BIT)

size_t mp_count_bits(const mp_int *a)
{
   size_t   r;
   mp_digit q;

   /* shortcut */
   if (MP_IS_ZERO(a)) {
      return 0;
   }

   /* get number of digits and add that */
   r = a->used;
   if (r > MP_USED_MAX) {
      return SIZE_MAX /*MP_SOF*/;
   }
   r = (r - 1) * MP_DIGIT_BIT;

   /* take the last digit and count the bits in it */
   q = a->dp[a->used - 1];
   while (q > 0u) {
      ++r;
      q >>= 1u;
   }
   return r;
}
```

But is it worth the hassle? I don't know.

nijtmans:
[...]

PR #391, opened 2019-10-19
Change type for result of mp_count_bits to size_t

czurnieden:
Change `int mp_count_bits(mp_int *a)` to `mp_err mp_count_bits(mp_int *a, size_t *size)`.

It is only a start (`demo/*c` not even touched for now) because changing `mp_count_bits` leads to a couple of changes in other functions. See comments in sources.

### DIFF
--- a/bn_mp_count_bits.c
+++ b/bn_mp_count_bits.c
@@ -17,7 +17,8 @@ mp_err mp_count_bits(const mp_int *a, size_t *size)
    }
 
    /* get number of digits and add that */
-   r = a->used;
+   /* TODO: cast not necessary once the mp_int struct gets its "int" types changed to "size_t" */
+   r = (size_t)a->used;
    if (r > (SIZE_MAX/MP_DIGIT_BIT)) {
       return MP_BUF;
    }
--- a/bn_mp_count_bits.c
+++ b/bn_mp_count_bits.c
@@ -4,18 +4,24 @@
 /* SPDX-License-Identifier: Unlicense */
 
 /* returns the number of bits in an int */
-int mp_count_bits(const mp_int *a)
+mp_err mp_count_bits(const mp_int *a, size_t *size)
 {
-   int     r;
+   size_t   r;
    mp_digit q;
+
+   *size = 0;
 
    /* shortcut */
    if (MP_IS_ZERO(a)) {
-      return 0;
+      return MP_OKAY;
    }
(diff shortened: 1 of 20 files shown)


minad:
Can you please delay working on this until we have the open PRs merged? You also have multiple open ones with unclear status.

Let's please discuss first on how we proceed with such a large scale refactoring since it touches many functions and will invalidate all PRs.

nijtmans:
Sorry, but I don't like this. If mp_count_bits() returns an error, it means a situation that mp_count_bits cannot handle because its result doesn't have enough bits to represent it. So, the correct solutions is to increase this number of bits, so change size_t to unsigned long or unsigned long long. This way, the callers of mp_count_bits don't have to handle the error-situation, they just have to be modified to handle enough bits for the count. Just my 2c....

minad:
I have to think about this more. I would also like the type should be sufficient to represent the number of bits. But using uint64_t on 64 bit platforms is also a bit crazy. I would like to have it such that size_t is large enough for the bits, checked before malloc. Unfortunately this means that only 8kib numbers can be represented on 64k platforms which might not be good enough?

minad:
We could also define our own bits_t type which is always large enough. size_t or uint32_t on 16 bit platforms.

minad:
@czurnieden can we close this for now. More discussion is needed before doing anything.

czurnieden:
> Can we close this for now? More discussion is needed before doing anything.

As a discussion is already starting we can discuss it here. Or do you want to do it more formally with an "Issue"? 

And to be honest, this PR was meant to start a discussion. Changing the type will be a great pain in the behind, that rabbithole is really deep and I have chosen this example because it nicely shows some of the complications we can expect if we replace all of the `int`s with a single native data type of variable size (here: `size_t`).

And yes, the check should be done at the (re)allocation stage(s). It is the right place for it and they already return errors so `mp_count_bits` can keep its comfortable API and just return the size.

And now back to rebasing because somebody really got to town this weekend with PR-merging! ;-)

minad:
Let's do it in the issue #363 please. I think we are both well aware of the complications. There is also the issue with signed loop variables vs unsigned loop variables. Furthermore since many things interact here it will be hard to split the work (you also have many TODOs due to that).

But still it should be better to make a nice plan. Maybe we can identify a few things that can be done in a step by step basis.

PR #392, opened 2019-10-19
Corrected type for the sign-handling in bit-banger div

czurnieden:
Attended the undeservedly neglected bit-banger division.


minad:
Could you add this to #370 instead?

czurnieden:
> Could you add this to #370 instead?

At one time you insist on sliceing the PR and here you insist on combining them.
You're killing me.
\*sigh\*

No, just kidding, I've put it already in #370 this morning, this PR was an excerpt because I thought you want those PRs individually.

Issue #395, opened 2019-10-21
check the manual, function prototypes in sync, deprecated functions


minad:
Check that everything is in sync.

PR #397, opened 2019-10-22
mp_prime_next_prime: use mp_bool for bbs_style

minad:
found this while looking over the api due to #363


sjaeckel:
> Has it been changed in the documantation, too?

I'd say we can ref to #395 ;)

PR #398, opened 2019-10-22
Use unsigned long for bitsizes

nijtmans:
See #363, the mp_bitcount part only.


nijtmans:
Thinking about it, using size_t is likely better than unsigned int. So, closing.

PR #401, opened 2019-10-22
add s_mp_log10

minad:
@czurnieden Regarding our discussion in #371, this is somehow what I would like to have. But the test is failing right now, maybe some rounding issue?.

### DIFF
--- a/demo/test.c
+++ b/demo/test.c
@@ -1524,8 +1524,8 @@ LBL_ERR:
    return EXIT_SUCCESS;
 #   endif /* LTM_DEMO_TEST_REDUCE_2K_L */
 }
-/* stripped down version of mp_radix_size. The faster version can be off by up t
-o +3  */
+
+/* stripped down version of mp_radix_size. The faster version can be off by up to +3  */
 /* TODO: This function should be removed, replaced by mp_radix_size, mp_radix_size_overestimate in 2.0 */
 static mp_err s_rs(const mp_int *a, int radix, uint32_t *size)
 {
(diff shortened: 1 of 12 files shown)


czurnieden:
> maybe some rounding issue?

Yes, but from an unexpected direction.
I was so concentrated on getting from log_2(x) to log_10(x) that I forgot to consider the rounding error in  floor(log_2(x)) and now it's off by one and back where I started.

How preposterous of me to think that I'm better than the people at GMP! ;-)

There are some tricks available to get some more bits of information out of `mp_count_bits` but they all have quite some overhead:

e.g.:  `log(x^n)/n = log(x)` (all numbers in decimal)
With one of the failures `1086654250330508058` and `n = 2` we get `floor(log_2(1086654250330508058)) = 119` round it up to `120` and divide by 2 to get `60` and feed that into the original machine to get the correctly rounded result. It is obvious that smaller numbers than that one need larger `n` and larger numbers get even larger when you square them.
Works quite well with floats but not with integers.

The other way is too take a couple of the most significant bits into consideration but I'm on my way already, needs to be put on hold until this evening.

minad:
No hurry! But I guess it is good to have it put into mp_log after all otherwise we wouldn't have caught this bug ;)

minad:
@czurnieden you can take over this PR if you want, it is your code after all.

czurnieden:
> @czurnieden you can take over this PR if you want, it is your code after all.

In that case let me take over to show some mercy and put it down.
\*sigh\*

PR #405, opened 2019-10-23
stdbool/stdint refactoring, various optimizations/simplifications

minad:
This PR includes the following changes:

* Optimization: Added mutating fast path to mp_add_d/mp_sub_d (this was previously part of mp_incr/mp_decr). The optimization is more general than mp_incr/mp_decr. Replace mp_incr/mp_decr by macros.
* Simplification: s_mp_reverse is only used by mp_to_radix, make it static
* Cleanup: prime_rand internals simplified, remnants of removed deprecations
* Simplification: stdbool introduced instead of mp_bool. Remove MP_IS_* macros and explicit comparisons for booleans. 
* mp_iseven/mp_isodd are defined as macros again as they have been before 1.2. This reverts the change where @fperrad introduced MP_IS_EVEN/MP_IS_ODD macros. These macros have been introduced to hide the explicit boolean comparisons, which have been necessary to make the linter happy. Since we have stdbool, we can go back to the previous state.
* Introduce make c89 and make c99 to convert back and forth between source versions which use stdbool/stdint and which don't. This makes the no-stdint-h branch obsolete. The converted source is tested in travis-ci and the c89/c99 roundtrip is also tested. This way we make sure that everything works as it should in both versions of the source. In develop we keep the "simpler source", which just uses the standard types. The c89 converted source uses custom types mp_bool, mp_u32 etc.


tommath.h
```diff
@@ -209,9 +209,9 @@ mp_err mp_init_size(mp_int *a, int size) MP_WUR;
 
 /* ---> Basic Manipulations <--- */
 #define mp_iszero(a) (((a)->used == 0) ? MP_YES : MP_NO)
-mp_bool mp_iseven(const mp_int *a) MP_WUR;
-mp_bool mp_isodd(const mp_int *a) MP_WUR;
 #define mp_isneg(a)  (((a)->sign != MP_ZPOS) ? MP_YES : MP_NO)
+#define mp_iseven(a) (((a)->used == 0) || (((a)->dp[0] & 1u) == 0u) ? MP_YES : MP_NO)
```
sjaeckel: that was a short period of time that they were functions... introduced for 1.2.0 and abandoned one day after 1.2.0 was out :rofl: see #142

minad: To be honest, I never was a fan of those and also not of the MP_IS_ZERO macros.
They exist only to make the fperrard linter happy but make the code worse. I mean this x?MP_YES:MP_NO pattern. I would use c99 stdbool.h instead.

czurnieden:
> What do you think about this? 

I had the impression that LTM  works on copy. (even when it makes not *that* much sense like in `mp_neg`) and any exception of this principle is made explicit.
The `const` in `mp_add_d(const mp_int *a, mp_digit b, mp_int *c)` is not only for the compiler, the unsuspecting programer might also think that the input doesn't get touched at all.

minad:
You mean that a, b doesn't alias c? Yes, I also thought that first, bit this is not the case. All functions may alias. Therefore mp_incr/decr are not very idiomatic. I think replacing them my this PR would be better. I didn't see this before when mp_incr was merged.

@czurnieden are you ok with this? This PR should yield the same benefits performance wise as mp_incr/decr, but more general.

tommath.h
sjaeckel: > I would use c99 stdbool.h instead

maybe it's time? We could start an issue on that topic!?
@nijtmans needs c89 as TCL needs to be compiled in VC6 for some ancient OS IIUC!?
@karel-m also had some reasons to have c89 code.
For how many more years have these legacy systems to be supported?

Just to make it clear: I could also live without this explicit compare dance, especially for `mp_bool`. How about switching to stdbool for c99 compilers and writing (and linting) the code against c99? But for now also still provide a fallback for c89. I'm aware that this is not 100% possible/compatible, but wouldn't it be an option!?

sjaeckel:
Now that you're touching stuff like that, wouldn't it then make sense to have a path `mp_add->mp_add_d` for `b->used == 1`?

tommath.h
minad: Yes, this is what I am arguing all the time. It is no problem to provide stdbool.h for c89 too on vc6. But we could also just change the linting and avoid explicit compare for mp_bool. Then I would be happy, even if we don't use stdbool.

minad:
Yes, I can add the mp_add_d path to mp_add

tommath.h
sjaeckel: > It is no problem to provide stdbool.h for c89

including this? ;-) 

```
mp_bool a = 1, b = 2;
if (a == b) printf("stdbool");
else printf("error");
```

> But we could also just change the linting rules and avoid explicit compare for mp_bool

@fperrad would that be acceptable?

czurnieden:
> All functions may alias.

They tend to work on copy (didn't check all!) and say so implicitly (by not having a `const` qualifier for the input) and explicitly (in the documentation) if they don't. That means that the pointer to the input is different from the pointer to the output.
Can it cause problems if that behaviour is changed?

minad:
No, this is not true. Pointer aliasing hast nothing to do with constness. The functions are such that they allow aliased pointers. Functions which don't support aliased pointers are broken.

minad:
@czurnieden btw, you call mp_add_d in incr.

czurnieden:
> Pointer aliasing hast nothing to do with constness.

I don't mean the technical side of it when I talk about `const` just the expectations of the user when they see it.

Oh, and because I think that your reaction is based on a misunderstanding: I'm not against it in a way that I want you to change any code in that regard,  but I think that the new behaviour is different enough  that it justifies a short note in the documentation. Just because we cannot come up with a problem (at least none that doesn't start with: a crappy parallelizing compiler might&hellip; ;-) ) doesn't mean that none exists.

> btw, you call mp_add_d in incr.

Yes. I was a bit torn between returning an error, calling `add_d` and doing the addition internally but found the call best: it makes the function useful in a general manner and avoids copying code for the prize of an extra copy which is very rare (needs to have the zeroth limp full which is quite unlikely now, without `MP_8BIT`).

mp_add_d.c
```diff
@@ -10,6 +10,15 @@ mp_err mp_add_d(const mp_int *a, mp_digit b, mp_int *c)
    int ix, oldused;
    mp_digit *tmpa, *tmpc;
 
+   /* fast path for a == c */
+   if (a == c &&
+       !mp_iszero(c) &&
+       c->sign == MP_ZPOS &&
+       c->dp[0] + b < MP_DIGIT_MAX) {
```
czurnieden: I think parentheses are mandatory now.

i would move the `mp_iszero` check into the block and add something like e.g.: `c->used += mp_iszero(c)` but that's just me.

minad:
Ok. As I understand, you agree with the code change then?

However you also say that the behavior is different now. No - it is not. It is simply an optimization of the fast path. The function does what it did before, just faster in the mutating case.

Furthermore you suspect that compilers could have problems with this code. In the code at hand the compiler must assume that a->dp and c->dp can point to the same region of memory. In some cases the aliasing analysis of the compiler can prove that such an aliasing doesn't happen and optimize accordingly. This is not the case here. There is also the restrict keyword to tell the compiler that aliasing won't happen.

Concerning the docs - I agree we should add a note that the functions work with input==output. I think all do, but I have checked only a few quite a while ago. We should also check that input==output works everywhere.

czurnieden:
> Ok. As I understand, you agree with the code change then?

Yes.
If I would have been against it I would have made it official in the "review" like I did with the parentheses.

> We should also check that input==output works everywhere.

I also think all do but: better safe than sorry and all that, y'know.

mp_add_d.c
minad: I reworked it a bit and added the parens. However I didn't to this c->used += mp_iszero(c) thing and the more complicated equivalent for subtraction to avoid adding too much code. I mean it is the fast path, it is sufficient if it handles the common cases.

minad:
@sjaeckel I switched to stdbool, see the commit message for the reasoning and @fperrad's comment in #411.

minad:
@nijtmans I looked how you work around stdint. Using stdbool wouldn't be as simple to work around. Maybe we should use the hack described by @fperrad in tommath.

~~~ c
typedefs bool mp_bool;
#define MP_YES true
#define MP_NO false
~~~

minad:
@nijtmans I added the command "make c89", which applies the necessary substitutions. The includes stdint.h and stdbool.h are removed and custom types are introduced. Please check if that works for you!

nijtmans:
Please don't remove the mp_iseven()/mp_isodd() functions: macro's externally visible should not access the digits directly, it's not portable.  I would suggest to keep mp_isodd(), and make mp_iseven() a macro equal to (!mp_isodd()).    Then, in tommath_private.h, you can re-define mp_isodd as a macro for internal use (adding an #undef in mp_iseven.c) so everything works again.   If you want, I can provide a separate pull-request doing that.

minad:
These were macros originally, and turned to functions only because we had issues with the linting by @fperrad. It is basically reverting an earlier mistake. The mp_iseven/isodd are called internally and things should inline.

I understand that you want to treat mp_int as an opaque type which is useful if you compile the library with a different compiler than the library consumer. But this is already a problematic approach. If you do that, I recommend that you patch the library and define mp_int as opaque "typedef struct mp_int_s mp_int;". Furthermore add mp_iseven and isodd as functions (after undefining iseven/isodd) such that they appear in the abi.

minad:
@nijtmans I am thinking about it. But this is the same issue with mp_get_ul and mp_get_l where you try to work around things by a hack. I am not happy with that. If you want to forbid accessing mp_int you should make the type opaque.

Is it really that of a big problem to use the unlatched library in tcl? The issue is the missing int128, right? And the mp_int layout is also not portable, right?

minad:
Yes, but I don't agree with you "portability argument". Since we are exposing mp_imt, we can access it. You try to shape the library in a bad way to enable your special use case. Is there no better way to achieve compatibility with vc2008.

minad:
@nijtmans I will experiment a bit and see if I can come up with a solution.

minad:
@nijtmans I thought about introducing an opaque type, which could be enabled selectively. But then I wonder about the mp_zero macro - why is this compatible if the struct layout is not?

minad:
Or is only the dp array not portable since your header doesn't agree with the compiled library? How is this possible? MSVC knows int64, so this should work?

minad:
@sjaeckel could you please take a look at this? I think the no-stdint issue is reasonably solved here, such that we won't get issues in the future. Furthermore this includes some other refactorings.

sjaeckel:
> could you please take a look at this?

I did and I like what I see! :-) 

would you mind splitting this into separate PR's? if you mind, can you please at least update the title and description? ;-) 

nice solution for the stdint/bool problem btw! if @nijtmans agrees this conversion could even be added to the 1.x branch (if you think this should go in 1.x please split the PR!)

minad:
@sjaeckel I updated the description. I would prefer if you merge it without previous splitting since I would have to untangle the mp_bool conversion from the rest. While this was a pretty mechanical change, I didn't do it with a script but by applying a series of substitutions and manual intervention.

Regarding 1.2.x/1.3 backports - I think it would be possible to backport the c89/c99 conversion for sure. However I would recommend to **not** do it and only backport urgent bugfixes if there are any. I believe we should push for projects to upgrade to 2.0 in reasonable time. At first I thought that it might make sense to do some incompatible code updates in 2.0 (moving to c99, see my thoughts in #416), but I believe now that this isn't a good idea. I think we should restrict c89+stdbool+stdint, since these headers are harmless. This means there won't be any unnecessary obstacles for projects to do the update. There will be some issue due to replaced/deprecated APIs and the probably upcoming change regarding size_t (see #363), but this is all. We didn't remove anything without a substitute, so I think projects can make the move quickly.

sjaeckel:
then there's going to be the no-stdint tarball only from 2.0 on - also a good incentive for projects to move to 2.0 :) 

I'll merge as soon as it's green

PR #407, opened 2019-10-23
CYGWIN does not suport visibility attribute

nijtmans:
trivial ....

### DIFF
--- a/s_mp_log_pow2.c
+++ b/s_mp_log_pow2.c
@@ -1,0 +1,12 @@
+#include "tommath_private.h"
+#ifdef S_MP_LOG_POW2_C
+/* LibTomMath, multiple-precision integer library -- Tom St Denis */
+/* SPDX-License-Identifier: Unlicense */
+
+uint32_t s_mp_log_pow2(const mp_int *a, uint32_t base)
+{
+   int y;
+   for (y = 0; (base & 1u) == 0u; y++, base >>= 1) {}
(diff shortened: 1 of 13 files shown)


nijtmans:
> @nijtmans could you check all _Win32 ifdefs if we also have cygwin there?

Yes, I checked all WIN32 ifdefs, all are OK (Cygwin is more like UNIX than win32, actually)

However, I would suggest to drop support for everything older than Windows XP. Tcl already dropped Windows CE/95/98/ME/2000 too. So, I added a commit doing that here (while we're on it, although this has no relation to Cygwin ....)

minad:
This still needs a rebase. It seems to touch unrelated code?

s_mp_log_pow2.c
```diff
@@ -0,0 +1,12 @@
+#include "tommath_private.h"
+#ifdef S_MP_LOG_POW2_C
+/* LibTomMath, multiple-precision integer library -- Tom St Denis */
+/* SPDX-License-Identifier: Unlicense */
+
+uint32_t s_mp_log_pow2(const mp_int *a, uint32_t base)
+{
+   int y;
+   for (y = 0; (base & 1u) == 0u; y++, base >>= 1) {}
+   return (uint32_t)((mp_count_bits(a) - 1) / y);
```
minad: Why is this here?

nijtmans: I don't know. Something went wrong with the rebase ....

PR #408, opened 2019-10-23
Fix comment in helper.pl

nijtmans:


minad:
@sjaeckel there are these merge when passing labels. Did you add them to the list or is this a new ui feature?

minad:
I think you have to approve to trigger the merging :)

sjaeckel:
> I think you have to approve to trigger the merging :)

that's what I want to test now if it's sufficient if there's one approval

> there are these merge when passing labels. Did you add them to the list or is this a new ui feature?

I added https://reporanger.com/ to try it out

minad:
Ah ok. The point is - I am not allowed to merge but I can add this label. So I am not sure what it does.

sjaeckel:
> Ah ok. The point is - I am not allowed to merge but I can add this label. So I am not sure what it does.

that's exactly what I wanted to try :-D 

I've used GitLab in the last months and as CI is baked in there the "merge when passing" option is there by default, so I thought it'd be useful to have this

...then there's another label missing "rebase and merge when passing" ;-)

sjaeckel:
@minad can you set the label again please?

minad:
Hmm, I checked reporanger. And tbh it doesn't seem too useful to me. For example this auto close feature - if I would report sth and then some bot autocloses the issue, I would for sure never again interact with this project. Second - the automerging feature sounds nice, but it means that you are giving access to an external (trusted?) entity. I am not a fan of that. Maybe github has very fine grained privileges, which only give very restricted access such that no abuse is possible. What I would do - I would rather give additional commit rights to "someone trusted" from the project, e.g., @czurnieden or me. But then - it might also be good to have only you as the final person doing the actual merge since this ensures a final review leading to better commits in the end.

minad:
CI failed. Travis seems unreliable and slow lately.

sjaeckel:
> For example this auto close feature - if I would report sth and then some bot autocloses the issue, I would for sure never again interact with this project.

very good point, I'd avoid such a project as well, but I also wouldn't use a feature like that.

> I am not a fan of that

me neither! I would have preferred to have this baked into GitHub, but hey it's all Actions, Services, Rainbows and Unicorns nowadays...

> Maybe github has very fine grained privileges, which only give very restricted access such that no abuse is possible.

I couldn't find it...

> What I would do - I would rather give additional commit rights to "someone trusted" from the project, e.g., @czurnieden or me.

I think @czurnieden has already those rights, but rarely uses them

> But then - it might also be good to have only you as the final person doing the actual merge since this ensures a final review leading to better commits in the end.

Maybe you're right there :-) 

Still it's cumbersome sometimes and I'd prefer to review stuff and then mark it as "merge when passing", do the rebase and go away ... nowadays GitHub even detects when a PR has changed and shows you that in the diffview UI but it still leaves your approval...

> CI failed. Travis seems unreliable and slow lately.

yep... already since some days...

czurnieden:
> I think @czurnieden has already those rights, but rarely uses them

Yes, I do, but I, too, think that the final commits should be done by a single hand at a time.

It would have been really nice if the flag "merge when passing" exists without any connected automation beside a notification. But, @sjaeckel , you can still ping me and I can (rebase and) merge in the night after passing the tests. Didn't make much sense in the past but might now since Travis is a bit slow these days.

I can take the documentation under my wings if nobody complains (Ha, as if! ;-) )?

And I can take over some of the chores (small things only like e.g.: linting) when you are offline for more than a week or so, such that you have a smaller pile of PRs when you come back.

Issue #406, opened 2019-10-23
Deprecated function still listed in the user manual


justinethier:
The user manual for 1.2.0 / master lists `mp_expt_d` even though that function is now deprecated in favor of `mp_expt_u32`.

minad:
Thanks for reporting! We will revise the manual soon since it got out of sync with the development version. See #395.

You seem to use ltm, in cyclone, your scheme implementation. It would be interesting where the library is used right now.

* Cyclone
* Tcl
* Moarvm
* Tomcrypt
* ...

sjaeckel:
I've added https://github.com/libtom/libtommath/wiki/Projects-using-libtommath

minad:
subissue of #395

PR #410, opened 2019-10-24
Cleanup and update of manual

czurnieden:
Regarding e.g.: #406 

It is a lot, please check if nothing was forgotten, is wrong, or can go, thank you!

### DIFF
--- a/doc/.latexindent.yaml
+++ b/doc/.latexindent.yaml
@@ -1,0 +1,1 @@
+defaultIndent: "   "
--- a/doc/.latexindent.yaml
+++ b/doc/.latexindent.yaml
@@ -1,1 +1,35 @@
-defaultIndent: "   "
+# 2 spaces
+defaultIndent: "  "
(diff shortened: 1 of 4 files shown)


minad:
@sjaeckel This is also good to merge I think!

sjaeckel:
now that the first build job also checks the doc it will stay up-to-date I hope ;-)

sjaeckel:
AFAIU this version of the docs also applies to 1.2.0, right!? so it'll make sense to use these changes (minus `MP_8BIT` removal) there as well!?

czurnieden:
> AFAIU this version of the docs also applies to 1.2.0, right!?

I finished it before the last commit feast, so I think it should, yes.

> so it'll make sense to use these changes (minus MP_8BIT removal) there as well!?

An official release should have an up-to-date documentation, yes.
(Is the 8-bit support even mentioned in the docs?)

PR #412, opened 2019-10-25
fix win32 includes, remove support for everything older than Windows XP

nijtmans:
The same as #407, but then everything in a single commit.

This removes support for Windows 95/98/ME/2000/CE, no-one cares about any more.

Also add the ".settings" directory (for Eclipse) to .gitignore.

### DIFF
--- a/tommath_private.h
+++ b/tommath_private.h
@@ -16,7 +16,7 @@
  * as a shared object. By default, symbols are visible.
  * On Win32 a .def file must be used to specify the exported symbols.
  */
-#if defined(__GNUC__) && __GNUC__ >= 4 && !defined(_WIN32)
+#if defined(__GNUC__) && __GNUC__ >= 4 && !defined(_WIN32) && !defined(__CYGWIN__)
 #   define MP_PRIVATE __attribute__ ((visibility ("hidden")))
 #else
 #   define MP_PRIVATE
--- a/s_mp_rand_platform.c
+++ b/s_mp_rand_platform.c
@@ -16,15 +16,14 @@ static mp_err s_read_arc4random(void *p, size_t n)
 }
 #endif
 
-#if defined(_WIN32) || defined(_WIN32_WCE)
+#if defined(_WIN32)
 #define S_READ_WINCSP_C
 
 #ifndef _WIN32_WINNT
-#define _WIN32_WINNT 0x0400
+#define _WIN32_WINNT 0x0501
 #endif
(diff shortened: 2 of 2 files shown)


sjaeckel:
> The same as #407, but then everything in a single commit.

FYI you could also rebase, squash and force-push the branch of the old PR

tommath_private.h
```diff
@@ -16,7 +16,7 @@
  * as a shared object. By default, symbols are visible.
  * On Win32 a .def file must be used to specify the exported symbols.
  */
-#if defined(__GNUC__) && __GNUC__ >= 4 && !defined(_WIN32)
+#if defined(__GNUC__) && __GNUC__ >= 4 && !defined(_WIN32) && !defined(__CYGWIN__)
```
sjaeckel: is this patch something that should also be in 1.2? if yes, can you please make it a separate patch so I can cherry-pick it for an eventual 1.2.1

.gitignore
```diff
@@ -29,6 +29,7 @@ mtest_opponent.exe
 # ignore eclipse project files
 .cproject
 .project
+.settings
```
sjaeckel: this also please as a separate commit

tommath_private.h
nijtmans: That wouldn't work anyway, because on 1.2.0 the #ifdef is different:
    #if defined (MP_PRIVATE_SYMBOLS) && defined(__GNUC__) && __GNUC__ >= 4
    #   define MP_PRIVATE __attribute__ ((visibility ("hidden")))
    #else
    #   define MP_PRIVATE
    #endif

.gitignore
nijtmans: > this also please as a separate commit

Why? I can make a separate commit for 1.2.1 anyway, together with the MP_PRIVATE change above.

sjaeckel: Because there's `cherry-pick -x` to do something like that, not a new commit.

tommath_private.h
sjaeckel: OK, so in 1.2.0 it should be `#if defined (MP_PRIVATE_SYMBOLS) && defined(__GNUC__) && __GNUC__ >= 4 && !defined(_WIN32) && !defined(__CYGWIN__)` right?

.gitignore
sjaeckel: Simply leave it as is, I'll go for the separate commit

PR #417, opened 2019-10-26
Make mp_iseven a macro

nijtmans:
Separate from the discussion what should be done with mp_isodd(), I would like to make mp_iseven() a macro negating mp_isodd()


minad:
I agree with this. But mp_isodd should be a macro too. I don't understand the portability issue. The hack you proposed with defining/undefining mp_isodd internally is not a good idea since I am trying to get rid of hacks, not introduce new ones :(

minad:
I added this to #405.

PR #424, opened 2019-10-27
functions guarded by MP_HAS are treated as optional dependencies

minad:
for example s_mp_div_recursive is an optional dependency of mp_div


minad:
discussed sth like this in #301

PR #425, opened 2019-10-27
replace gen.pl by cat, rename mpi.c to mp_all.c

minad:
gen.pl is an old artifact which can be easily replaced by standard unix cat. we already did some consolidation of perl scripts in #293.


makefile
```diff
@@ -71,13 +71,13 @@ profiled:
 
 #make a single object profiled library
 profiled_single:
-	perl gen.pl
-	$(CC) $(LTM_CFLAGS) -fprofile-arcs -DTESTING -c mpi.c -o mpi.o
-	$(CC) $(LTM_CFLAGS) -DTESTING -DTIMER demo/timing.c mpi.o -lgcov -o timing
+	cat *mp_*.c > mp_all.c
```
sjaeckel: how about

```
diff --git a/makefile b/makefile
index 7889c5a..dab28c6 100644
--- a/makefile
+++ b/makefile
@@ -72,5 +72,4 @@ profiled:
 #make a single object profiled library
-profiled_single:
-       cat *mp_*.c > mp_all.c
-       $(CC) $(LTM_CFLAGS) -fprofile-arcs -DTESTING -c mp_all.c -o mp_all.o
+profiled_single: pre_gen
+       $(CC) $(LTM_CFLAGS) -fprofile-arcs -DTESTING -c pre_gen/mp_all.c -o mp_all.o
        $(CC) $(LTM_CFLAGS) -DTESTING -DTIMER demo/timing.c mp_all.o -lgcov -o timing
@@ -78,3 +77,3 @@ profiled_single:
        rm -f *.o timing
-       $(CC) $(LTM_CFLAGS) -fbranch-probabilities -DTESTING -c mp_all.c -o mp_all.o
+       $(CC) $(LTM_CFLAGS) -fbranch-probabilities -DTESTING -c pre_gen/mp_all.c -o mp_all.o
        $(AR) $(ARFLAGS) $(LIBNAME) mp_all.o
```

minad: Yes, or call it tommath_generated.c instead of mp_all.c. this is not a good name due to the helper.pl wildcards.

sjaeckel: I'll add it to the PR

minad: Why do we need the pre_gen directory? Maybe this can be simplified. I take a look later.

sjaeckel: I already thought if it wouldn't make a bit more sense to have a folder structure with all the source files as in tfm

minad: Yes for larger projects I do that usually. But in this case I think it is nice to have the source in a flat hierarchy to browse it directly in github. The utilities and tests should be separate however. In particular if we have a relatively clean repository and rename demo->test, etc->util, remove logs/ etc. It is also a kind of accessibility statement - that the library has a simple structure. It is not a deeply nested project with subsystems and bells and whistles.

makefile
```diff
@@ -118,12 +117,17 @@ coveralls: lcov
...
 	$(MAKE) -C doc/ $@ V=$(V)
 
-.PHONY: pre_gen
+.PHONY: pre_gen cmp
 pre_gen:
 	mkdir -p pre_gen
-	perl gen.pl
-	sed -e 's/[[:blank:]]*$$//' mpi.c > pre_gen/mpi.c
-	rm mpi.c
+	cat *mp_*.c > mp_all.c
```
minad: I would get rid of the pre_gen directory since it seems there will be only mp_all.c inside it.  This sed call seems also unnecessary? Only mp_all.c should be called differently such that helper.pl doesn't find it.

minad: Alternatively generate mp_all.c directly in pre_gen and not in ./

I sometimes had the issues with this stale file lying around and others had too I think.

sjaeckel: `ltm_amalgamated.c` fine?

minad: I would call it tommath_amalgam.c. This fits more with the rest.

sjaeckel:
This can be merged?!

sjaeckel:
btw. can anyone confirm that multiplication and exptmod are slower in the meant to be optimized/profiled build (the one called "after")?

sjaeckel:
btw. I just realized that the "before-after" graphics allow now an easy comparison of performance differences of branches

```sh
$ git checkout develop
$ make -j9 timing CFLAGS='"-DMP_VERSION=\"before\""' && ./timing
$ git checkout fature-branch
$ make -j9 timing CFLAGS='"-DMP_VERSION=\"after\""' && ./timing
$ make -C logs/ cmp
$ xdg-open logs/before_after.html
```

PR #431, opened 2019-10-29
clean-up test.c

sjaeckel:
* no more `MP_WUR` in test.c
* clean-up console output

### DIFF
--- a/tommath.h
+++ b/tommath.h
@@ -204,7 +204,7 @@ mp_err mp_init_size(mp_int *a, int size) MP_WUR;
 
 /* ---> Basic Manipulations <--- */
 #define mp_iszero(a) ((a)->used == 0)
-#define mp_isneg(a)  ((a)->sign != MP_ZPOS)
+#define mp_isneg(a)  ((a)->sign == MP_NEG)
 #define mp_iseven(a) (((a)->used == 0) || (((a)->dp[0] & 1u) == 0u))
 #define mp_isodd(a)  (!mp_iseven(a))
 
--- a/demo/mtest_opponent.c
+++ b/demo/mtest_opponent.c
@@ -1,3 +1,4 @@
+#define MP_WUR /* TODO: result checks disabled for now, T.b.d. if it should even be done here */
 #include "shared.h"
 
 #ifdef LTM_MTEST_REAL_RAND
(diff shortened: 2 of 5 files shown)


minad:
No wur! Great!

tommath.h
```diff
@@ -204,7 +204,7 @@ mp_err mp_init_size(mp_int *a, int size) MP_WUR;
 
 /* ---> Basic Manipulations <--- */
 #define mp_iszero(a) ((a)->used == 0)
-#define mp_isneg(a)  ((a)->sign != MP_ZPOS)
+#define mp_isneg(a)  ((a)->sign == MP_NEG)
```
sjaeckel: oh, nice change sneaked in this PR ;-)

nijtmans: Hm... This is wrong:  What if (a)->sign = 2 ???   This sort of comparsions should always be against the value 0, in this case MP_ZPOS.   Well spotted, @sjaeckel !!

sjaeckel: Which brings me immediately to: how about changing `mp_sign sign` to `bool neg`? this would (at least for c-std >= 99) solve the described problem

nijtmans: Well, a lot of external code (read: Tcl ...) uses the "sign" field directly. If you change the field name, then a macro mp_setneg(a, bool) should be introduced (in libtommath 1.2.0 it will be mp_setneg(a, mp_sign))  ...

sjaeckel: this was only loud thinking :) and we won't be able to have this in 1.x

minad: While I am for using bool in generic situations, I think mp_sign is good to be separate. This will prevent type confusion in some cases. At least give a warning.

nijtmans: Realistic:  How about a new macro mp_setneg(), so you can use a boolean to set the sign:

#define mp_setneg(a, b)  ((a)->sign = (b ? MP_NEG : MP_ZPOS))

sjaeckel: there's already `mp_neg`!?

minad: @nijtmans Proposes something else (just a setter for the field), but I don't think we shouldn't change anything here. I don't want to have a setter, this is not C++. Let's just keep it as is, i.e. `a->sign = MP_NEG`.

PR #432, opened 2019-10-29
introduce MP_NO_DEPRECATED_PRAGMA

sjaeckel:
IMO it's no problem if a library user decides to keep on using the old API.

To be able to do so the user can add `-Wno-deprecated-declarations` for the build, but there was no way to disable the deprecated pragma's.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,7 @@ branches:
     - master
     - develop
     - /^release/
+    - /^support/
     - /^travis/
 
 # Additional installs: Valgrind for memory tests.
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,6 +4,7 @@ branches:
   - master
   - develop
   - /^release/
+  - /^support/
   - /^travis/
 image:
 - Visual Studio 2019
--- a/demo/timing.c
+++ b/demo/timing.c
@@ -4,7 +4,6 @@
 #include <unistd.h>
 #include <inttypes.h>
 
-#define MP_WUR
 #include <tommath.h>
 
 #ifdef IOWNANATHLON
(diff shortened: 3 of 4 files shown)


minad:
Why do we need this? The compiler already provided a way to disable warnings. Also selectively via warning push/pop.

sjaeckel:
> Why do we need this? The compiler already provided a way to disable warnings. Also selectively via warning push/pop.

How can I then disable the warning generated by `MP_DEPRECATED_PRAGMA` without disabling all/other warnings as well? That'd also be an option for me.

minad:
You can disable them by pragmas blockwise. I know that providing this will make it easier. But I am generally against providing means to disable warnings. They are there for a reason and it is good that they are annoying. The same holds for MP_WUR. Maybe we shod remove the possibility to disable it in 2.0. The user can always choose to disable them manually via the compiler/pragmas, but this annoying as well. So better fix the warnings.

sjaeckel:
> But I am generally against providing means to disable warnings.

IMO these warnings are fine to be disabled as we decided that those functions will be deprecated in a future version X.
I think it's legitimate if a user decides that it will keep her sources the way they are and will do the update to version X when it is out!
Which means that this user deserves an easy way to disable those warnings.

> The same holds for MP_WUR. Maybe we shod remove the possibility to disable it in 2.0.

For this topic I'm 100% on your side!

minad:
Well but actually MP_WUR is the much worse issue. So the user will be annoyed by that even more. Disabling MP_WUR will hide real errors.

Since the deprecated warnings require less effort to work around, I wouldn't provide a means to disable them either.

But I see your point. If a user wants the program to work with the old and new version then they need to disable the warnings. And this is different from MP_WUR.

sjaeckel:
> Well but actually MP_WUR is the much worse issue.

Very true! That's also why I'm pro disallowing its disabling in 2.0

minad:
You can add the patch for MP_WUR here too?

sjaeckel:
> You can add the patch for MP_WUR here too?

uuuhhhhmmmmm ... I'm not sure if we really need that, I think people should still have the ability to do stupid things sometimes ;-)
if someone updates to 1.2 there appear these warnings ... and if this person sees them and decides to actively ignore them I think there's something not right ... but for me it's okay!
we warned -> the warnings got ignored -> something bad happens -> not our fault :-) 

also it'd be against our self-defined rule of making 1.2 a drop-in replacement

minad:
Ah is this for 1.x? I mean only for 2.0! The we should have a separate pr.

sjaeckel:
This is only for 1.x

2.0 has no deprecation warnings anymore :) 

ah, I just saw that the macros still exist... I'll update the base branch of this PR to develop, add the WUR change and then I'll cherry-pick the relevant part after merging over to support/1.x

minad:
the deprecated macros exist for future use, but we could also remove them in 2.0 and reintroduce them for 2.1 ;)

PR #436, opened 2019-10-29
some linting

fperrad:


minad:
uL suffix is wrong for small digits (MP_16BIT).

fperrad:
reworked

demo/test.c
```diff
@@ -588,7 +588,7 @@ static int test_mp_set_double(void)
 
    for (i = 0; i < 1000; ++i) {
       int tmp = rand_int();
-      double dbl = (double)tmp * rand_int() + 1;
+      double dbl = (double)((tmp * rand_int()) + 1);
```
czurnieden: The chance is quite high that `tmp * rand_int()` overflows, which just happend in Travis. If you want ot add parenthese it needs to be more something like `double dbl = ( (double)(tmp * rand_int() ) + 1.0;` or maybe even cast `rand_int()` to and do a `double dbl =  (double)tmp * (double)rand_int()) + 1.0;`

fperrad: @czurnieden fixed with another way

czurnieden: > fixed with another way

Travis moans: "warning: cast from function call of type ‘int’ to non-matching type ‘double’"

`rand_int()` -> `rand_double()`?

```c
static double rand_double(void)
{
   uint64_t x;
  /* Doesn't work with feeding a float directly?  */
   if (s_mp_rand_source(&x, sizeof(x)) != MP_OKAY) {
      fprintf(stderr, "s_mp_rand_source failed\n");
      exit(EXIT_FAILURE);
   }
   return (double)x;
}
```

And then:

```c
double tmp = rand_double();
double dbl = (tmp * rand_double()) + 1.0;
```

But there is a change that this produces `NaN`, `+-inf`, and denormal numbers, we would need to weed them out in `rand_double()`.

fperrad: I think that a `rand_double()` could help.

But at this time, I'll remove this part of the commit.

czurnieden: > But there is a chance that this produces `NaN`, `+-inf`, and denormal numbers, we would need to weed them out in `rand_double()`.

No it wouldn't, because it is just a cast; we use the value not the bitpattern.
Sorry, my fault.

PR #438, opened 2019-10-30
fix link error when building shared lib.

cntrump:
if one of S_READ_* macro undefined, linker can't find s_read_* function.

### DIFF
--- a/s_mp_rand_platform.c
+++ b/s_mp_rand_platform.c
@@ -120,10 +120,15 @@ mp_err s_read_urandom(void *p, size_t n);
 mp_err s_mp_rand_platform(void *p, size_t n)
 {
    mp_err err = MP_ERR;
-   if ((err != MP_OKAY) && MP_HAS(S_READ_ARC4RANDOM)) err = s_read_arc4random(p, n);
-   if ((err != MP_OKAY) && MP_HAS(S_READ_WINCSP))     err = s_read_wincsp(p, n);
-   if ((err != MP_OKAY) && MP_HAS(S_READ_GETRANDOM))  err = s_read_getrandom(p, n);
-   if ((err != MP_OKAY) && MP_HAS(S_READ_URANDOM))    err = s_read_urandom(p, n);
+#if defined(S_READ_ARC4RANDOM_C)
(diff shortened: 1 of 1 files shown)


minad:
What is your problem exactly? Which compiler, linker, system? Please open an issue with a detailed problem description instead. This PR is not a proper fix.

minad:
To clarify - if the macros are undefined, the function should be disabled by the if MP_HAS conditional. If this doesn't work for some reason in your case, we have to understand why. For that we need the details.

minad:
@sjaeckel good to know. I can look into this later. The MP_HAS macros ensure that the function calls never reach the linker. So this shouldn't be a linking error.

cntrump:
> @sjaeckel good to know. I can look into this later. The MP_HAS macros ensure that the function calls never reach the linker. So this shouldn't be a linking error.

thanks you two!

I have another question, it should add `|| (defined(__APPLE__) && defined(__MACH__))` for enable `arc4random_buf` in `s_read_arc4random ` when building on macOS ?

sjaeckel:
FYI I just had a second reporter for the same problem, but this time on Windows...

minad:
Which compiler, when does this happen? Can you reproduce it? I think it doesn't happen on VC6 because we tested this and it doesn't happen on vs2015+.

But maybe the compiler behaves in a weird way if shared libraries are build.

make -f makefile.shared works well though.

minad:
It could also be a issue with optimization. On gcc if the library is compiled with O0 the dead code is still removed because it is in fact dead code.

But maybe other compilers behave differently? But then the issue should be gone if minimal optimizations are turned on.

minad:
@sjaeckel I assume the issue occurs for the rand functions, since these are the only ones where we rely on dead code elimination for a standard library build.

One possiblity to fix this could be to to it as follows, such that the standard build will never have this issue.

~~~ c
#if windows
#define S_RAND_SOURCE_WIN_C
static s_rand_source_win(...) { ... }
#else 
static void s_rand_source_win(...) { return MP_ERR; }
#endif
~~~

sjaeckel:
#452 reproduces the issue

PR #439, opened 2019-10-30
mp_radix off-by-one error and other related code-cleanup

nijtmans:
Noted an off-by-one error, testing for MP_RMAP_REVERSE_SIZE < pos.  Explanation: If MP_RMAP_REVERSE_SIZE == pos this returns false, so the code continues to read the table. Lucky, this only happens when the char value is 128, which is impossible for signed characters. So, this bug can actually only happen on a system with unsigned characters.

Further on, I noted various other tiny mistakes, e.g. the use of SIZE_MAX in stead of the real available bffer-size.   Also the s_mp_rmap_reverse table can be reduced 8 in lenght, if we start with '+' in stead of with '('. This branch contains all I found related to mp_radix stuff.

Anyway, I hope you like this clean-up.


nijtmans:
One more remark: ((y == 0xff) || (y >= radix)).    The check for y == 0xff is redundant, because the radix is already checked before to be between 2 and 64. So if y == 0xff, it will be >= radix for sure.

minad:
Good find!

minad:
You might only want to reformat the table 😁

minad:
And the 0xff doesn't seem redundant since there are entries in the table.

minad:
@sjaeckel Good to merge?

nijtmans:
@sjaeckel ping?

PR #430, opened 2019-10-28
use size_t

minad:
Work in progress #363. Just to let you know @sjaeckel, @czurnieden, @nijtmans. I am going over the functions one by one. See TODO file.

PR #446, opened 2019-11-05
Alternative 2: Suffix renamings (_n suffix)

minad:
alternative to #437


doc/bn.tex
```diff
@@ -1906,9 +1906,9 @@ \section{Combined Modular Reduction}
 
 \chapter{Exponentiation}
 \section{Single Digit Exponentiation}
-\index{mp\_expt\_u32}
+\index{mp\_expt\_n}
 \begin{alltt}
-mp_err mp_expt_u32 (const mp_int *a, uint32_t b, mp_int *c)
+mp_err mp_expt_n(const mp_int *a, int b, int *c)
```
nijtmans: Signature of mp_expt_n in bn.tex doesn't match the signature in header-file and implementation

minad:
Is this _n suffix acceptable for you? The root function was called n_root before. I think the _d suffix should be reserved for digits.

minad:
Yes, we also had the unsigned discussion in #437. Let's keep things separate.

mp_root_n.c
```diff
@@ -82,7 +85,7 @@ mp_err mp_root_u32(const mp_int *a, uint32_t b, mp_int *c)
 
       /* denominator */
       /* t3 = t1**(b-1) * b  */
-      if ((err = mp_mul_d(&t3, b, &t3)) != MP_OKAY)               goto LBL_ERR;
+      if ((err = mp_mul_d(&t3, (mp_digit)b, &t3)) != MP_OKAY)               goto LBL_ERR;
```
czurnieden: Before I forget it again: with `MP_16BIT` an <del>`uint32_t`</del> `int` might be bigger  than an `mp_digit` and the value of `b` has only been checked to be smaller than or equal to `INT_MAX/2`.
This might be a problem at other places, too, please check.

mp_root_n.c
```diff
@@ -12,15 +12,18 @@
...
  */
-mp_err mp_root_u32(const mp_int *a, uint32_t b, mp_int *c)
+mp_err mp_root_n(const mp_int *a, int b, mp_int *c)
 {
    mp_int t1, t2, t3, a_;
-   mp_ord cmp;
    int    ilog2;
    mp_err err;
 
+   if (b < 0 || (unsigned)b > (unsigned)MP_DIGIT_MAX) {
```
minad: @czurnieden There is a check here.

czurnieden: So you think I need a new prescription for my reading glasses? ;-)

minad:
@sjaeckel I guess you prefer this over alternative 1? Ready from my side.

sjaeckel:
> @sjaeckel I guess you prefer this over alternative 1? Ready from my side.

I am not sure, that's why I left them both untouched. Does this block you?

minad:
This is not a big blocker. Maybe a bit for #430. But at at some point we have to resolve this.

sjaeckel:
Okay, so you think we should go with this version?! :)

I'm fine with this version of the API.

Now let's head for 2.0 with this version as there are still some discussions upcoming e.g. regarding unsigned arguments.

minad:
I don't really care, but having it like this leaves the door open for all-mp_int functions.

minad:
This is for 2.0. but we should backport the renaming/deprecation. That's all for 1.x

minad:
> Now let's head for 2.0 with this version as there are still some discussions upcoming e.g. regarding unsigned arguments.

What do you mean by that? Can this be merged?

sjaeckel:
> Can this be merged?

as soon as it's rebased

PR #448, opened 2019-11-06
some refactoring

fperrad:


minad:
I don't think this is more readable :-1:

mp_div_2d.c
```diff
@@ -34,16 +34,15 @@ mp_err mp_div_2d(const mp_int *a, int b, mp_int *c, mp_int *d)
...
+      mp_digit mask = ((mp_digit)1 << b) - 1uL;
 
       /* shift for lsb */
-      shift = (mp_digit)(MP_DIGIT_BIT - b);
+      mp_digit shift = (mp_digit)(MP_DIGIT_BIT - b);
 
       /* carry */
-      r = 0;
+      mp_digit r = 0;
       for (x = c->used; x --> 0;) {
```
minad: This is good however.

sjaeckel: but would break c89

minad: No these are still only a list of inits. We check c89 with a compiler flag.

minad:
Close this since it is outdated. And we probably don't want these single line initializations.

PR #449, opened 2019-11-06
Fix wrong use of uLL suffix

nijtmans:
Some platforms (e.g. VC++ 6.0) don't use the uLL postfix for constants. Even ... it's wrong to use it for uint64_t constants, because uint64_t might be actually "long" while "long long" is 128-bit on some platforms. Therefore <stdint.h> has macro's for it doing it in a portiable way. But ... in most places, simply using the "u" suffix is sufficient.

This PR contains fixes for all places I found. And a few more things, like explicit casts for malloc() results (on most places this is done correctly, needed for C++, but I found 3 places this was missing)

This PR can be merged as-is for the support/1.x branch, it gives conflicts when merged on master.

### DIFF
--- a/demo/test.c
+++ b/demo/test.c
@@ -763,7 +763,7 @@ static int test_mp_get_u64(void)
    }
 
    for (i = 0; i < (int)(MP_SIZEOF_BITS(uint64_t) - 1); ++i) {
-      r = (UINT64_C(1) << (i+1)) - 1;
+      r = ((uint64_t)1 << (i+1)) - 1;
       if (!r)
          r = UINT64_MAX;
       printf(" r = 0x%" PRIx64 " i = %d\r", r, i);
--- a/demo/test.c
+++ b/demo/test.c
@@ -754,7 +754,7 @@ LBL_ERR:
 
 static int test_mp_get_u64(void)
 {
-   unsigned long long q, r;
+   uint64_t q, r;
    int i;
 
    mp_int a, b;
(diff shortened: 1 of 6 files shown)


minad:
You also have to replace the UINT64_C with MP_UINT64_C for the c89 mode I guess.

nijtmans:
> You also have to replace the UINT64_C with MP_UINT64_C for the c89 mode I guess.

Yes, but the support/1.x doesn't have that yet

demo/test.c
```diff
@@ -2313,7 +2313,7 @@ static int test_mp_read_write_ubin(void)
 
    size = mp_ubin_size(&a);
    printf("mp_to_ubin_size  %zu\n", size);
-   buf = malloc(sizeof(*buf) * size);
+   buf = (unsigned char *)malloc(sizeof(*buf) * size);
```
minad: uint8_t

minad: Disregard this on 1.x

minad:
@nijtmans could you please prepare the same pr for develop please?

demo/test.c
nijtmans: > uint8_t

No, uint8_t is wrong here, because mp_to_ubin has an "unsigned char *" argument. (on develop it might be different, but this PR is for support/1.x)

nijtmans:
> Now that I think about it. I prefer the (uint64_t)1 instead of UINT64_C(1). The UINT64_C macro is only needed for the very large constants.

I can live very well with (uint64_t) cast as well. :-) Done now.

nijtmans:
> @nijtmans could you please prepare the same pr for develop?

I'm happy to do that, once this one is merged to support/1.x.   I would like to do the rebase and conflict fixes only once.

PR #450, opened 2019-11-06
make mp_sqr and mp_div_3 internal and add optimizations

minad:
Do we want that or do we want the user to explicitly call mp_sqr if the sqr function should be used?
GMP does this optimization I think, but they also don't expose a sqr function in the API.


minad:
@nijtmans For example in Tcl, does it use mp_sqr there or always calls mp_mul? Then it might make sense to push this optimization to ltm.

Furthermore we could think about removing mp_sqr from the public api alltogether and replacing it by a macro `mp_mul(a, a, b)` given we merge this PR.

minad:
The advantage of making sqr purely an internal optimization of sqr is certainly that if the lib is downstripped to the bare minimum, the sqr functions could be removed completely. And mp_sqr could still be used via the macro.

MasterDuke17:
Fwiw, I don't believe MoarVM uses mp_sqr.

minad:
> Yes, but deprecate it formally, please, if you want to privatize mp_sqr.

We will do deprecations in 1.x when we remove the functions in develop. I will prepare a backport for that.

minad:
I also made mp_div_3 internal since it is only an optimization. I added more optimizations to mp_div_d and mp_mul_d.

minad:
mp_sqr will stay as macro btw. The point is really API surface reduction.

minad:
@sjaeckel this is ready from my side

minad:
@sjaeckel Rebased

PR #454, opened 2019-11-11
Better use of isneg

nijtmans:
Consistancy in the use of mp_isneg() and mp_iszero().   Many places compare a->sign explicitly with MP_ZPOS or MP_NEG, or compare a->used explictly with 0. Using mp_isneg()/mp_iszero() makes this code much more readable. No difference in functionality.

Also the use of variables (e.g. always using variable name "sign" of type mp_sign, and variable name "neg" of type bool) consistantly.


minad:
looks good

PR #457, opened 2019-11-12
Wrong use of ull for develop

nijtmans:
As promised in #449 : This is the same PR as already merged to support/1.x, but now cherry-picked to "develop"

### DIFF
--- a/demo/test.c
+++ b/demo/test.c
@@ -223,9 +223,9 @@ static int test_mp_get_set_i64(void)
 
    DOR(mp_init(&a));
 
-   check_get_set_i64(&a, (int64_t)0);
-   check_get_set_i64(&a, (int64_t)-1);
-   check_get_set_i64(&a, (int64_t)1);
+   check_get_set_i64(&a, 0);
+   check_get_set_i64(&a, -1);
+   check_get_set_i64(&a, 1);
    check_get_set_i64(&a, INT64_MIN);
    check_get_set_i64(&a, INT64_MAX);
 
--- a/demo/test.c
+++ b/demo/test.c
@@ -223,9 +223,9 @@ static int test_mp_get_set_i64(void)
 
    DOR(mp_init(&a));
 
-   check_get_set_i64(&a, 0LL);
-   check_get_set_i64(&a, -1LL);
-   check_get_set_i64(&a, 1LL);
+   check_get_set_i64(&a, (int64_t)0);
+   check_get_set_i64(&a, (int64_t)-1);
+   check_get_set_i64(&a, (int64_t)1);
    check_get_set_i64(&a, INT64_MIN);
    check_get_set_i64(&a, INT64_MAX);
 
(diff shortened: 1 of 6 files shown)


demo/test.c
```diff
@@ -223,9 +223,9 @@ static int test_mp_get_set_i64(void)
 
    DOR(mp_init(&a));
 
-   check_get_set_i64(&a, 0LL);
```
minad: Here the cast is unnecessary. I would prefer if we only add casts where needed, e.g., in code like `(uint64_t)1 << ...`

nijtmans: Agreed. Just as the "LL" postfix shouldn't have been here originally in the first place ;-)

demo/test.c
```diff
@@ -637,8 +637,8 @@ static int test_mp_get_u32(void)
       printf("\nmp_get_u32() bad result!");
       goto LBL_ERR;
    }
-   mp_set_ul(&a, 0xFFFFFFFFuL);
-   if (mp_get_u32(&a) != 0xFFFFFFFFuL) {
+   mp_set_ul(&a, (unsigned long)UINT32_MAX);
```
minad: there are more unnecessary casts, for example this one.

demo/test.c
```diff
@@ -625,7 +625,7 @@ static int test_mp_get_u32(void)
    DOR(mp_init_multi(&a, &b, NULL));
 
    for (i = 0; i < 1000; ++i) {
-      t = (unsigned long)rand_long() & 0xFFFFFFFFuL;
+      t = (unsigned long)rand_long() & (unsigned long)UINT32_MAX;
```
minad: x

demo/test.c
```diff
@@ -2177,7 +2177,7 @@ static int test_mp_read_write_ubin(void)
 
    size = mp_ubin_size(&a);
    printf("mp_to_ubin_size  %zu - ", size);
-   buf = malloc(sizeof(*buf) * size);
+   buf = (unsigned char *)malloc(sizeof(*buf) * size);
```
minad: I think we use uint8_t* here now

demo/timing.c
```diff
@@ -47,7 +47,7 @@ static unsigned long lfsr = 0xAAAAAAAAuL;
 static unsigned int lbit(void)
 {
    if ((lfsr & 0x80000000uL) != 0uL) {
-      lfsr = ((lfsr << 1) ^ 0x8000001BuL) & 0xFFFFFFFFuL;
+      lfsr = ((lfsr << 1) ^ 0x8000001BuL) & (unsigned long)UINT32_MAX;
```
minad: x

s_mp_rand_jenkins.c
```diff
@@ -28,7 +28,7 @@ static uint64_t s_rand_jenkins_val(void)
 void s_mp_rand_jenkins_init(uint64_t seed)
 {
    int i;
-   jenkins_x.a = 0xF1EA5EEDuLL;
+   jenkins_x.a = (uint64_t)0xF1EA5EEDuL;
```
minad: x

s_mp_rand_jenkins.c
```diff
@@ -42,7 +42,7 @@ mp_err s_mp_rand_jenkins(void *p, size_t n)
       int i;
       uint64_t x = s_rand_jenkins_val();
       for (i = 0; (i < 8) && (n > 0u); ++i, --n) {
-         *q++ = (char)(x & 0xFFuLL);
+         *q++ = (char)(x & (uint64_t)0xFFu);
```
minad: x

demo/test.c
minad: Yes, unfortunately fperrad is overdoing it. You've seen me probably complaining about the linting before ;)

nijtmans:
I think all those unnecessary type-cast are gone now.

minad:
@sjaeckel This is ready to merge

sjaeckel:
can we somehow make sure that this doesn't happen again in the future?

minad:
Actually through the c89-c99 roundtrip some notations are fixed. Besides we might want to investigate if there is a better possibility for the linting in the CI. Furthermore we could add rules to this helper.pl checker. But then I don't think too much harm is done with such things, these are mostly stylistic things and only rarely lead to bugs.

PR #458, opened 2019-11-14
some linting

fperrad:


minad:
Yes, we should probably write down the rules. But I said this a few times before - I am not a fan of most of the lintings. There are many reason for my opinion.

* I don't like explicit braces in the trivial cases. For example `a+b*c` is obvious to anyone. If you cannot cope with operator precedence, use lisp. Even worse there are associative things like `a+b+c`, where I've also seen parentheses somewhere in the codebase if I remember correctly. There the parentheses are even harmful, since they cloud the algebraic properties of the operation.
* I don't like the explicit type suffix. For example 0u is useless, 0 is 0 if it is signed or unsigned. There is simply no reason for the suffix. Even strongly typed programming languages support literal overloading, e.g., Haskell.
* fperrad basically touches every line of code. Every time I want to browse the git history, this is so tedious since I have to move over all those meaningless linting commits. This single point would be reason enough for me to NOT do any of those lintings. I have to admit that I also did some mega commits touching many lines, but I think we should weigh the benefits.
* I think the lintings are weak and don't give sufficient benefit in contrast to *real* static analyzers like Coverity (which we should finally add to the CI...). However I am aware of one single linting, where a single bug has been uncovered due to wrong precendence (There was something like x*2/3 or so...).
* We are not testing the lintings in CI. If we could do that we wouldn't get all the unnecessary linting commits and the rules would be written down once in a configuration file. This would also make the style guide partially obsolete.

czurnieden:
Took the liberty to add two rules for parentheses to the [wiki page](https://github.com/libtom/libtommath/wiki/The-Sourcecode:-Formatting-and-Style) to start the process.

> I don't like explicit braces in the trivial cases.

Nobody does, but it leads to the question: what is trivial, where do you draw the line?

If you can define the place of that figurative line you should draft a rule and put it in the [wiki page](https://github.com/libtom/libtommath/wiki/The-Sourcecode:-Formatting-and-Style).

> I don't like the explicit type suffix.

Some things are also done for the human readers.

> I think we should weigh the benefits.

So let's do it:

> However I am aware of one single linting, where a single bug has been uncovered due to wrong precendence

An argument good enough for me to go on with the linting.

And we don't *have* to use the lintings, we can also say: "Nuh, don't like it!". But if you say so and can persuade the rest of us, you should also draft a rule for the [wiki page](https://github.com/libtom/libtommath/wiki/The-Sourcecode:-Formatting-and-Style).

> If we could do that [test in CI] we wouldn't get all the unnecessary linting commits and the rules would be written down once in a configuration file.

Yes, that kind of automatism would be nice to have.

minad:
For me things would be fine if automated and specified in a configuration file. But the way the process is done right now doesn't work well.

demo/test.c
```diff
@@ -1519,10 +1519,10 @@ static int test_mp_log_n(void)
 
    /*Test upper edgecase with base UINT32_MAX and number (UINT32_MAX/2)*UINT32_MAX^10  */
    mp_set(&a, max_base);
-   DO(mp_expt_n(&a, 10uL, &a));
+   DO(mp_expt_n(&a, 10u, &a));
```
nijtmans: Shouldn't this simply be "10" in stead of "10u" ?

mp_kronecker.c
```diff
@@ -85,7 +85,7 @@ mp_err mp_kronecker(const mp_int *a, const mp_int *p, int *c)
       }
 
       if ((v & 1) == 1) {
-         k = k * table[p1.dp[0] & 7u];
+         k = k * (int)table[p1.dp[0] & 7u];
```
nijtmans: Why not "7" in stead of "7u".  also the cast to (int) is (in my view) unnecessary here. (but I won't make a big point of that)

fperrad:
@nijtmans, from __MISRA C 2012__

__Rule 7.2__: A "u" or "U" suffix shall be applied to all integer constants that are represented in an _unsigned_ type

__Rule 10.1__: Operands shall not be of an inappropriate essential type
__Rational__:
6. Shift and bitwise operations should only be performed on operands of essentially _unsigned_ type.     The numeric value resulting from their use on essentially _signed_ types is implementation defined.
7. The right hand operand of a shift operator should be of essentially _unsigned_ type to ensure that undefined behaviour does not result from a negative shift.

sjaeckel:
> For me things would be fine if automated and specified in a configuration file. But the way the process is done right now doesn't work well.

Are there any FLOSS linters that could do that job? Maybe one of this list https://github.com/mcandre/linters ? or another that is not listed there?

nijtmans:
> @nijtmans, from **MISRA C 2012**
> 
> **Rule 7.2**: A "u" or "U" suffix shall be applied to all integer constants that are represented in an _unsigned_ type

Well, I never heared about MISRA before, but adding an u postfix in the wrong space can introduce bugs very difficult to discover. For example:
<pre>
#include <stdio.h>
void main(void) {
	int x = -5;
	printf("%d\n", x/7u);
}
$ ./demo
613566755
</pre>

I am not aware of any situation where adding an u postfix for constants <= INT_MAX gives any advantage or makes it more readable.

That said, it's a project decision, not mine, to decide whether to confirm to MISRA.

> 
> **Rule 10.1**: Operands shall not be of an inappropriate essential type
> **Rational**:
> 6. Shift and bitwise operations should only be performed on operands of essentially _unsigned_ type. The numeric value resulting from their use on essentially _signed_ types is implementation defined.
> 7. The right hand operand of a shift operator should be of essentially _unsigned_ type to ensure that undefined behaviour does not result from a negative shift.

If a shift is not expected to be negative, neither is it expected to have a high positive value. Most processors don't allow shifts bigger than the word-size (usually 32 or 64). So, using an unsigned type doesn't make the code work, it only hides the problem from the compiler.

So, reading this, I'm sceptical about MISRA 2012 (while - most likely - 90% of their rules make very much sense)

minad:
@nijtmans I am also very sceptical about that. Many of the lintings just lead to unnessecarily more verbose code without advantage. I would never write the suffices for example if not necessary. Additionally to the criticism regarding to "bad lintings", misra is also not available freely. There are no open source linters implementing misra since the misra specification cannot be distributed etc etc. I think we should stop using misra and replace it with something else.

minad:
Furthermore, there is a big difference in doing linting and prover based static analysis. These linting rules (add a suffix, add a cast there, initialize all variables directly etc etc) are mostly of syntactical nature. Fixing those things just leads to code consistency but won't find any serious issues. This only gives a false sense of security.

minad:
@sjaeckel I would propose to set up clang-analyze and coverity in the CI, which is serious static analysis. Additionally to that we could add some FOSS linter running in the CI which implements some of the stylistic, syntactical rules. Unfortunately until now I haven't found a linter which does the job.

The linter should support some of the rules we consider necessary. E.g. explicit parentheses where it makes sense (I think we add more right now than necessary). 

What are other projects doing - in particular the big ones written in C?

Issue #418, opened 2019-10-27
new version v1.x


sjaeckel:
A placeholder issue to reference PR's that should be (at least partially) included in a potential 1.2.1 or 1.3.0

minad:
#437 should be backported (only the renaming and deprecations in order to ease the transition)

minad:
#442

minad:
#450 deprecate mp_div_3 (private in 2.0)

minad:
#452 MP_HAS issue

minad:
#471 overflow check for safety

minad:
#476 double cpp check

dod38fr:
Hi

Debian will enter freeze in January to prepare next Debian release. Currently, Debian ships libtommath 1.2.0. 

If you think libtommath should be updated for next Debian release, you should cut a new libtommath release no later than December.

All the best

Dod (for Debian)

royjit:
Hi, 

Do you have plans or ETA of when the next stable version will be released?

Thanks,
Jithin

dod38fr:
Hi
The next Debian stable version (2) will be released in Summer 2023.

To prepare this release, freeze will begin in January. 

All the best

dod38fr:
Hi

Debian will begin freeze process in January to prepare Debian 12 release. Currently, Debian still ships libtommath 1.2.0 which is 3 years old.

If you think libtommath should be updated for Debian 12, you should cut a new libtommath release as soon as possible.

All the best

Dod (for Debian)

czurnieden:
Oh crap! Too late again, hu?
Otherwise: it would be ready as it it is.
@sjaeckel I haven't done a proper publishing before and am too afraid to eff up, sooooo...if you don't mind  :-)
Or is there a hindrance I am unaware of?

dod38fr:
> Oh crap! Too late again, hu?

Not too late. Barring backward compatibility issue, I should be able to update libtommath for Debian 12 if it's released in the next 3 weeks. But the sooner a new release is done, the better.

Happy new year !

sjaeckel:
Cool, could you please take develop and check whether a release would be fine?

dod38fr:
Unfortunately, at least a symbol is gone (`mp_iseven` is replaced by a `#define`). This breaks the ABI compatibility of libtommath. 

[...]

PR #463, opened 2019-11-22
Proposal: Remove long long support

minad:
I propose to simply remove long long functions, since it is not c89 and u64/i64 can be used instead everywhere. I think that would work for most people. See also the table https://de.wikipedia.org/wiki/Datentypen_in_C#Datenmodell

These functions have been originally added in 2014 by @sjaeckel with the name mp_get_long_long etc.
Then I turned them in a macro in #285, which I think would still be an acceptable solution. Afterwards they were made explicit symbols by @nijtmans in #321. Now it seems @nijtmans needs some MSC_VER ifdef guards around these functions.

Would that work for you @nijtmans?

### DIFF
--- a/doc/bn.tex
+++ b/doc/bn.tex
@@ -978,30 +978,6 @@ unsigned long mp_get_mag_ul (const mp_int *a);
 This will return the least significant bits of the big integer $a$ that fit into the native data
 type \texttt{long}.
 
-\subsection{Long Long Constants - platform dependant}
-
-\index{mp\_set\_ll} \index{mp\_set\_ull}
-\begin{alltt}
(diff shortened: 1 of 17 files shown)


minad:
@nijtmans We would only deprecate the functions in 1.x.

I also thought about making mp_get_ul macros pointing to either u32 or u64 as it has been before your PR #321. However one thing I realized and I am not too happy about is the selection of the function using the ternary operator, which again relies on dead code elimination. While we have that too for the MP_HAS macro, there it is purely a library internal thing - but mp_get_ul is public api.

sjaeckel:
This needs a sibling patch for 1.x #418

Issue #433, opened 2019-10-29
make exptmod algorithms private?


minad:
Does it make sense to make the exptmod algorithms private, i.e., the mp_reduce* functions in 2.0? Are they useful by itself? Some of these functions look as if they are more of internal use, but I might be wrong. The same applies maybe to some of the prime functions. Could someone enlighten me? @czurnieden

sjaeckel:
I just wanted to say "I think there are even more currently public functions which could be private" but after having a look it seems like there's only the reduce stuff left

```
mp_err mp_reduce_setup(mp_int *a, const mp_int *b) MP_WUR;
mp_err mp_reduce(mp_int *x, const mp_int *m, const mp_int *mu) MP_WUR;
mp_err mp_montgomery_setup(const mp_int *n, mp_digit *rho) MP_WUR;
mp_err mp_montgomery_calc_normalization(mp_int *a, const mp_int *b) MP_WUR;
mp_err mp_montgomery_reduce(mp_int *x, const mp_int *n, mp_digit rho) MP_WUR;
bool mp_dr_is_modulus(const mp_int *a) MP_WUR;
void mp_dr_setup(const mp_int *a, mp_digit *d);
mp_err mp_dr_reduce(mp_int *x, const mp_int *n, mp_digit k) MP_WUR;
bool mp_reduce_is_2k(const mp_int *a) MP_WUR;
mp_err mp_reduce_2k_setup(const mp_int *a, mp_digit *d) MP_WUR;
mp_err mp_reduce_2k(mp_int *a, const mp_int *n, mp_digit d) MP_WUR;
bool mp_reduce_is_2k_l(const mp_int *a) MP_WUR;
mp_err mp_reduce_2k_setup_l(const mp_int *a, mp_int *d) MP_WUR;
mp_err mp_reduce_2k_l(mp_int *a, const mp_int *n, const mp_int *d) MP_WUR;
```

maybe also those

```
mp_err mp_prime_fermat(const mp_int *a, const mp_int *b, bool *result) MP_WUR;
mp_err mp_prime_miller_rabin(const mp_int *a, const mp_int *b, bool *result) MP_WUR;
mp_err mp_prime_strong_lucas_selfridge(const mp_int *a, bool *result) MP_WUR;
mp_err mp_prime_frobenius_underwood(const mp_int *N, bool *result) MP_WUR;
```

sjaeckel:
your opinion? @czurnieden @nijtmans @karel-m

[...]

PR #465, opened 2019-11-27
Made reduce and its helpers private

czurnieden:
Made all `*reduce*` functions and their helpers private. The `*montgomery*` functions are used by LibTomCrypt and have been excepted from privatization in consequence.

### DIFF
--- a/tommath.h
+++ b/tommath.h
@@ -436,16 +436,6 @@ mp_err mp_is_square(const mp_int *arg, bool *ret) MP_WUR;
 /* computes the Kronecker symbol c = (a | p) (like jacobi() but with {a,p} in Z */
 mp_err mp_kronecker(const mp_int *a, const mp_int *p, int *c) MP_WUR;
 
-/* used to setup the Barrett reduction for a given modulus b */
-mp_err mp_reduce_setup(mp_int *a, const mp_int *b) MP_WUR;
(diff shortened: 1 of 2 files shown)


minad:
If we make them private, we should call them `s_*`. But then I also wonder if we somehow want to make barrett reduction etc available. I again have no real idea on how the API should look like. I am not sure if *just hiding* things will do it or if we have to come up with a better API. See what I wrote regarding the prime tests in #433.

czurnieden:
> If we make them private, we should call them s_*.

Argh!
It went so smooth, I knew I forgot something! ;-)

> But then I also wonder if we somehow want to make barrett reduction etc available. 

Yes, we can set the line somewhere in between, but where exactly?

> I again have no real idea on how the API should look like.

You were on the "simplification" path. And the motto of simplification is: make it as simple as possible but never more, which is a good example of the "not very helpful" mottos so let's just simplify until somebody complains.

Also: LTM is open source with a license that allows you to take it and run with it. It will never be a problem for any programmer to remove `MP_PRIVATE` if needed. Third party developers relying on thing being public&hellip;were all ears for complains, our development process is transparent.


(Documentation not yet updated)

czurnieden:
This PR is obsoleted by many others.

PR #469, opened 2019-11-29
Radix overestimate with 29 bits precision

czurnieden:
New try.
With approximation error this time.

The values `l_t` in the table are calculated in advance as `l_t = INT(log_2(b)*2^p)+1` where `b` is a base from the set `{2,..,64}, `p` the precision and `INT()` extracts the integer part. The extraction of the integer part is equivalent to truncating, the addition of unity keeps the sign of the absolute error positive.

The bit count of the input number `a` is `l_a = INT(log_2(a)) + 1` and again: the addition of unity keeps the sign of the absolute error positive.

The digit count `n` is evaluated by `n = floor((la *  k) / 2^29) + 1` with a positive absolute error for the truncating division by `2^29`. The computation `n = floor((la *  k) / 2^29) + 1` is the fixed point version of `n = floor(log(a)/log(b)) + 1` where `la` corresponds to `log(a)` and `k` (without the scaling) corresponds to `1/log(b)`. Both `la` and `k` are bigger than their real counterparts because of the addition of unity and teh multiplication `la * k` keeps this advantage, so `(la * k) > (log(a) * 1/log(b))`. The scaling reduces the precision and hence enlarges the error but the absolute error will still be positive.

The size of the reciprocal of the value for the scaling `1/2^29 ~ 10^-8` corresponds well to the results of the implementation where a maximum absolute error of 5 (without EOS and sign) was found experimentally (large range [2, 2^(2^31)-1] although small N).

The test in `demo/test.c` contains a branch to check the number `2^(2^31)-1`. Because of the large memory requirement it is made optional and can be included with the preprocessor flag `YES_REALLY`.

The size of the object file on this machine is of a reasonable 2.8k. `mp_radix_size` with `mp_log_n` and all helpers on the other hand sums up to 9k and is much slower.

This function works with a 16 or a 32 bit `int`. Larger sizes (e.g.: a 64 bit `size_t`) need a new table.

### DIFF
--- a/mp_radix_size_overestimate.c
+++ b/mp_radix_size_overestimate.c
@@ -1,0 +1,17 @@
+#include "tommath_private.h"
+#ifdef MP_RADIX_SIZE_OVERESTIMATE_C
+/* LibTomMath, multiple-precision integer library -- Tom St Denis */
+/* SPDX-License-Identifier: Unlicense */
+
+mp_err mp_radix_size_overestimate(const mp_int *a, const int radix, size_t *size)
+{
+   if (MP_HAS(S_MP_RADIX_SIZE_OVERESTIMATE)) {
+      return s_mp_radix_size_overestimate(a, radix, size);
+   }
+   if (MP_HAS(MP_RADIX_SIZE)) {
+      return mp_radix_size(a, radix, size);
+   }
+   return MP_ERR;
+}
+
+#endif
--- a/demo/shared.c
+++ b/demo/shared.c
@@ -6,8 +6,8 @@ void ndraw(const mp_int *a, const char *name)
    size_t size = 0;
    mp_err err;
 
-   if ((err = mp_radix_size(a, 10, &size)) != MP_OKAY) {
-      fprintf(stderr, "\nndraw: mp_radix_size(a, 10, %zu) failed - %s\n", size, mp_error_to_string(err));
+   if ((err = mp_radix_size_overestimate(a, 10, &size)) != MP_OKAY) {
+      fprintf(stderr, "\nndraw: mp_radix_size_overestimate(a, 10, %zu) failed - %s\n", size, mp_error_to_string(err));
       exit(EXIT_FAILURE);
    }
    buf = (char *)malloc(size);
--- a/demo/test.c
+++ b/demo/test.c
@@ -2220,7 +2220,6 @@ static int test_s_mp_radix_size_overestimate(void)
        284u, 283u, 281u, 280u, 279u, 278u, 277u, 276u, 275u,
        273u, 272u
    };
-#ifdef MP_RADIX_SIZE_OVERESTIMATE_TEST_BIG
    size_t big_results[65] = {
               0u,         0u,         0u,  1354911329u, 1073741825u,
       924870867u, 830760078u, 764949110u,   715827883u,  677455665u,
@@ -2236,7 +2235,6 @@ static int test_s_mp_radix_size_overestimate(void)
       371449582u, 369786879u, 368168034u,   366591092u,  365054217u,
       363555684u, 362093873u, 360667257u,   359274399u,  357913942
    };
-#endif
 
 /* *INDENT-ON* */
    if ((err = mp_init(&a)) != MP_OKAY)        goto LBL_ERR;
@@ -2267,7 +2265,6 @@ static int test_s_mp_radix_size_overestimate(void)
       }
       a.sign = MP_ZPOS;
    }
-#ifdef MP_RADIX_SIZE_OVERESTIMATE_TEST_BIG
    if ((err = mp_2expt(&a, INT_MAX - 1)) != MP_OKAY) {
       goto LBL_ERR;
    }
@@ -2295,7 +2292,6 @@ static int test_s_mp_radix_size_overestimate(void)
       }
       a.sign = MP_ZPOS;
    }
-#endif
    mp_clear(&a);
    return EXIT_SUCCESS;
 LBL_ERR:
--- a/demo/test.c
+++ b/demo/test.c
@@ -2016,6 +2016,7 @@ LBL_ERR:
    return EXIT_FAILURE;
 }
 
+
 static int test_mp_radix_size(void)
 {
    mp_int a;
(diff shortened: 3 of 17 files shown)


minad:
Hmm I thought we is wanted to accelerate base 10?? Power of 2 bases are already fast. Or will we also get a fast general to_radix for every base?

minad:
But besides that, I like the direction this is going! We could add a flag to enable the table only if more than base 10 is needed.

~~~ c 
size_t table_lookup(int base) { if (only_base10) return ... etc }
~~~

But I think this only makes sense if we get a fast to radix for all bases.

czurnieden:
> Or will we also get a fast general to_radix for every base?

That was my plan.

If @MasterDuke17 is still with us: do you have time to do it or do you want us (which most likely translates to "me" ;-) ) to take over?

> We could add a flag to enable the table only if more than base 10 is needed.

Yes, will be made modular once the full-table number conversion runs because the single main difference to base-10-only (plus powers-of-two because it's cheap) number-conversion would also be a table (even of the same size), which means that we can offer the choice between base-10-only and full table at compile time. (I would pack both tables in a single function for simplicity)

But one step after the other. Radix-size and the actual number conversion first and then the switch.

MasterDuke17:
> If @MasterDuke17 is still with us: do you have time to do it or do you want us (which most likely translates to "me" ;-) ) to take over?

I haven't had the time to work on this recently, but I plan to and should over the next couple weeks (I particularly expect to have some time in early January). If you'd like to work on it sooner, I won't have any hurt feeling or anything like that, you'd probably do it quicker and better. But if not I'll keep plugging away.

mp_radix_size_overestimate.c
```diff
@@ -0,0 +1,90 @@
...
+   }
+
+   if (MP_HAS(S_MP_LOG_2EXPT) && MP_IS_2EXPT((mp_digit)radix)) {
+      /* floor(log_{2^n}(a)) + 1 + EOS + sign */
+      *size = (size_t)(s_mp_log_2expt(a, (mp_digit)radix));
+      /* Would overflow with base 2 otherwise */
+      if (*size > (INT_MAX - 4)) {
+         return MP_VAL;
+      }
+      *size += 1u + 1u + (mp_isneg(a) ? 1u : 0u);
```
minad: we discussed this before and @sjaeckel agreed that we should simply write +3

mp_radix_size_overestimate.c
```diff
@@ -0,0 +1,90 @@
...
+   mp_set_l(&bi_bit_count, bit_count);
+   /* k = floor(2^29/log_2(radix)) + 1 */
+   mp_set_u32(&bi_k, logbases[radix]);
+   /* n = floor((la *  k) / 2^29) + 1 */
+   if ((err = mp_mul(&bi_bit_count, &bi_k, &bi_bit_count)) != MP_OKAY)                         goto LTM_ERR;
+   if ((err = mp_div_2d(&bi_bit_count, LTM_RADIX_SIZE_SCALE, &bi_bit_count, NULL)) != MP_OKAY) goto LTM_ERR;
+
+   /* The "+1" here is the "+1" in "floor((la *  k) / 2^29) + 1" */
+   /* n = n + 1 + EOS + sign */
+   *size = (size_t)(mp_get_l(&bi_bit_count) + 1U + 1U + (mp_isneg(a) ? 1U : 0U));
```
minad: same here

minad:
@czurnieden This looks good and simple - I like it!

mp_radix_size_overestimate.c
czurnieden: Now that you mentioned it I faintly remember, yes. ;-)
Why did I do&hellip;uh, wanted to add a kind of refinement to get an exact value out of it, too. It would need only one big exponentiation smaller than the input, only one single digit multiplication in the normal case and additionaly one very small multiplication with very big inputs. All in all a bit faster and smaller than using `mp_log` but you still have one big exponentiation and if you want to use `mp_log` elsewhere you have that balast on top, so I dropped that idea.

I still have the concept in Pari/GP here:

```text
log2(x)=log(x)/log(2);
numdigits_under_p(a, b) = {
   local(k, k2, n, bn, lbn, la, ldiff, i=0);
   \\ Result "k" fits into a 32-bit integer
   k =  floor(2^29/log2(b)) + 1;
   \\ la = floor(log(a)/log(2)) + 1;
   la = logint(a,2);
   \\ result can and will overflow a 32-bit integer!
   n = ((la *  k) >> 29) + 1;
   \\print("n = " n);
   \\ This is the place where a mp_radix_size_overestimate would return.
   \\ Otherwise: go on.
   \\return (n);
   \\ At this point n >= (floor(log(a)/log(b)) + m) with m > 0
   \\ so we decrement n by one.
   n = n - 1;
   \\ Now bn <= a in almost all cases because
   \\ "m" is not larger than 1 for numbers smaller than ~10^8 bits.
   bn = b^n;

   \\ comparing a, bn can be done by comparing flog2(a), flog2(bn)
   \\ instead which might be computationally cheaper.

   \\ Check the size of "bn" in bits and compute the difference to "a"
   \\ to get the correction necessary
   lbn = logint(bn,2);
   \\ first negatives at "random(2^(2^27))" at b=50,51
   \\ Some empirical tests, although with a small N but with very large
   \\ numbers (<2^(2^31)) gave ldiff = -16 bits as a max. (5 decimal)
   ldiff = la - lbn;

   if(ldiff != 0,
      \\if ( ldiff < 0 , print("ldiff = " ldiff));
      \\ k2 can be zero if ldiff <= round(log_2(b))
      k2 = (k * ldiff) >> 29;
      \\print("k2 = " k2 );
      if( k2 > 0,
         bn = bn * (b^k2);
      );
      if( k2 < 0,
         \\ exact division.
         bn = bn / (b^(-k2));
      );
      n = n + k2;
   );

   if(a == bn,
      return(n + 1);
   );
   if(a > bn,
      while( a > bn,
        bn = bn * b;
        n = n + 1;
      );
      \\ a < bn, return n
      return(n);
   );
   if(a < bn,
      while( a < bn,
        \\ again, exact division
        bn = bn / b;
        n = n - 1;
      );
   );

   \\ a >= bn, return n + 1.
   return(n + 1);
}
```
It's a rough sketch but probably better than my english description ;-)

czurnieden:
@MasterDuke17 
> I haven't had the time to work on this recently, but I plan to and should over the next couple weeks (I particularly expect to have some time in early January)  If you'd like to work on it sooner,

If I find time for it I'll give it a try. It was always very quiet this time of the year the last years, decades even, don't know what is happening this year. Climate change? Brexit? I don't know.

> But if not I'll keep plugging away.

Just keep plugging for now, but please update your workbench here at Github regularily, so others can step in if necessary. And don't be shy: ask if you don't understand something, I'm a friendly bear ;-)

mp_radix_size_overestimate.c
minad: Hmm I am not sure. If we could get a fast and small exact version I would prefer that since we could simply replace mp_radix_size. So we would only get mp_expt_n additionally? Maybe that would be ok, I think mp_expt is pretty small.

minad: @czurnieden how would it compare to the currently used mp_radix_size? This is already faster than the division approach we did before. But not particularly small. Maybe it makes sense to offer the very downstripped mp_radix_size_overestimate as proposed here since it is really only needed for allocation size estimation.

czurnieden: > So we would only get mp_expt_n additionally?

Yes. Instead of both `mp_log` and `mp_expt`

> how would it compare to the currently used mp_radix_size?

Not implemented in LTM, can't say. But there is negligible overhead beside the single big exponentiation. But there is not much use beside `mp_radix_size`.

> Maybe it makes sense to offer the very downstripped mp_radix_size_overestimate as proposed here since it is really only needed for allocation size estimation.

So you want to remove the exact `mp_radix_size`, which is basically just a convenient wrapper for `mp_log` now and replace it with a note in the documentation to use `mp_log` directly if an exact result is needed or `mp_radix_size_overestimate` for all of the other 99.999something% involving number conversions?
I'm all for it.

minad: I don't want to remove mp_radix_size. We discussed this before. The question is only if we add the overestimating function or if we change the internals of mp_radix_size instead.

minad:
@czurnieden I pushed some modifications. Do you agree with the changes?

czurnieden:
@minad 
> can there be an overflow? [de99c51](https://github.com/libtom/libtommath/pull/469/commits/de99c51e8c8b66939a093c7ffa2ecd1c08ead6ee)

Not for bases `>=3` because `1/log_2(3) ~ 0.6309`.

And also if you ignore the systemic problem that `(INT_MAX * MP_DIGIT_BIT) > INT_MAX` (For those not familiar with that problem: `used` is an `int` and `mp_bit_count` returns an `int`). Are there plans to change it for 2.0? (e.g.: let the `mp_bit_count` return type be larger than the type used for `used`).

Otherwise:

> Do you agree with the changes?

Yes, of course.
But you could rebase it, too, while you're at it.
Less fuddling for me ;-)

minad:
> Are there plans to change it for 2.0? (e.g.: let the mp_bit_count return type be larger than the type used for used)

I think we concluded that we should limit the allocation size such that mp_bit_count will never overflow.

czurnieden:
> I think we concluded that we should limit the allocation size such that mp_bit_count will never overflow.

"Multae viae ad Romam ducent" ;-)
Remember that there was a discussion but wasn't sure anymore if there was something planned for the near future, that is version 2.0.

PR #471, opened 2019-12-04
Prevent overflow

minad:
disabled the overestimate test for the INT_MAX-1 number for now since this number is not allowed anymore.

### DIFF
--- a/mp_grow.c
+++ b/mp_grow.c
@@ -8,6 +8,7 @@ mp_err mp_grow(mp_int *a, int size)
 {
    /* if the alloc size is smaller alloc more ram */
    if (a->alloc < size) {
+           /* TODO */
       /* reallocate the array a->dp
        *
        * We store the return in a temporary variable
--- a/mp_grow.c
+++ b/mp_grow.c
@@ -11,7 +11,7 @@ mp_err mp_grow(mp_int *a, int size)
       mp_digit *dp;
 
       if (size > MP_MAX_DIGIT_COUNT) {
-         return MP_MEM;
+         return MP_OVF;
       }
 
       /* reallocate the array a->dp
--- a/mp_grow.c
+++ b/mp_grow.c
@@ -8,16 +8,21 @@ mp_err mp_grow(mp_int *a, int size)
 {
    /* if the alloc size is smaller alloc more ram */
    if (a->alloc < size) {
-           /* TODO */
+      mp_digit *dp;
+
+      if (size > MP_MAX_DIGIT_COUNT) {
+         return MP_MEM;
+      }
+
       /* reallocate the array a->dp
        *
        * We store the return in a temporary variable
        * in case the operation failed we don't want
        * to overwrite the dp member of a.
        */
-      mp_digit *dp = (mp_digit *) MP_REALLOC(a->dp,
-                                             (size_t)a->alloc * sizeof(mp_digit),
-                                             (size_t)size * sizeof(mp_digit));
+      dp = (mp_digit *) MP_REALLOC(a->dp,
+                                   (size_t)a->alloc * sizeof(mp_digit),
+                                   (size_t)size * sizeof(mp_digit));
       if (dp == NULL) {
          /* reallocation failed but "a" is still valid [can be freed] */
          return MP_MEM;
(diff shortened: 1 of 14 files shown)


tommath_private.h
```diff
@@ -140,22 +140,29 @@ typedef uint64_t mp_word;
...
-MP_STATIC_ASSERT(prec_geq_min_prec, MP_PREC >= MP_MIN_PREC)
+#define MP_MIN_SIZE MP_MAX(3, (((int)MP_SIZEOF_BITS(uint64_t) + MP_DIGIT_BIT) - 1) / MP_DIGIT_BIT)
+MP_STATIC_ASSERT(prec_geq_min_prec, MP_DEFAULT_SIZE >= MP_MIN_SIZE)
+
+/* Maximum number of digits.
+ * - Must be small enough such that mp_bit_count does not overflow.
+ * - Must be small enough such that mp_radix_size for base 2 does not overflow.
+ *   mp_radix_size needs two additional bytes for zero termination and sign.
+ */
+#define MP_MAX_SIZE ((INT_MAX - 2) / MP_DIGIT_BIT)
```
czurnieden: `MP_MAX_SIZE` is, although only internally used, a bit nondescript. It is the maximum number of limbs allowed, so something in the line of `MP_MAX_LIMB_COUNT` maybe?

minad: We could also go back to MAX_PREC if that's desired.

minad:
@czurnieden could you do me a favor and recompute the biggest number table for the overestimate test which is disabled now.

tommath_private.h
minad: MAX_DIGIT_COUNT, DEFAULT_DIGIT_COUNT, MIN_DIGIT_COUNT?

czurnieden: I like `MAX_DIGIT_COUNT` (`MP_MAX_DIGIT_COUNT`?).

You don't have to change it, but I found `MP_MAX_SIZE` a bit too ambiguous.

czurnieden:
> could you do me a favor and recompute the biggest number table for the overestimate test which is disabled now?

If you want to make the maximum size dependent on `MP_DIGIT_BIT` we would need several tables, one for each size/`MP_xBIT` and two for `MP_16BIT` because `sizeof(int)` can be 2.

That may take a day or two ;-)

minad:
Uuh could we do it somehow differently?

minad:
@czurnieden I added a randomized test for now.

minad:
@sjaeckel Right, I added a workaround using the tommath_c89.h header.

minad:
@sjaeckel Good to merge?

minad:
@sjaeckel Can we merge this?

mp_grow.c
```diff
@@ -8,15 +8,21 @@ mp_err mp_grow(mp_int *a, int size)
 {
    /* if the alloc size is smaller alloc more ram */
    if (a->alloc < size) {
+      mp_digit *dp;
+
+      if (size > MP_MAX_DIGIT_COUNT) {
+         return MP_MEM;
```
czurnieden: Ah, not yet merged, good.
Would you do me a favour and use a dedicated error--code instead of `MP_MEM` for it, here and at `mp_init_size`? Maybe `MP_SIZE` or something in that line?
Not out of principle, I can make use of it.

minad: Good idea! I added that. What is your use case, may I ask?

czurnieden:
>  What is your use case, may I ask?

A new and improved `mp_log_n`.
And also extended to full bigint if my calculations are not too much off.

It nagged me from the start on, that it has a lot of full bigint operations with up to twice the number of digits of the input, the trial exponentiations can overflow `MP_MAX_DIGIT_COUNT * MP_DIGIT_BIT` (the squarings too, but you could put a lid on them), and that the base is, although not theoretically but <i>in praxi</i> restricted to a small native integer instead of the actual upper limit  `(MP_MAX_DIGIT_COUNT * MP_DIGIT_BIT) - 1` .

You can get that down to one trial exponentiation that does not overflow and up to two small trials. One of the trials is by multiplying with the base which can overflow but that overflow is usable by the trial, that's why I wanted the distinct error. The other trial is a division by the base. It is exact, but as long as we don't have "exact division" also slower than the multiplication for bigint bases.

That together with a rather large "hidden constant" might make it significantly slower than the old method for small input but I have no benchmarks at this stage.

So don't hold your breath ;-)

PR #472, opened 2019-12-05
Change version numbers before 2.0 release

nijtmans:
The (microsoft) convension for dll's doesn't start with "lib" as on UNIX. So I suggest to use "tommath2.dll" as dll name. Also I made the other necessary corresponding makefile-related changes for the major upgrade.

### DIFF
--- a/makefile.mingw
+++ b/makefile.mingw
@@ -25,7 +25,7 @@ LTM_LDFLAGS = $(LDFLAGS) -static-libgcc
 #Libraries to be created
 LIBMAIN_S =libtommath.a
 LIBMAIN_I =libtommath.dll.a
-LIBMAIN_D =libtommath.dll
+LIBMAIN_D =tommath2.dll
 
 #List of objects to compile (all goes to libtommath.a)
 OBJECTS=mp_2expt.o mp_abs.o mp_add.o mp_add_d.o mp_addmod.o mp_and.o mp_clamp.o mp_clear.o mp_clear_multi.o \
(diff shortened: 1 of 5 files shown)


minad:
I think the "big change" here is the name change of the dll. I am not sure if we want to include the version number in the name. This is the job of a distribution or package builder I think. @sjaeckel will take care of the other changes regarding the version numbers.

nijtmans:
On other platforms than Windows - indeed - it is not customary to have a version number in the library name. On Windows it's normal (examples: ws2_32.dll, zlib1.dll, winsqlite3.dll, tcl86.dll, msvcr100.dll ...).  Windows dll's simply don't have a suitable mechanism as other platforms do.

makefile.mingw
```diff
@@ -25,7 +25,7 @@ LTM_LDFLAGS = $(LDFLAGS) -static-libgcc
 #Libraries to be created
 LIBMAIN_S =libtommath.a
 LIBMAIN_I =libtommath.dll.a
```
sjaeckel: so this name and the one above is okay?

nijtmans: Yes, that's OK. Static libraries are normally not versioned, just as header files like tommath.h. We wouldn't make that "tommath2.h" either.  Only runtime-files should have their version number in the file-name, because two dll's need to co-exist eventually on the same system.

nijtmans:
ping?

sjaeckel:
This will only be merged right before the 2.0.0 release, so there's no need to rebase now.

czurnieden:
> This will only be merged right before the 2.0.0 release, so there's no need to rebase now.

Ah, thanks!

PR #477, opened 2020-02-21
Fix clang detection when cross-compiling.

J08nY:
`echo` needs -e to output newlines, without the switch make complains:
```
    <stdin>:1:17: warning: extra tokens at end of #ifdef directive
    <stdin>:1: error: unterminated #ifdef
```
and does not detect clang properly when cross-compiling.

### DIFF
--- a/makefile_include.mk
+++ b/makefile_include.mk
@@ -26,7 +26,7 @@ ifeq ($(PLATFORM),FreeBSD)
   # XXX: FreeBSD needs extra escaping for some reason
   CSTR := $$$(CSTR)
 endif
-ifneq (,$(shell echo $(CSTR) | $(CC) -E - | grep CLANG))
+ifneq (,$(shell printf $(CSTR) | $(CC) -E - | grep CLANG))
   CC := $(CROSS_COMPILE)clang
 else
   CC := $(CROSS_COMPILE)gcc


czurnieden:
Option `-e` is not in [Posix](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/echo.html). XOpen conformant implementations should recognize the "standard" escapes without telling.

Are you able to use [`printf`](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/printf.html) instead?

minad:
@czurnieden Are you sure that our makefiles and not gmakefiles?

czurnieden:
> @czurnieden Are you sure that our makefiles and not gmakefiles?

OP changed `makefile_include.mk` which should contain the largest common denominators only (Posix in this case). So no extensions (GNU or otherwise, not even XOpen as we just saw), just plain Posix.

Which leads to the question: how common is `printf`?

J08nY:
https://pubs.opengroup.org/onlinepubs/9699919799/utilities/echo.html encourages to use `printf` instead of `echo`. So I changed this to use `printf`.

minad:
@J08nY Thanks! It seems @czurnieden knows his stuff, the guardian of the Posix.

J08nY:
> Can you please rebase on top of develop?

Done.

czurnieden:
> It seems @czurnieden knows his stuff, the guardian of the Posix.

I think I can safely take that as a thinly veiled "Ok, boomer!", can I not? ;-)

PR #482, opened 2020-04-12
add mp_find_prime() from Heimdal

lhoward:
Heimdal adds a `mp_find_prime()` function for its internal use. We'd like to move Heimdal to be able to use the system-provided libtommath and this would help us (of course, we could include the function locally).

### DIFF
--- a/libtommath_VS2008.vcproj
+++ b/libtommath_VS2008.vcproj
@@ -429,6 +429,10 @@
 			>
 		</File>
 		<File
+			RelativePath="mp_find_prime.c"
+			>
+		</File>
+		<File
 			RelativePath="mp_fread.c"
 			>
 		</File>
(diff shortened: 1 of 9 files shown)


lhoward:
Will investigate, I didn't actually spend time to understand what the code was doing – was just wearing "build maintainer" hat. I shall look more deeply.

lhoward:
OK, your comments make sense to me. I'll have someone review this change on our side. FWIW our code looks like:

```c
    /* generate p and q so that p != q and bits(pq) ~ bits */
    counter = 0;
    do {
        BN_GENCB_call(cb, 2, counter++);
        CHECK(random_num(&p, bitsp), 0);
        CHECK(mp_prime_next_prime(&p,128,0), MP_OKAY);

        mp_sub_d(&p, 1, &t1);
        mp_gcd(&t1, &el, &t2);
    } while(mp_cmp_d(&t2, 1) != 0);

    BN_GENCB_call(cb, 3, 0);
```

sjaeckel:
Just FYI `t=128` is really too high and would be very expensive...

You can call `mp_prime_rabin_miller_trials(bitsp)` for appropriate values of `t`!

lhoward:
Great, thank you – will discuss with the rest of the team who understand this better than I. Thank you for your help :)

PR #484, opened 2020-04-26
Add _M_IX86 for testing for MSVC x86

Prince213:
This PR fixes using mp_set_double with MSVC x86 since the compiler complains: `error LNK2001: unresolved external symbol _mp_set_double`.

By looking at the source code I see that `_M_X86` is used, but this macro doesn't seem to available anymore according to VS2019 documentation at [here](https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=vs-2019). You may want to use `_M_IX86` in addition for testing x86 targets.


sjaeckel:
@karel-m @nijtmans you're the two windows guys I'm aware of, this looks fine, right?

sjaeckel:
Thanks for the PR! Merged.

Issue #486, opened 2020-07-21
mp_sqrtmod_prime can get caught in an infinite loop if "prime" isn't prime


friedrichsenm:
If `*prime` isn't prime but is composite that is 1 mod 4, you can hit infinite loops in two places. (My use case only involves numbers 1 mod 4, so this may happen other times too)

The first is in lines 62-69
```C
mp_set(&Z, 2uL);
   /* Z = 2 */
   for (;;) {
      if ((err = mp_kronecker(&Z, prime, &legendre)) != MP_OKAY)     goto LBL_END;
      if (legendre == -1) break;
      if ((err = mp_add_d(&Z, 1uL, &Z)) != MP_OKAY)               goto LBL_END;
      /* Z = Z + 1 */
   }
```

I was able to fix this in a local copy adding the following check after line 66
```C
if (legendre == 0) {err = MP_VAL; goto LBL_END;}
```

The second infinite loop was in lines 84-111
```C
for (;;) {
      if ((err = mp_copy(&T, &t1)) != MP_OKAY)                    goto LBL_END;
      i = 0;
      for (;;) {
         if (mp_cmp_d(&t1, 1uL) == MP_EQ) break;
         if ((err = mp_exptmod(&t1, &two, prime, &t1)) != MP_OKAY) goto LBL_END;
         i++;
      }
      if (i == 0u) {
         if ((err = mp_copy(&R, ret)) != MP_OKAY)                  goto LBL_END;
         err = MP_OKAY;
         goto LBL_END;
      }
      if ((err = mp_sub_d(&M, i, &t1)) != MP_OKAY)                goto LBL_END;
      if ((err = mp_sub_d(&t1, 1uL, &t1)) != MP_OKAY)             goto LBL_END;
      if ((err = mp_exptmod(&two, &t1, prime, &t1)) != MP_OKAY)   goto LBL_END;
      /* t1 = 2 ^ (M - i - 1) */
      if ((err = mp_exptmod(&C, &t1, prime, &t1)) != MP_OKAY)     goto LBL_END;
      /* t1 = C ^ (2 ^ (M - i - 1)) mod prime */
      if ((err = mp_sqrmod(&t1, prime, &C)) != MP_OKAY)           goto LBL_END;
      /* C = (t1 * t1) mod prime */
      if ((err = mp_mulmod(&R, &t1, prime, &R)) != MP_OKAY)       goto LBL_END;
      /* R = (R * t1) mod prime */
[...]

PR #487, opened 2020-08-05
Added checks for input in mp_sqrtmod_prime that caused infinite loops

czurnieden:
Bugfix regarding issue #486.

I cannot imagine how that issue might be exploited but because I cannot imagine it doesn't mean there is *no* way at all.

 - check that parameter `prime` is odd and `> 2`
 - the check if parameter `prime` is congruent to 3 mod 4 is done at bit-level now (which makes that check constant time, too)
 - variables `M`, `s`, and `i` changed to native signed integer `int` (values cannot exceed `INT_MAX` )
 - ( #486 ) check for `(Z|prime) == 0` in the first loop and return `MP_VAL` in that case. Although if `prime` (p) is an odd prime `Jacobi(Z|p) == 0` for `Z \cong 0 (mod p)` there is at least one non-quadratic residue before `Z>=p` if `p` is an odd prime, hence no false positives. 
 - ( #486 )  check condition `M < i` in the loop searching for smallest `0 < i < M` for  `t^(2^i)`

The extra checks are cheap and the function itself should even be a little bit faster together with less need for stack and heap with the other changes.

I did not pick any reviewer explicitly because of the current situation and its accompanying mess but I would appreciate it highly if somebody would take a look. Thank's!

### DIFF
--- a/mp_sqrtmod_prime.c
+++ b/mp_sqrtmod_prime.c
@@ -13,19 +13,23 @@ mp_err mp_sqrtmod_prime(const mp_int *n, const mp_int *prime, mp_int *ret)
 {
    mp_err err;
    int legendre;
-   mp_int t1, C, Q, S, Z, M, T, R, two;
-   mp_digit i;
+   /* The type is "int" because of the types in the mp_int struct.
+      Don't forget to change them here when you change them there! */
+   int S, M, i;
+   mp_int t1, C, Q, Z, T, R, two;
 
    /* first handle the simple cases */
    if (mp_cmp_d(n, 0uL) == MP_EQ) {
       mp_zero(ret);
       return MP_OKAY;
    }
-   if (mp_cmp_d(prime, 2uL) == MP_EQ)                            return MP_VAL; /* prime must be odd */
-   if ((err = mp_kronecker(n, prime, &legendre)) != MP_OKAY)        return err;
-   if (legendre == -1)                                           return MP_VAL; /* quadratic non-residue mod prime */
+   /* "prime" must be odd and > 2 */
+   if (mp_iseven(prime) || (mp_cmp_d(prime, 3uL) == MP_LT))      return MP_VAL;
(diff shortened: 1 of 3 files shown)


sjaeckel (approved):
LGTM

@friedrichsenm can you please verify

mp_sqrtmod_prime.c
```diff
@@ -13,28 +13,30 @@ mp_err mp_sqrtmod_prime(const mp_int *n, const mp_int *prime, mp_int *ret)
...
    /* first handle the simple cases */
    if (mp_cmp_d(n, 0uL) == MP_EQ) {
       mp_zero(ret);
       return MP_OKAY;
    }
-   if (mp_cmp_d(prime, 2uL) == MP_EQ)                            return MP_VAL; /* prime must be odd */
-   if ((err = mp_kronecker(n, prime, &legendre)) != MP_OKAY)        return err;
+   if (mp_iseven(prime) || (mp_cmp_d(prime, 3uL) == MP_LT))      return MP_VAL; /* "prime" must be odd and > 2 */
+   if ((err = mp_kronecker(n, prime, &legendre)) != MP_OKAY)     return err;
    if (legendre == -1)                                           return MP_VAL; /* quadratic non-residue mod prime */
```
friedrichsenm, line 28: Presumably a user might check that gcd(n, prime) = 1 before running this function, but might be better to change this line to
```C
if (legendre != 1) return MP_VAL;
```

friedrichsenm (changes requested):
I have one question about a kronecker symbol check in mp_sqrtmod_prime.c. When checking (n | prime) it might make sense to check instead that legendre != 1 instead of legendre == -1 to handle the case when it is 0.

czurnieden:
> When checking (n | prime) it might make sense to check instead that legendre != 1 instead of legendre == -1 to handle the case when it is 0.

Yepp, missed that. Good find, thanks!

friedrichsenm (approved):

Issue #488, opened 2020-09-11
Build shared libraries using Visual Studio


madebr:
The Visual Studio makefile does not support building shared libraries.
https://github.com/libtom/libtommath/blob/44ee82cd34d0524c171ffd0da70f83bba919aa38/makefile.msvc#L20

Would a patch adding this be accepted?

sjaeckel:
sure!

hint: it will be accepted even faster if the PR also adds corresponding CI jobs

ping @karel-m 

I'm closing this and wait for the PR!

madebr:
I will open one in the coming days.
I suppose the exports in [tommath.def](https://github.com/libtom/libtommath/blob/develop/tommath.def) will be kept up to date?

madebr:
Creating the implib+dll is not hard, but linking the test to the dll is harder.
After adding some missing exports to `tommath.def`, linking the test to the dll fails.
This is because the test is using private symbols.
```
        cl /nologo /I./ /D_CRT_SECURE_NO_WARNINGS /D_CRT_NONSTDC_NO_DEPRECATE /D__STDC_WANT_SECURE_LIB__=1 /D_CRT_HAS_CXX17=0 /Wall /wd4146 /wd4127 /wd4668 /wd4710 /wd4711 /wd4820 /wd5045 /WX /Ox  tommath.dll.lib  advapi32.lib demo/shared.c demo/test.c /Fetest_dll.exe
shared.c
test.c
Generating Code...
test.obj : error LNK2019: unresolved external symbol s_mp_div_3 referenced in function very_random_source
test.obj : error LNK2019: unresolved external symbol s_mp_div_recursive referenced in function s_rs
test.obj : error LNK2019: unresolved external symbol s_mp_div_school referenced in function s_rs
test.obj : error LNK2019: unresolved external symbol s_mp_div_small referenced in function s_rs
test.obj : error LNK2019: unresolved external symbol s_mp_mul referenced in function s_rs
test.obj : error LNK2019: unresolved external symbol s_mp_mul_balance referenced in function s_rs
test.obj : error LNK2019: unresolved external symbol s_mp_mul_karatsuba referenced in function s_rs
test.obj : error LNK2019: unresolved external symbol s_mp_mul_toom referenced in function s_rs
test.obj : error LNK2019: unresolved external symbol s_mp_sqr referenced in function s_rs
test.obj : error LNK2019: unresolved external symbol s_mp_sqr_karatsuba referenced in function s_rs
[...]

PR #489, opened 2020-09-11
enable building dll's using makefile.msvc

madebr:
- This pr enables building a dll with `makefile.msvc`
- the dll + static library are built
- the tests are **not** built for the shared libraries, because the tests use private symbols (that are not exported)
See https://github.com/libtom/libtommath/issues/488#issuecomment-691113893 for the link error.

@sjaeckel @nijtmans

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,9 @@
 *.gcda
 *.gcno
 *.gcov
+*.dll
+*.exp
+*.pdb
 *.lib
 *.tmp
 [Dd]ebug/
--- a/makefile.msvc
+++ b/makefile.msvc
@@ -12,13 +12,16 @@
 #The following can be overridden from command line e.g. make -f makefile.msvc CC=gcc ARFLAGS=rcs
 PREFIX    = c:\devel
 CFLAGS    = /Ox
+LDFLAGS   =
 
 #Compilation flags
 LTM_CFLAGS  = /nologo /I./ /D_CRT_SECURE_NO_WARNINGS /D_CRT_NONSTDC_NO_DEPRECATE /D__STDC_WANT_SECURE_LIB__=1 /D_CRT_HAS_CXX17=0 /Wall /wd4146 /wd4127 /wd4668 /wd4710 /wd4711 /wd4820 /wd5045 /WX $(CFLAGS)
-LTM_LDFLAGS = advapi32.lib
+LTM_LDFLAGS = $(LDFLAGS) advapi32.lib
 
-#Libraries to be created (this makefile builds only static libraries)
-LIBMAIN_S =tommath.lib
+#Libraries to be created
+LIBMAIN_S = tommath.lib
+LIBMAIN_I = tommath.dll.lib
+LIBMAIN_D = tommath.dll
 
 #List of objects to compile (all goes to tommath.lib)
 OBJECTS=mp_2expt.obj mp_abs.obj mp_add.obj mp_add_d.obj mp_addmod.obj mp_and.obj mp_clamp.obj mp_clear.obj mp_clear_multi.obj \
@@ -62,12 +65,16 @@ $(OBJECTS): $(HEADERS)
 $(LIBMAIN_S): $(OBJECTS)
 	lib /out:$(LIBMAIN_S) $(OBJECTS)
 
+#Create DLL + import library tommath.dll.lib
+$(LIBMAIN_D) $(LIBMAIN_I): $(OBJECTS) tommath.def
+	link /dll /out:$(LIBMAIN_D) /implib:$(LIBMAIN_I) /def:tommath.def $(LTM_LDFLAGS) $(OBJECTS)
+
 #Build test suite
 test.exe: $(LIBMAIN_S) demo/shared.obj demo/test.obj
 	cl $(LTM_CFLAGS) $(TOBJECTS) $(LIBMAIN_S) $(LTM_LDFLAGS) demo/shared.c demo/test.c /Fe$@
 	@echo NOTICE: start the tests by launching test.exe
 
-all: $(LIBMAIN_S) test.exe
+all: $(LIBMAIN_S) test.exe $(LIBMAIN_D)
 
 tune: $(LIBMAIN_S)
 	$(MAKE) -C etc tune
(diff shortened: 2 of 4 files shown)


sjaeckel:
Did you read the comment in `s_mp_rand_platform.c` before the declaration of `s_mp_rand_platform()`? :)

madebr:
> `s_mp_rand_platform`

The reason I added these empty declarations is that when building a debug dll, the linker complains.
Running:
```
nmake -f makefile.msvc tommath.dll CFLAGS=-Zi LDFLAGS=-debug
```
causes this error to be thrown when creating the dll:
```
  Creating library tommath.dll.lib and object tommath.dll.exp
s_mp_rand_platform.obj : error LNK2019: unresolved external symbol s_read_arc4random referenced in function s_mp_rand_platform
s_mp_rand_platform.obj : error LNK2019: unresolved external symbol s_read_getrandom referenced in function s_mp_rand_platform
s_mp_rand_platform.obj : error LNK2019: unresolved external symbol s_read_urandom referenced in function s_mp_rand_platform
```

madebr:
travis ci is failing because I have added some symbols to `tommath.def`, which are not automaticallly added by `update.pl`.
To fix this, I will have to create 5 files names:
- mp_rand_source.c
-    MP_MUL_KARATSUBA_CUTOFF.c
-    MP_SQR_KARATSUBA_CUTOFF.c
-    MP_MUL_TOOM_CUTOFF.c
-    MP_SQR_TOOM_CUTOFF.c

What do you think?

madebr:
Not adding implementations and depending on the compiler to optimize away impossible function calls, seems wrong to me.
I added a build script for libtommath at conan-center-index.
See https://github.com/conan-io/conan-center-index/pull/2904#issuecomment-691671889
It also fails for macos.

sjaeckel:
> Not adding implementations and depending on the compiler to optimize away impossible function calls, seems wrong to me.

I was also skeptical whether relying on this functionality is good, but nowadays I write code in the same style in my professional life. It has more advantages than disadvantages.

> I added a build script for libtommath at conan-center-index.

cool, whatever this 'conan' is :)

> It also fails for macos.

That's possible, as macos is not something we're testing regularly against.

How can I see a summary of what failed?

madebr:
> > Not adding implementations and depending on the compiler to optimize away impossible function calls, seems wrong to me.
> 
> I was also skeptical whether relying on this functionality is good, but nowadays I write code in the same style in my professional life. It has more advantages than disadvantages.

It's easier to use and code, I understand.
But not so robust in the face of the variety of compilers.
For e.g. MSVC it requires at least `-O1`, `-O0` would fail.

> > I added a build script for libtommath at conan-center-index.
> 
> cool, whatever this 'conan' is :)

conan is a package manager for c/c++, and conan-center-index is a library for useful comon libraries.
Think of it as a pypi for c/c++.
> > It also fails for macos.
> 
> That's possible, as macos is not something we're testing regularly against.
> 
> How can I see a summary of what failed?

I've fixed the la error.
See https://github.com/conan-io/conan-center-index/pull/2904#issuecomment-691674157 for the latest CI output.
The CI is a wip, so no nice table (yet).

sjaeckel:
> But not so robust in the face of the variety of compilers.
> For e.g. MSVC it requires at least `-O1`, `-O0` would fail.

TBH who really cares about MSVC? I don't ... support for it is a nice goodie, but to mess up code for this compiler - no way!

> conan is a package manager for c/c++

nice!

> (Don't forget these have the patch applied to bn_s_mp_rand_platform.c)

and sorry if our decision that MSVC support is only secondary makes additional work you :)

I guess if you have to use MSVC and this poses problems to you, you will have to maintain those patches.

IIRC @nijtmans builds Windows libraries with GCC, which then also have full 64bit support which MSVC is lacking!

madebr:
> TBH who really cares about MSVC? I don't ... support for it is a nice goodie, but to mess up code for this compiler - no way!

I understand your opinion, but as you see it's failing on clang too.
And it's always nice to have more test coverage.

> 
> I guess if you have to use MSVC and this poses problems to you, you will have to maintain those patches.

That's always a risk when packaging things :smile: . 
(Please don't change too much in your next release :rofl: )

> 
> IIRC @nijtmans builds Windows libraries with GCC, which then also have full 64bit support which MSVC is lacking!

What missing 64-bit support are you referring to?

sjaeckel:
> I understand your opinion, but as you see it's failing on clang too.

hmm... I remember having seen this on clang as well, but I can't reproduce on my machine.

```
$ clang --version
clang version 10.0.1 
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
$ CC=clang make -f makefile.shared -j9 IGNORE_SPEED=1                                                                                                                   makefile.shared:56: warning: ignoring prerequisites on suffix rule definition                                                                                                                                                                        
libtool --mode=compile --tag=CC clang -I./ -Wall -Wsign-compare -Wextra -Wshadow -Wdeclaration-after-statement -Wbad-function-cast -Wcast-align -Wstrict-prototypes -Wpointer-arith -Wsystem-headers -Wno-typedef-redefinition -Wno-tautological-compare -Wno-builtin-requires-header   -o mp_2expt.o -c mp_2expt.c     
...
libtool: link: ranlib .libs/libtommath.a
libtool: link: ( cd ".libs" && rm -f "libtommath.la" && ln -s "../libtommath.la" "libtommath.la" )
```

> What missing 64-bit support are you referring to?

The fact that MSVC has no 128bit data type for `mp_word` ... which leads to libtommath with MSVC getting compiled as 32bit for 64bit Windows. If you want the full (64/128bit) experience you have to use a proper compiler ;)

madebr:
> hmm... I remember having seen this on clang as well, but I can't reproduce on my machine.

It also doesn't occur on my machine, running clang 9.
I'm not running ubuntu, but 20.04 has clang-6.0.

> The fact that MSVC has no 128bit data type for `mp_word` ... which leads to libtommath with MSVC getting compiled as 32bit for 64bit Windows. If you want the full (64/128bit) experience you have to use a proper compiler ;)

I don't mind some friendly competition :smile:

MasterDuke17:
FWIW, MoarVM carries https://github.com/MoarVM/libtommath/commit/628cf8407a49af620197af278dbf00c210deac64 around on top of libtommath because we use some debug builds in our CI testing.

sjaeckel:
> It also doesn't occur on my machine, running clang 9.
> I'm not running ubuntu, but 20.04 has clang-6.0.

so it's clear that it's a compiler bug that has since been fixed in clang and the only compiler that is still buggy is MSVC.

madebr:
@MasterDuke17 
> FWIW, MoarVM carries [MoarVM@628cf84](https://github.com/MoarVM/libtommath/commit/628cf8407a49af620197af278dbf00c210deac64) around on top of libtommath because we use some debug builds in our CI testing.

Your patch is a lot easier. Can I borrow it for the cci patch? With attribution of course.

> so it's clear that it's a compiler bug that has since been fixed in clang and the only compiler that is still buggy is MSVC.

Some things never change.

MasterDuke17:
> Your patch is a lot easier. Can I borrow it for the cci patch? With attribution of course.

Sure. Don't worry about attribution, and now that you mention it, I think I cribbed it from an earlier version of libtommath itself.

sjaeckel:
Can this be closed in favor of #490 ?

madebr:
Of course, your pr is building on top of this one.

PR #490, opened 2020-09-13
Follow-up of #489

sjaeckel:
This is a follow-up of #489 

It adds "automation" of the modifications to `tommath.def` and a travis job to check for this.

Additionally the proposed split-up of `mp_rand.c` is done and the modifications to `s_mp_rand_platform.c` are reverted.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,9 @@
 *.gcda
 *.gcno
 *.gcov
+*.dll
+*.exp
+*.pdb
 *.lib
 *.tmp
 [Dd]ebug/
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,6 +72,13 @@ matrix:
           packages:
             - astyle
 
+    # Check public symbols of dynamic libraries
+    - env: BUILDOPTIONS='--symbols'
+      addons:
+        apt:
+          packages:
+            - libtool-bin
+
     # Run always with valgrind (no sanitizer, but debug info)
     - env: COMPILE_DEBUG=1 BUILDOPTIONS='--with-cc=gcc-4.9 --with-m64 --with-valgrind'
       addons:
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,8 @@
 #                                                                           #
 #############################################################################
 
-# Run the tests based on Ubuntu 16.04
-dist: xenial
+# The Ubuntu version we're going to use to run the tests
+dist: bionic
 
 # Compilation failures are in gcc_errors_*.log
 # Failed tests in test_*.log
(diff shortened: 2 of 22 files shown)


madebr:
FYI, this are the results of building libtommath on linux gcc, linux clang, macos and windows msvc: https://github.com/conan-io/conan-center-index/pull/2904#issuecomment-691671889
Don't mind the .la errors, I forgot to remove these.

czurnieden (approved):
I can't tell a lot about the Windows related changes, but the rest I like!

madebr:
With the following changes, this pr allows building the tests linked to the dll.
```patch
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -19,3 +19,4 @@ build_script:
         nmake -f makefile.msvc all
 test_script:
 - cmd: test.exe
+- cmd: test_dll.exe
--- a/makefile.msvc
+++ b/makefile.msvc
@@ -71,10 +71,15 @@ $(LIBMAIN_D) $(LIBMAIN_I): $(OBJECTS) tommath.def

 #Build test suite
 test.exe: $(LIBMAIN_S) demo/shared.obj demo/test.obj
-       cl $(LTM_CFLAGS) $(TOBJECTS) $(LIBMAIN_S) $(LTM_LDFLAGS) demo/shared.c demo/test.c /Fe$@
+       link $(TOBJECTS) $(LTM_LDFLAGS) $? /out:$@
        @echo NOTICE: start the tests by launching test.exe

-all: $(LIBMAIN_S) test.exe $(LIBMAIN_D)
+#Build test suite for dll
+test_dll.exe: $(LIBMAIN_I) demo/shared.obj demo/test.obj
+       link $(TOBJECTS) $(LTM_LDFLAGS) $? /out:$@
+       @echo NOTICE: start the tests by launching test_dll.exe
+
+all: $(LIBMAIN_S) test.exe $(LIBMAIN_D) test_dll.exe

 tune: $(LIBMAIN_S)
        $(MAKE) -C etc tune
```

madebr:
The tests was failing in `test_mp_fread_fwrite`.
It looks like we need to override the default runtime when building dll's.
By default static runtime is used, but that does not work because the `mp_f*` api calls pass `FILE*` handles.
So use the shared `MD` runtime by default instead.
People can override the default by setting it in `CFLAGS`.

madebr:
FYI, defining `LTM_TEST_DYNAMIC` is not absolutely required for msvc if you add `/O1` (or higher) to `CFLAGS` and add `/OPT:REF,ICF` to `LDFLAGS` (when builiding in debug mode).

czurnieden (approved):
Ubuntu's Bionic Beaver is version 18.04, you might change the comment above the distribution settings in `.travis.yml` otherwise: looks good for me!

PR #491, opened 2020-11-06
Added methods to convert radices that are powers of two to and from strings in linear time.

czurnieden:
As it is known that radix conversion can be done in O(n) I gave it a try. Normally a simple thing but with the quite weird architecture of LTM it took a bit more effort and code.

The test `test_s_mp_radix_power_of_two` in `test.c` starting at line 1013 tests all bases via round-trips (string -> big-int -> string) although with rather small strings. This can either replace the old tests or, after stripping every other base except powers of two, kept for the new methods only. Testing the new functions stand-alone is doable but would need the old (stripped down) versions of `to_radix` and `read_radix` to compare against.

I did not run any benchmarks yet, but the constant in O(n) is rather small, so I don't think a cut-off would make much sense. Both binaries together are 4k, stripped, with MP_64BIT--less with smaller architectures.

It is working but still a bit rough around the edges, see code for details.

Before you are asking: the implementation of the D&C algorithm with [Lars Helmström's trick](https://sourceforge.net/p/tcl/mailman/message/31972670/) does not work as intended: some of the computations start to overflow at around MP_MAX_DIGIT_COUNT/2 for some bases including base 10. The computation of the Newton-Raphson round is also about as expensive as computing the `2^(2*k)/n` reciprocals individually (there are only `log(log(n))` reciprocals needed but still) just more complicated. This might need a day or ten to implement and I didn't want to wait that long, hence this "partial" commit.

Because of the still ... uhm ...  turbulent times I refrained from pinning reviewers explicitly.

(I just saw that I have two typos in my commit message. Two! How embarrassing!)

### DIFF
--- a/demo/test.c
+++ b/demo/test.c
@@ -1373,14 +1373,15 @@ static int test_s_mp_radix_power_of_two(void)
    char *buf_n;
    mp_int N;
    size_t len, written;
+   int number_base, number, radix;
 
    DOR(mp_init_multi(&N, NULL));
 
-   for (int number_base = 0; number_base < 63; number_base++) {
-      for (int number = 0; number < (int)(sizeof(numbers[0])/sizeof(numbers[0][0])); number++) {
+   for (number_base = 0; number_base < 63; number_base++) {
+      for (number = 0; number < (int)(sizeof(numbers[0])/sizeof(numbers[0][0])); number++) {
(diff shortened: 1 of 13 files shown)


czurnieden:
To complicated, too much code, I can do better.

PR #495, opened 2020-12-20
New and expanded log n

czurnieden:
**EDIT:** Because of too much organisational mess in this push (I somehow messed up the rebase to develop completely) I made a new one at #497 and  this pull request will be deleted in ca 24 hours. Sorry for that!

My old method suffers from OVF errors that are not salvageable, at least not easily. Replaced it with a very simple one where the overflow can be caught (hard to explain, see code). It is also possible now to use a bigint as the base. I hope I didn't mess up as much as I do regularly.

The proof is in the pudding.

EDIT: just found out, that I did: used `mp_log_u32` instead of `log_n`. \*grr\*


czurnieden:
Replaced with #497

PR #498, opened 2020-12-24
Added calculation of the number of Miller-Rabin rounds necessary for DEA primes if done with M-R alone.

czurnieden:
The calculation of the number of rounds for DEA primes have been explained in the documentation but the code itself is still misleading as it seems. I added two functions: `mp_prime_rabin_miller_trial_dea` to compute the number of rounds for DSA primes with M-R alone and `mp_prime_rabin_miller_trial_rsa` to make the logic a bit more symmetric. I just realized that I did not deprecate the original function `mp_prime_rabin_miller_trial` yet, but it should if this PR gets accepted.

### DIFF
--- a/mp_prime_rabin_miller_trials_dea.c
+++ b/mp_prime_rabin_miller_trials_dea.c
@@ -1,0 +1,18 @@
+#include "tommath_private.h"
+#ifdef MP_PRIME_RABIN_MILLER_TRIALS_DEA_C
+/* LibTomMath, multiple-precision integer library -- Tom St Denis */
+/* SPDX-License-Identifier: Unlicense */
+
+/* Compute number of Miller-Rabin rounds necessary for DSA primes
+   if only Miller-Rabin tests were used. Use
+      mp_prime_rabin_miller_trials(size)
+   or
+      mp_prime_rabin_miller_trials_rsa(size)
+   if the M-R tests are followed by a Lucas test.
+ */
+int mp_prime_rabin_miller_trials_dea(int error)
+{
+   return ((error/2) + 1);
+}
+
+#endif
--- a/doc/bn.tex
+++ b/doc/bn.tex
@@ -2031,11 +2031,11 @@ to help out.
 \begin{alltt}
 mp_err mp_prime_rabin_miller_trials(int size)
 \end{alltt}
-This returns the number of trials required for a low probability of failure for a given
-\texttt{size} expressed in bits.  This comes in handy specially since larger numbers are slower to
-test. For example, a 512--bit number would require 18 tests for a probability of $2^{-160}$ whereas
-a 1024--bit number would only require 12 tests for a probability of $2^{-192}$. The exact values as
-implemented are listed in table \ref{table:millerrabinrunsimpl}.
+The function returns the number of trials
+required for a low probability of failure for a given \texttt{size} expressed in bits. This comes in
+handy specially since larger numbers are slower to test. For example, a 512--bit number would require
+18 tests for a probability of $2^{-160}$ whereas a 1024--bit number would only require 12 tests for a
+probability of $2^{-192}$. The exact values as implemented are listed in table \ref{table:millerrabinrunsimpl}.
 
 \begin{table}[h]
   \begin{center}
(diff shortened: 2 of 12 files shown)


mp_prime_rabin_miller_trials_dea.c
```diff
@@ -0,0 +1,18 @@
...
+/* SPDX-License-Identifier: Unlicense */
+
+/* Compute number of Miller-Rabin rounds necessary for DSA primes
+   if only Miller-Rabin tests were used. Use
+      mp_prime_rabin_miller_trials(size)
+   or
+      mp_prime_rabin_miller_trials_rsa(size)
+   if the M-R tests are followed by a Lucas test.
+ */
+int mp_prime_rabin_miller_trials_dea(int error)
```
sjaeckel, line 13: I don't like the fact that the only difference between this API and the RSA one is the variable name, but the behavior is totally different ... and also the input is expected to be something entirely different...

How is the input to this function supposed to be calculated? (or is this supposed to be fixed depending on the probability that is supposed to be used?)
Would it maybe make sense to make this API `int mp_prime_rabin_miller_trials_dea(int size, int error)`?
And maybe to also take into account whether `LTM_USE_ONLY_MR` resp. `LTM_USE_FROBENIUS_TEST` are defined?

czurnieden: > I don't like the fact that the only difference between this API and the RSA one is the name of the argument, but the behavior is totally different ... and also the input is expected to be something entirely different...

Yes, naming things was never my greatest forte.

The function name was already quite long to begin with and I didn't want to add too much to it but it is of course possible to make it a bit more "talkative". I'm wide open for suggestions.

> How is the input to this function supposed to be calculated?

The documentation for this function is short, admitted, is it too short?
(And it has a typo \*sigh\*)

> (or is this supposed to be fixed depending on the probability that is supposed to be used?)

It depends on the absolute value of the exponent of the binary representation of the probability the user wants. It is half of that, rounded to positive infinity in case of a tie. According to Fips. It is independent of the size of the prime. Also according to Fips.
So&hellip;

> Would it maybe make sense to make this API `int mp_prime_rabin_miller_trials_dea(int size, int error)`?

&hellip;would make no sense. But&mdash;and here it gets interesting&mdash;only if the tests use M-R only. If you follow your M-R tests by a Lucas test you can use the values for RSA calculated with that complicated formula. According to Fips.
So&hellip;

> And maybe to also take into account whether `LTM_USE_ONLY_MR` resp. `LTM_USE_FROBENIUS_TEST` are defined?

&hellip;is tempting but we don't know if the Frobenius test is used, even if `LTM_USE_FROBENIUS_TEST` is defined (M-R is a public function and many, way too many roll their own). On the other side `LTM_USE_ONLY_MR` would be useful but only if the RSA method is the default and the RSA method is only the default for computing RSA trials.

Now that I wrote that: yes it is a very good idea to check for `LTM_USE_ONLY_MR` in computing RSA trials and returning the DEA computation (half the error as described above) in that case. RSA error depends also on the size of the prime in the table and we can spend that table another column for little money.

It is not that easy for DEA but I think that you should not automate everything even if you can. Some things are better left to the documentation.

BTW: The Frobenius test can and should go, it was for MP_8BIT only. It is not known if it is independent of the Strong-Lucas-Test and was only meant as a replacement for the Strong-Lucas-Test. The Extra-Strong-Lucas-Test is said to be independent of the S-L-T for some parameters different from the parameters used in the implementation by Thomas Nicely but I haven't found time to verify that properly (Just because Wikipedia said so? Nope!).

czurnieden: > Now that I wrote that: yes it is a very good idea to check for LTM_USE_ONLY_MR in computing RSA trials and returning the DEA computation (half the error as described above) in that case.

No, that is nonsense. The table *is* for M-R-only. (see e.g.: [Damgard et. al.](http://www.math.dartmouth.edu/~carlp/PDF/paper88.pdf) or [Burthe](http://www.ams.org/journals/mcom/1996-65-213/S0025-5718-96-00695-3/S0025-5718-96-00695-3.pdf))

Let me see if I can remove that commit accident-free.

czurnieden:
This seems to be obsoleted by #541 , too.
Feel free to reopen if that is not the case.

PR #500, opened 2020-12-28
Replaced "fgets" with a "get_token" function in demo/mtest_opponent.c

czurnieden:
We shouldn't leave the clean-up of the mtest-tokens to `mp_read_radix`. This PR would also allow to get rid of the abominable input-sanitizing in #499

### DIFF
--- a/demo/mtest_opponent.c
+++ b/demo/mtest_opponent.c
@@ -11,10 +11,49 @@ static void draw(const mp_int *a)
    ndraw(a, "");
 }
 
-#define FGETS(str, size, stream) \
+/*
+   Get tokens. It is just a very(!) simple fgets(3) that does not keep line endings.
+
+   Implementation follows along "man 3 fgets", some of which is quoted.
+ */
+static char *s_mp_get_token(char *s, int size, FILE *stream)
+{
+   char *s_bar = s;
+   int c;
+   bool eol_hit = false;
+
+   /* "fgets [...] reads in at most one less than size characters from stream"  */
+   while (--size)  {
+      /* "Reading stops after an EOF or a newline." We stop only for EOF here */
+      if ((c = fgetc(stream)) == EOF) {
+         /* "Returns [...] NULL on error or when end of file occurs while no characters have been read" */
+         if ((s_bar == s) || (ferror(stream) != 0)) {
+            return NULL;
+         }
+         break;
+      }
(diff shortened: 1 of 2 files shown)


sjaeckel (changes requested):
Tests broken, something like the following "solves the problem"^TM (not sure what other problems it introduces ;) )
```$  PAGER= git diff -U1
diff --git a/demo/mtest_opponent.c b/demo/mtest_opponent.c
index 9f80e65..dbf8745 100644
--- a/demo/mtest_opponent.c
+++ b/demo/mtest_opponent.c
@@ -21,3 +21,3 @@ static char *s_mp_get_token(char *s, int size, FILE *stream)
    char *s_bar = s;
-   int c;
+   int c, eol_hit = 0;
 
@@ -35,4 +35,9 @@ static char *s_mp_get_token(char *s, int size, FILE *stream)
       if ((c == '\n')  || (c == '\r')) {
+         eol_hit = 1;
          continue;
       }
+      if (eol_hit) {
+         ungetc(c, stream);
+         break;
+      }
       *s_bar++ = c;
@@ -110,3 +115,2 @@ static int mtest_opponent(void)
       GET_TOKEN(cmd, 4095, stdin);
-      cmd[strlen(cmd) - 1u] = '\0';
       printf("%-6s ]\r", cmd);
```

czurnieden:
> Tests broken

I always try not to leave my PRs broken but Travis took way too long to start that day^Wnight so I went to bed. Sorry!
(Why didn't it come up when I run `--test-vs-mtest` locally? That's annoying!)

> not sure what other problems it introduces

Good question. I'll take a deeper look this time.

Thanks!

czurnieden:
Does seem to work, now that `s_mp_get_token` stops at the right time.
Spot-checked input/output (If the names gets over correctly and the numbers for `sub_d`).

Why the local tests did fail before PR is still a mystery to me, because now they do. But I also forgot to remove `cmd[strlen(cmd) - 1u] = '\0';`, so it was most likely me to blame here.

There is still a non-technical problem left: because of the age of this part (`read_radix`) of LTM there might be many uses of the undocumented feature of `read_radix` to accept linebreaks at the end of the input string. This PR won't break them but if this PR gets included I can and will remove the input cleaning in #499 which *will* break those builds.
I admit I'm a bit torn here.

sjaeckel (approved):

PR #502, opened 2021-01-07
Raised upper limit for Comba squaring

czurnieden:
This PR replaces the comba-squaring cut-off `MP_MAX_COMBA/2` with `MP_MAX_COMBA`.

Saw @tomstdenis taking part in the discussion over at #429 lately and remembered the curious results for the faster squaring algorithms (Karatsuba and Toom-Cook-3) where the cut-offs for both coincided with the upper limit of comba-squaring but not for comba-multiplying.

The upper limit for comba-multiplying is the size of `min(a.used,b.used)` but for comba-squaring it is `a.used/2`. Neither the actual code nor tests with inputs of the form `-1 + 2^(n*MP_DIGIT_BIT)` (all bits set in every limb with `MP_MASK`) for all `MP_16BIT`, `MP_32BIT`, and `MP_64BIT` showed a reason for that difference.

Also: running `tune` gives reasonable cut-offs now for the Karatsuba and Toom-Cook-3 squaring algorithms.

PR #503, opened 2021-01-09
optimized s_mp_sqr

czurnieden:
`s_mp_sqr` is significantly slower than `s_mp_mul` here (without comba). Changing the addition `r + r` to `r<<1` made it already faster and changing the computation in the carry propagating loop to a manual multi-word doubled the speed-up already gained. The relation between `s_mp_mul` and `s_mp_sqr` follows the theoretical runtimes much closer now.
A test for `s_mp_sqr` against `s_mp_mul` has been added to `etc/test.c`.

Those advantages depend on compiler and/or architectur, of course, but the changes should not make it worse.

diff_original_vs_new_s_mp_sqr_vs_s_mp_mul

### DIFF
--- a/demo/test.c
+++ b/demo/test.c
@@ -1846,6 +1846,33 @@ LTM_ERR:
    return err;
 }
 
+static int test_s_mp_sqr(void)
+{
+   mp_int a, b, c;
+   int i;
+
+   DOR(mp_init_multi(&a, &b, &c, NULL));
+
+   /* s_mp_mul() has a hardcoded branch to s_mul_comba if s_mul_comba is available,
+      so test another 10 just in case. */
+   for (i = 1; i < MP_MAX_COMBA + 10; i++) {
+      DO(s_fill_with_ones(&a, i));
+      DO(s_mp_sqr(&a, &b));
+      DO(s_mp_mul(&a, &a, &c, 2*i + 1));
+      EXPECT(mp_cmp(&b, &c) == MP_EQ);
+      DO(mp_rand(&a, i));
+      DO(s_mp_sqr(&a, &b));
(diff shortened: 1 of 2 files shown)


sjaeckel (commented):
Thanks for reminding me of those PR's!

I rebased after merging #502 and fixed the merge conflict ... Does that merge make sense? Both tests `test_s_mp_sqr()` and `test_s_mp_sqr_comba()` do something slightly different, or should there only be one?

czurnieden:
> I rebased after merging https://github.com/libtom/libtommath/pull/502 and fixed the merge conflict ... Does that merge make sense?

As far as I can see: yes.

> Both tests `test_s_mp_sqr()` and `test_s_mp_sqr_comba()` do something slightly different, or should there only be one?

The outer loops are different, that is intentional but `test_s_mp_sqr_comba()` only works if Comba is switched off at compile time. Add a round to the CI-tests?

And, no, I don't have the faintest idea anymore why I added that small inner loop in `test_s_mp_sqr_comba()`. The known edge case is the main point, one test at a random point for the unknown edge cases should suffice.

sjaeckel:
> Add a round to the CI-tests?
> 
> And, no, I don't have the faintest idea anymore why I added that small inner loop in `test_s_mp_sqr_comba()`. The known edge case is the main point, one test at a random point for the unknown edge cases should suffice.

you wanna update these tests or should we merge as is?

czurnieden:
> you wanna update these tests or should we merge as is?

Merge as it is, because I know me and if I start fiddling around I will find this&hellip;and that&hellip;and those&hellip;and at the end a week is gone ;-)

Issue #509, opened 2021-07-10
Build fails on Linux/Sparc with 32 bit userspace


DerDakon:
[Downstream Gentoo bug](https://bugs.gentoo.org/781938)

```
In file included from tommath_private.h:7,
                 from bn_cutoffs.c:1:
tommath.h:80:1: error: unable to emulate ‘TI’
   80 | typedef unsigned long        private_mp_word __attribute__((mode(TI)));
      | ^~~~~~~
```

czurnieden:
We can guard those lines in `tommath.h` with an extra check for `__SIZEOF_INT128__` or you can (temporarily) add ` -DMP_32BIT ` to your CFLAGS (all of them. Those that compile LibTomMath and all others that use LibTomMath for compilation).

The macro `__SIZEOF_INT128__` is available in GCC >= 4.1, CLANG >= 3.3 and ICC (Intel's compiler) >= 16 which should be old enough now.

I don't have your specifications as a qemu system ready, so it would be *really* nice if you could give the patch

```
  #   if !(defined(MP_32BIT) || defined(MP_16BIT) || defined(MP_8BIT))
- #      if defined(__GNUC__) && !defined(__hppa)
+ #      if defined(__GNUC__) && defined(__SIZEOF_INT128__) && !defined(__hppa)
  /* we support 128bit integers only via: __attribute__((mode(TI))) */
  #         define MP_64BIT
  #      else

```

in `tommath.h` a quick try?

DerDakon:
Builds fine and passes the testsuite when applied to 1.2.0.

czurnieden:
Ah, thanks a lot!

I'm currently alone here as it seems, so I can't tell you how long the new version takes. @sjaeckel tried to put it in Debian's "Bullseye"\* but was a bit too late?

\* I know that Bruce Perens used to work at Pixar and found it funny to use the names of the Toystory characters as Debian names but "Bullseye" (I think it is the horse?) for an OS version? Really? What comes next, "Oneeiiiiiighty"?

PR #510, opened 2021-07-10
Added check for __SIZEOF_INT128__ to avoid some problems with a 64-bit arch. and 32-bit userspace

czurnieden:
Regarding issue #509 

The macro `__SIZEOF_INT128__` is in GCC >= 4.6, CLANG >= 3.3, and ICC >= 16 (probably earlier but I don't have them), I do not know if anybody uses older compilers.


DerDakon:
If you can give me a simple test I can test on 3.4 and 4.1..4.5, which I have around but which are too old to be used as system compilers.

czurnieden:
As we are slowly removing the support for some the stone-old compilers we should be able to assume that the mentioned macro `__SIZEOF_INT128__` is available.
Negative sideeffects of including it are not known but highly unlikely.

sjaeckel (approved):

PR #511, opened 2021-07-20
Reduce stack usage for embedded targets

jlaitine:
Hi,

I was using libtomcrypt on an embedded target (pixhawk 4 autopilot), and found out that in order to do RSA2048 encryption, I had to increase stack size for my calling thread to ridiculously big.

In order to make it more usable I decided to allocate some large arrays from heap instead of stack in libtommath. I decided to also post the change in here as a PR, in case it is seen usable for others as well.

---------

In s_mp_exptmod*.c and s_mp*comba.c, allocate large arrays from
heap instead of stack. In targets where stack sizes are
typically fixed at task creation time, this is more efficient memory usage

Signed-off-by: Jukka Laitinen <jukkax@ssrc.tii.ae>

### DIFF
--- a/mp_prime_strong_lucas_selfridge.c
+++ b/mp_prime_strong_lucas_selfridge.c
@@ -52,7 +52,8 @@ mp_err mp_prime_strong_lucas_selfridge(const mp_int *a, bool *result)
 {
    /* CZ TODO: choose better variable names! */
    mp_int Dz, gcd, Np1, Uz, Vz, U2mz, V2mz, Qmz, Q2mz, Qkdz, T1z, T2z, T3z, T4z, Q2kdz;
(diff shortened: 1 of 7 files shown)


sjaeckel:
Hi!

thanks for the PR! Just FYI:

* I'm going to cherry-pick fd73d7630b9d3ed5a79d613ff680a549e9780de7, that one is fine
* I'm not going to include d6d709341e088bc954f232102cc74bd8d661c88e as that one results in a severe performance impact. This has already been proposed and discussed multiple times in several issues, e.g. #441 

ecc_decrypt_key
ecc_encrypt_key
ecc_sign_hash
ecc_verify_hash

jlaitine:
Yes, the performance impact was expected. It would be nice to find a solution which enables embedded targets and still is perfornant. Thanks for the review! I am closing this PR

PR #512, opened 2021-09-21
s_mp_rand_platform.c: s_read_urandom: correctly handle split read

arnout:
s_read_urandom has a while loop to handle read() that returns less
than the full buffer (either due to EINTR or because more than the
atomic guarantee from urandom was requested). However, the target
of the read was always the base pointer p instead of the updated
pointer q, so in the end less than the requested randomness is returned.

Use q instead of p in the read() call.


sjaeckel (approved):
thx!

PR #518, opened 2021-12-03
Support use `#include <tommath.h>` for SwiftPM dependencies

cntrump:
for https://github.com/libtom/libtomcrypt/pull/575

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -11,8 +11,8 @@ let package = Package(
     products: [
         // Products define the executables and libraries a package produces, and make them visible to other packages.
         .library(
-            name: "TomMath",
-            targets: ["TomMath"])
+            name: "libtommath",
+            targets: [ "libtommath" ])
     ],
     dependencies: [
         // Dependencies declare other packages that this package depends on.
@@ -22,7 +22,7 @@ let package = Package(
         // Targets are the basic building blocks of a package. A target can define a module or a test suite.
         // Targets can depend on other targets in this package, and on products in packages this package depends on.
         .target(
-            name: "TomMath",
+            name: "libtommath",
             path: ".",
             exclude: ["demo", "doc", "etc", "logs", "mtest"],
             sources: ["."],
@@ -31,9 +31,9 @@ let package = Package(
                 .unsafeFlags(["-flto=thin"])  // for Dead Code Elimination
             ]),
         .testTarget(name: "TomMathTests",
-                    dependencies: ["TomMath"],
+                    dependencies: [ "libtommath" ],
                     path: "demo",
-                    sources: ["tommath_tests.swift"])
+                    sources: [ "tommath_tests.swift" ])
     ],
     cLanguageStandard: .gnu11,
     cxxLanguageStandard: .gnucxx14
--- a/Package.swift
+++ b/Package.swift
@@ -4,15 +4,15 @@
 import PackageDescription
 
 let package = Package(
-    name: "tommath",
+    name: "TomMath",
     platforms: [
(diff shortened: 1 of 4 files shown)


sjaeckel:
How about setting the include path correctly instead? Or is that not possible in the swift thing?

@royjit maybe you can give some hints here how this is done properly!? Adding another header simply doesn't look right to me ...

cntrump:
> How about setting the include path correctly instead? Or is that not possible in this swift package manager thing?

`A` depend on package  `B` use `dependencies: [.package(url:...)]` in `Package.swift`
`A` import `B` use `#include <B.h>` in source
SwiftPM find `B.h` in `publicHeadersPath` on `Package.swift` of B

now, SwiftPM can't find `tommath.h` on path `modulemap` defined in `Package.swift` of libtommath

sjaeckel:
After reading SE-0038 I'm still not entirely sure if that's the correct way to do things.

@ddunbar sorry highlighting you here, but you are the author of said specification and maybe you can shed some light on this?

sjaeckel:
Reply to https://github.com/libtom/libtomcrypt/pull/575#discussion_r762003741

Why is this building `mtest` if it's on the exclude list?

https://github.com/libtom/libtommath/blob/2cec6addaf4acbd8250611ec4cecadf1f4d655ee/Package.swift#L27

cntrump:
> Why is this building `mtest` if it's on the exclude list?

It seems `.unsafeFlags(["-flto=thin", "-fmodule-map-file=modulemap/module.modulemap"])` is ignored.

a new `module.modulemap` created:

```
module libtommath {
    umbrella "/a/b/c/libtommath"
    export *
}
```

`umbrella "/a/b/c/libtommath"` expose all symbols in `*.h`

cntrump:
@sjaeckel

How about build `libtomcrypt` as objectiv-c source and add `@import TomMath;` in `math/ltm_desc.c` ?

I tried, it works and just `math/ltm_desc.c` modified.

cntrump:
> [creating-c-language-targets](https://github.com/apple/swift-package-manager/blob/main/Documentation/Usage.md#creating-c-language-targets)
To allow a Swift target to import a C language target, add a target dependency in the manifest file. Swift Package Manager will automatically generate a modulemap for each C language library target for these 3 cases:
> 
> -    If `include/Foo/Foo.h` exists and `Foo` is the only directory under the include directory, and the include directory contains no header files, then `include/Foo/Foo.h` becomes the umbrella header.
>
> -    If `include/Foo.h` exists and `include` contains no other subdirectory, then include/Foo.h becomes the umbrella header.
>
>  -   Otherwise, the `include` directory becomes an umbrella directory, which means that all headers under it will be included in the module.
>
> In case of complicated `include` layouts or headers that are not compatible with modules, a custom `module.modulemap` can be provided in the `include` directory.

Keep trying

cntrump:
Swift 5.2 Issue with headers in same dir as `.c/m` files
https://forums.swift.org/t/swift-5-2-issue-with-headers-in-same-dir-as-c-m-files/33292

[PackageLoading] Handle header files in TargetSourcesBuilder
https://github.com/apple/swift-package-manager/pull/2555

keeping a close watch on it.

cntrump:
> 
> Why is this building `mtest` if it's on the exclude list?
> 

I found it, it'is a bug: https://bugs.swift.org/browse/SR-14456

sjaeckel:
I'm closing this as there was no action for too long.
Please re-open if you have further questions/comments/improvements and still want this to be merged.

Issue #521, opened 2021-12-13
mp_is_square says 0 is not a square


guidovranken:
mp_is_square says 0 is not a square, but it is.

czurnieden:
This is the original behaviour  (Try `git log --follow -p --  mp_is_square.c`) and other software, especially LibTomCrypt needs to be checked before we change something here.

But at least the documentation should get an update; the function `mp_is_square` is listed without a proper description.

Thanks a lot!

sjaeckel:
Thanks for reporting!

Closed via #522

guidovranken:
Thank you for fixing.

PR #522, opened 2021-12-13
Introduced 0 (zero) to mp_is_square as a perfect square

czurnieden:
Instead of just adding some documentation regarding issue #521 I decided to fix that bug.

 - changed definition of `0` (zero) from not-a-square to is-a-square.
 - test added to check `0` (zero)
 - additional test added to check negative numbers

Does not seem to conflict with LibTomCrypt but that needs a proper check, I just grep'd for `is_square`.

Changes API and hence might conflict with other software that uses LTM but on the other side: it *was* a bug that gets fixed with this patch.

### DIFF
--- a/demo/test.c
+++ b/demo/test.c
@@ -677,6 +677,17 @@ static int test_mp_is_square(void)
 
    DOR(mp_init_multi(&a, &b, NULL));
 
+
+   /* Domain is {x \in \mathbb{Z} : x \le 0}  */
+   mp_set_l(&a, -1);
+   EXPECT(mp_is_square(&a, &res) == MP_VAL);
+   EXPECT(!res);
+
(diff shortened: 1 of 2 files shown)


sjaeckel:
> Changes API and hence might conflict with other software that uses LTM but on the other side: it _was_ a bug that gets fixed with this patch.

How does this change the API? FMU it only fixes the bug.

> Does not seem to conflict with LibTomCrypt but that needs a proper check

LTC doesn't even have a direct mapping of `mp_is_square()` into the MPI provider, so it doesn't matter.

The single LTM-internal consumer is `mp_prime_is_prime()` and there the behavior doesn't change since `0` is even and therefor we return before `mp_is_square()` would even be called.

PR #535, opened 2022-09-07
Fix: removed sign operation in s_mp_invmod_odd (issue #534)

### LINKED ISSUE #534: s_mp_invmod_odd returns wrong result for negative numbers
Using `mp_invod` can return incorrect results for negative numbers. Using it to find the inverse of `-1 mod 7` yields `-6` instead of `6`. The problem looks like it could be because of a couple of places:

https://github.com/libtom/libtommath/blob/4b47368501321c795d5b54d87a5bab35a21a7940/s_mp_invmod_odd.c#L31-L32
The comment is assuming that `mp_mod` gives the absolute value, but it doesn't. It will make the value positive though.

Then, in https://github.com/libtom/libtommath/blob/4b47368501321c795d5b54d87a5bab35a21a7940/s_mp_invmod_odd.c#L98-L109
the sign of `a` is used to flip the sign of the inverse, but the sign should already be correct since `mp_mod` is using a positive member of the equivalence class for `a`. Removing this sign change should fix it.

czurnieden:
Changed a check for <0 with mp_isneg() in s_mp_invmod

Additional tests for mp_invmod() in demo/test.c

This fixes #534


demo/test.c
```diff
@@ -564,6 +612,28 @@ static int test_mp_invmod(void)
...
+   /* Some small general tests https://github.com/libtom/libtommath/issues/534 */
+   for (i = -10; i < 11; i++) {
+      for (j = -10; j < 11; j++) {
+         mp_set_i32(&a, i);
+         mp_set_i32(&b, j);
+         e = mp_invmod(&a, &b, &c);
+         if (e != MP_OKAY) {
+            if (results[i+10][j+10] != -1) {
+               printf("error = %s from ", mp_error_to_string(e));
+               printf("error at i = %d, j =%d should be an error but gave %d\n",i,j, results[i+10][j+10]);
```
friedrichsenm, line 624: Shouldn't this print out the value of `c` instead of the expected results table?

czurnieden: Yes, of course it should!
Thanks!

friedrichsenm (approved):

PR #548, opened 2023-05-25
typo fix.

renweibo:
typo fix.

### DIFF
--- a/demo/test.c
+++ b/demo/test.c
@@ -8,7 +8,7 @@ static long rand_long(void)
 {
    long x;
    if (s_mp_rand_jenkins(&x, sizeof(x)) != MP_OKAY) {
-      fprintf(stderr, "s_mp_rand_source failed\n");
+      fprintf(stderr, "s_mp_rand_jenkins failed\n");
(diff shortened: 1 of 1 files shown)


czurnieden:
Sorry for the delay but I was unsure what it is. It is not a typo, it was intentionally worded that way: the kind of (P)RNG is not relevant, it is just a random (P)RNG. There are some that can fail (e.g.: lack of entropy) but this one cannot (and if it does: get you checkbook ready because the likelihood is  very high that your PC is falling apart)..
So my verdict is; notabug,

But don't get too disappointed by that we are always thankful for any help!

Oh, btw:
Please don't push to `develop` next time, open a new branch and push that, Example:
In your local fork do;
```
$ git checkout develop
$ git pull upstream develop
$ git checkout -b name_for_new_branch
$  ... do your work here ... 
$ git push origin name_for_new_branch 
```

You most likely will get a URL if you are using Github that you  c&p into your browser to come to a page where you fill out the forms, get some more information (does it have conflicts etc.) and wait for somebody to come by. May take a day or two if it isn't a zero-day or sth alike.

PR #550, opened 2023-05-29
Implementation of mp_printf

czurnieden:
Implementation of a `mp_(f)printf` with the additional type modifiers `Z` (signed big integer), `M` (unsigned limb), and `N` a printout of `big_integer->dp`, all formatable (to some extent) as in `printf(3)` and with a binary representation, too. See documentation and source for details.

Uses `fprintf(3)` for output and some of the formatting so it is bracketed inside `#ifndef MP_NO_FILE`.

It is rather large (6k - 12k stripped, depending on `MP_xxBIT`) but the folks over at GMP were right: you basically have to write a complete `printf`. Well, almost complete.

PR #549, opened 2023-05-28
made MP_31BIT working

czurnieden:
Found out (the hard way) that `MP_31BIT` is not working. Added `mp_hash` values for `MP_31BIT` and hope that was all.

PR #551, opened 2023-05-29
Print big integers with (f)printf

czurnieden:
Simple implementation of a big integer extension to `printf(3)`


czurnieden:
Variation of #550 following a suggestion by @sjaeckel  . GLibC only.
This is only the bare minimum, more to come but the documentation about `printf.h` is sparse&mdash&more or less: "See the source".
Odd errors incoming. Hu?

czurnieden:
Failure of `mp_log` is unsanitized input. Can happen if you use a PRNG for the input ;-) Needs a check, will add it tommorow.

The other one "/home/runner/work/libtommath/libtommath/demo/test.c:382: undefined reference to `mp_printf_extension'" needs a bit more work, I'm afraid.

czurnieden:
I broke it?

czurnieden:
Rebased to hide all of the pain ;-)
Had to install a dummy function for MSVC but that is not how it should be done. At least I hope that that is not how it should be done but I am no expert in that regard.
Any ideas?

czurnieden:
The problem with the unchecked input for `mp_log` tests mentioned above is in #552 .

czurnieden:
These are prefix modifiers like `ll` or `z`. It is not possible (Might possible but I have not found a way to do so) to use the common specifiers, so things like `%Nx` for hexadecimal or `%No` for octal do not work, sadly, it is `%kN` and `%rN` respectively.

 - `k` for hexadecimal representation ("k" like in "eks")
 - `b` for binary representation (it is planned in C2x so might need to be changed later)
 - `r` for octal representation ("r" like in "Rear Admiral Grace Brewster Murray Hopper"  which has 07 r's and a reason )
 - `@` for base-64 representation (because `@` is not in any base-64 character list according to RFC-4648)

Prefixes of the printed numbers are:
 - `0x`  for hexadecimal representation
 - `0b`  for binary representation
 - `0o` for octal representation (can be changed to `0` if you wish so)
 - `0@` for base-64 representation

Most letters of the English alphabet, both upper and lower case, are already taken by printf(3)'s format.

The modifiers can be more than one character, so something like "b16N" or "b64N" would come handy but the letter "b" might get taken by C2x for binary representation. Digits without anything else are possible, too.
It is quite easy to change, so don't be shy and tell me if you hate my idea and prefer e.g.: the `bxxN` style.

BTW: the way it works is that for every modifier one bit of a `short` is set, so there is no  way to implement all 62 possible bases.

The modifier accepting function takes `wchar_t *` but I would refrain from using unicode, more trouble than it's worth.

Not implemented are:
- leading zeros
- thousands separators
- precision ( That is the minimum digits to print for an integer, so basically the same as adding leading zeros.)

This function is not thread safe! But who would expect thread safety from a `printf` anyways?

The functions to print a single limb and the raw limb-array (`a.dp`) are not yet implemented.

czurnieden:
No, that is way to much effort to print a simple native integer, there must be a simpler way to do it. Or at least without the heavy artillery (`sprintf`) I used.

czurnieden:
Conflict is an empty line in `tommath.h`? Hu?

czurnieden:
@sjaeckel I saw that I forgot to add `MP_31BIT` tests to #549 Kept that commit separate and at the end if it is wanted.

Phew, took by far more time to write the tests for all our optional flavours than to write the actual code! ;-)

But that is what the GLibC printf-extension is able to do: print the big integer, one limb and the limb-array in all useful bases and the formatting that printf offers (sans the thousand-separators, don't think it is a good idea to start meddling with locales, but maybe an underbar for all locales, a space (U+2009?) maybe?).

The documentation is a bit sparse for now. I'll check later.

The extension allows for some more fiddling. It is, for example, possible to set the padding character (a `wchar_t` so one can have some fun with it).

`mp_printf_extension_init` takes no arguments now, but could take e.g.: a string with a list of the specifiers that are wanted (might get complicated with `mp_printf_extension_clear` if there are different/more specifiers given than those that get unregistered by default).

PR #557, opened 2023-06-15
Deploy via gh actions

sjaeckel:
In order to be able to build ltc, we need an installable ltm.

I did this manually in the past, lets use GH actions now.


sjaeckel:
I couldn't test this w/o merging it first, please add comments if there are any!

e.g. if it makes sense to rename the generated packages depending on whether we're on a tag or not

PR #561, opened 2023-06-21
Update of examples in directory "etc"

czurnieden:
Should compile woithout warnings and `pprime.c` should work now.


czurnieden:
Changed `rand()` to `mp_prime_rand()` because inadvertently somebody *will* use it to generate RSA primes.

sjaeckel (approved):

PR #564, opened 2023-07-14
Deny 2k-reduce if lsd is zero

czurnieden:
Algorithm as implemented does not work if the least significant digit is zero.

Did not want to touch the algorithm itself (for now) so I put a  check in `mp_reduce_is_2k.c` instead.

### DIFF
--- a/demo/test.c
+++ b/demo/test.c
@@ -1229,11 +1229,20 @@ LBL_ERR:
 static int test_mp_reduce_2k(void)
 {
    int ix, cnt;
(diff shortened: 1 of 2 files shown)


czurnieden:
It is an edgecase that falls right in the gap between `reduce_2k` and `reduce_2k_l`.

I'll try to come up with a solution that does not change the API and is not too slow but until than this fix is good enough; the edgcase `n \equiv 0 (mod \beta)` is relatively rare.

PR #571, opened 2024-03-11
Minor improvements

sjaeckel:
See commit messages for details.


sjaeckel:
92e9f32b9f5448bae66180d2cab57408fabc4db8 didn't make it better, but also not worse. let's keep it.

PR #572, opened 2024-03-13
Deprecate APIs

sjaeckel:
I can't remember what the plan was of @minad regarding those APIs, it's too long ago.

Still we most likely will make a new release from develop at one point and then we shouldn't pull those APIs without warning our users first.


czurnieden (approved):
Yes, looks good, although I cannot remember the exact reasons for that either.

PR #576, opened 2024-03-27
Build both static and shared libraries with CMake

gahr:
This uses CMake's Object Libraries feature to compile object files once and used them to build both static and shared libraries. See https://cmake.org/cmake/help/latest/command/add_library.html#object-libraries for details.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,11 +29,6 @@ include(CTest)
 include(sources.cmake)
 
 #-----------------------------------------------------------------------------
-# Options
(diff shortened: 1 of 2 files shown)


sjaeckel:
Thanks for this PR. It took me a while to think about it and research what's possible, usual, resp. doable.

FMU that PR would turn our CMake project into a very "non-CMake-style" state. None of the library projects I've looked at does this, they're all either static or shared.

I can understand why you would want that, but I don't think it'd be a good idea to diverge from the usual way how things are done in this case. Sorry :)

PR #577, opened 2024-03-31
cmake: Simplify ccache logic

diizzyy:
* Use USE_CCACHE switch, this seems to be a more common option having a quick look using Google
* Make use of find_program() functionality introduced in CMake 3.18 to simplify code

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,20 +34,6 @@ include(sources.cmake)
 option(BUILD_SHARED_LIBS "Build shared library and only the shared library if \"ON\", default is static" OFF)
 
 #-----------------------------------------------------------------------------
-# Add support for ccache if desired
-#-----------------------------------------------------------------------------
-find_program(CCACHE ccache)
-
-if(CCACHE)
-    option(ENABLE_CCACHE "Enable ccache." ON)
-endif()
-
-# use ccache if installed
-if(CCACHE AND ENABLE_CCACHE)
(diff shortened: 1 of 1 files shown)


sjaeckel (changes requested):
DDG doesn't turn up a single hit for `USE_CCACHE` but G apparently finds some.

This rename falls under API breakage IMO, i.e. it must be backported to 1.3 and a warning must be added that it will change in the future.

This PR doesn't only simplify code, but also changes behavior.

I don't like the functionality change where the old solution defaults to enable its usage if `ccache` is found and with your change it's disabled by default. As there was no comment describing why this was done, I suspect that this was accidental!?

Also: please fix CI :)

diizzyy:
I think the "most" correct way would be to rip it out completely and leave it up to the user compiling it or the framework to specify rather than blindly looking for a binary.  https://stackoverflow.com/questions/1815688/how-to-use-ccache-with-cmake
Edit: Engrish ;-)

I can go either way :)

sjaeckel:
> I think the "most" correct way would be to rip out of completely and leave it up to the user compiling it or the framework to specify it

Thanks, that's exactly what we should do.

Do you wanna modify this PR to simply drop it or should I?

diizzyy:
I'll update the PR, thanks! :)

diizzyy:
Updated

sjaeckel:
"Backported" via 5410d0bfbfb804b08b7b789a893d76d1fb2f720d

PR #579, opened 2024-05-15
Addition of rudimentary error tracing

czurnieden:
This adds a compile-time-optional macro that inserts a line to print out the filename, line-number, and function to `stderr` before all of the `goto LTM_ERR;` instances and all `return err;` if `err != MP_OKAY`.


czurnieden:
Eases the work quite a bit if the error comes from deep inside a recursive function.
Found it very useful while debugging the fast-radix-conversion at the upper limit of LTM and had to add higher Toom-Cook functions (4 - 7) to avoid waiting a week to get the result. (One `MP_OVF` had the reason that `mp_add` adds an extra limb unconditionally, so the maximum number is `2^((MP_MAX_DIGIT_COUNT - 2)*MP_DIGIT_BIT)`)

Ah, the first typo has already been found :-(

PR #581, opened 2024-06-16
Repairs and clean-ups regarding tuning and profiling

czurnieden:
- added tuning and graphics to cmake
 - removed RTDSC based timers
 - gnuplot generated plots via tune_it.sh
 - tuning is available in `makefile.shared` now


czurnieden:
Changes from the additional tuning are small but YMMV, as always.

czurnieden:
CMake is a bit more complicated once it gets a bit more complicated \*sigh\*

czurnieden:
Ah, forgot: cannot check if the windows part works. Will somebody who can take a look, please?

czurnieden:
Well&hellip;
I posted at [stackoverflow](https://stackoverflow.com/questions/78638550/run-profiling-fprofile-arcs-in-cmake) although I doubt it.

But I found a working [CMake formatter](https://github.com/cheshirekow/cmake_format) so my search wasn't a *complete* waste ;-)

czurnieden:
Rewound the CMake part (including the formatting, sorry) and  collapsed the commits.

I think I found the problem with the odd results in tuning. We use "-O3" for tuning for a good reason but it has a side effect. With checks included (option `-c` for `tune.c`) the numbers are all reasonable, especially the first higher T-C-squarings after TC-3 that was *really* off. Not switched on here&ndash;nothing higher than TC-3&ndash;but put in this comment such that I do not forget it later.

czurnieden:
Timer works with MSVC, at least in [Godbolt's Compiler Explorer](https://godbolt.org/z/4Yx8a8xKs), I just couldn't get it work under Wine. Probably too old.

PR #583, opened 2024-11-19
Make sure s_mp_rand_source is defined for all its uses

### LINKED ISSUE #582: [linux-error]undefined symbol: s_mp_rand_source
I use libtommath in java, the error is as follows(it’s OK on mac os)：
```
xxx/linux-x86_64/libjniTommath.so: undefined symbol: s_mp_rand_source
	at java.base/jdk.internal.loader.NativeLibraries.load(Native Method)
	at java.base/jdk.internal.loader.NativeLibraries$NativeLibraryImpl.open(NativeLibraries.java:331)
	at java.base/jdk.internal.loader.NativeLibraries.loadLibrary(NativeLibraries.java:197)
	at java.base/jdk.internal.loader.NativeLibraries.loadLibrary(NativeLibraries.java:139)
	at java.base/java.lang.ClassLoader.loadLibrary(ClassLoader.java:2418)
	at java.base/java.lang.Runtime.load0(Runtime.java:852)
	at java.base/java.lang.System.load(System.java:2025)
	at org.bytedeco.javacpp.Loader.loadLibrary(Loader.java:1779)
```

levitte:
mp_prime_rand.c uses s_mp_rand_source as well, so ensure that the
latter is defined on the conditions of the former as well.

Fixes #582

### DIFF
--- a/mp_rand_source.c
+++ b/mp_rand_source.c
@@ -1,5 +1,7 @@
 #include "tommath_private.h"
-#ifdef MP_RAND_C
+/* MP_RAND_C relates to mp_rand.c, MP_PRIME_RAND_C relates tp mp_prime_rand.c
+   Both use s_mp_rand_source */
+#if defined(MP_RAND_C) || defined(MP_PRIME_RAND_C)
 /* LibTomMath, multiple-precision integer library -- Tom St Denis */
 /* SPDX-License-Identifier: Unlicense */
 


mp_rand_source.c
```diff
@@ -1,5 +1,7 @@
 #include "tommath_private.h"
-#ifdef MP_RAND_C
+/* MP_RAND_C relates to mp_rand.c, MP_PRIME_RAND_C relates tp mp_prime_rand.c
+   Both use s_mp_rand_source */
+#if defined(MP_RAND_C) || defined(MP_PRIME_RAND_C)
```
sjaeckel, line 4:
```suggestion
#if defined(MP_RAND_SOURCE_C) || defined(S_MP_RAND_SOURCE_C)
```

This should be selected automagically via `tommath_class.h`, but if we change it as proposed this won't be the case. My suggestion is also not correct ... I guess the really correct way would be if the file looked as follows!? What do you think?

```c
#include "tommath_private.h"
#ifdef S_MP_RAND_SOURCE_C
/* LibTomMath, multiple-precision integer library -- Tom St Denis */

mp_err(*s_mp_rand_source)(void *out, size_t size) = s_mp_rand_platform;
#endif

#ifdef MP_RAND_SOURCE_C
void mp_rand_source(mp_err(*source)(void *out, size_t size))
{
   s_mp_rand_source = (source == NULL) ? s_mp_rand_platform : source;
}
#endif
```

Or maybe this should even be two files?

sjaeckel (changes requested):

mp_rand_source.c
sjaeckel: I took the liberty to push https://github.com/libtom/libtommath/tree/fix-582

levitte: Yeah that works too.  I suggest you merge that, then close this PR

PR #584, opened 2025-01-29
Support "Cryptography API: Next Generation" on Windows

paperchalice:
Use Cryptography API: Next Generation, because UWP doesn't support old api.
There is a simpler implementation if we don't care about RSA:
```c
BCryptGenRandom(NULL, (PUCHAR)p, (ULONG)n, BCRYPT_USE_SYSTEM_PREFERRED_RNG);
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,6 +23,8 @@ set(PACKAGE_RELEASE_VERSION 1)
 include(GNUInstallDirs)
 include(CheckIPOSupported)
 include(CMakePackageConfigHelpers)
+include(CMakePushCheckState)
+include(CheckSymbolExists)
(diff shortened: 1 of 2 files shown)


sjaeckel:
Thanks for the PR :)

Could you please check how we did the bcrypt integration in ltc and copy that instead?

https://github.com/libtom/libtomcrypt/blob/develop/src/prngs/rng_get_bytes.c

sjaeckel (changes requested):
Changes look fine!

Can you please squash the two commits and then we're ready to merge.

Thanks again :)

levitte (approved):
Yup, looks good to me too

sjaeckel (approved):

PR #585, opened 2025-01-30
Generate import library in CMake

paperchalice:
CMake supports `.def` as input to generate import library.


paperchalice:
BTW, cmake supports export symbols automatically (https://www.kitware.com/create-dlls-on-windows-without-declspec-using-new-cmake-export-all-feature/), but it exports some unexpected functions...

sjaeckel:
Huh, why closed?

paperchalice:
Oops, I mistakenly closed this PR when I was cleaning my forks... I opened #589 now.

PR #597, opened 2026-04-20
Check size of argument

### LINKED ISSUE #595: mp_from_sbin size check
mp_err mp_from_sbin(mp_int *a, const uint8_t *buf, size_t size)

Any call with size <= 0 through this API may exhaust process memory

Suggest to return early if size <= 0

czurnieden:
An input of `size == 0` will cause a wrap-around when a one is subtracted later which itself will grab a very large amount of memory.


sjaeckel (approved):

sjaeckel:
Thanks! :)

